PR #266, opened 2026-07-14
Unblock Cypress from database pool exhaustion

silasfelinus:
## What this fixes

The current production deployment is timing out every database-backed route with MariaDB pool state `active=0 idle=0 limit=10`. Cypress first exposed this through the orphan sweep, but user registration and `/api/health/database` fail the same way.

- Replace the full `/api/users` orphan scan with a narrow admin-only database cleanup endpoint.
- Make stale-user maintenance best-effort so it never prevents the actual test suite from starting.
- Reuse cached seed users after interrupted local runs instead of creating three more accounts.
- Support exact `cypress-*` username rollback after ambiguous registration failures.
- Bound cleanup batches for serverless runtime limits.
- Reduce the default per-function MariaDB pool from 10 connections to 2.
- Reduce pool acquisition from 10s to 6s and transient retries from 5 to 2 so requests can return 503 before Vercel's 40-second timeout.

## Evidence

Production runtime logs show repeated pool acquisition timeouts on users, projects, heartbeat, art queue, and the database health endpoint, all with zero active and zero idle connections. Five retries cannot finish within the function deadline because each acquisition may consume ten seconds.

## Validation

TypeScript syntax parsing succeeded for the changed seed helper and cleanup endpoint. Exact-head Vercel validation may still be blocked by the account build-rate limit; production database health must be checked after deployment before rerunning Cypress.

### DIFF
--- a/content/channels/home/account.md
+++ b/content/channels/home/account.md
@@ -1,0 +1,17 @@
+---
+contentType: tab
+channelKey: home
+tabKey: account
+dashboardKey: user
+dashboardTab: account
+label: Account
+title: Account & Privacy
+subtitle: Your settings, your consent
+description: Change your password, verify your email, tune privacy and consent, and manage newsletter updates.
+icon: kind-icon:settings
+route: /account
+sort: 70
+requiredPermission: authenticated
+---
+
+Account settings and consent controls belong in the user's Home channel.
--- a/content/channels/home/friends.md
+++ b/content/channels/home/friends.md
@@ -1,0 +1,17 @@
+---
+contentType: tab
+channelKey: home
+tabKey: friends
+dashboardKey: user
+dashboardTab: friends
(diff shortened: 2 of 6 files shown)


silasfelinus:
Contract Tests are green on run 719 after adding the missing Home-channel tab manifests for `/account`, `/friends`, and `/messages`.

Root cause: the workflow stopped at `verifyChannelContent.ts` because those three existing page documents referenced undefined tabs. This was unrelated to the Cypress/database-pool changes in this PR.

Validated steps now passing: channel content, channel resolver, navigation route access, Challenge Center dry-run seed validation, and channel artwork audit.

The separate TypeScript check previously reported repository-wide Prisma JSON/string type drift; none of its diagnostics referenced this PR's Cypress cleanup or Prisma pool files.

PR #294, opened 2026-07-15
ci(cypress): accept a superseding commit as a live deploy

silasfelinus:
## Summary
- `.github/workflows/cypress.yml`'s deploy-wait step polled `/api/version` for an *exact* match on `TARGET_SHA`, timing out at 600s (now configurable, default 1200s) if production never served that exact commit.
- During a burst of several merges within 1-2 hours (e.g. PR #238-#242), Vercel starts a new production build for each new `main` commit before the previous one finishes. By the time any single build would go live, a newer commit has already superseded it — production stays pinned on an older SHA and the exact-match poll fails red even though nothing is actually broken (each commit's own build was still queued/building).
- Fix: when the live commit doesn't exactly match `TARGET_SHA`, check whether `TARGET_SHA` is a git ancestor of the live commit. If so, the live commit already contains `TARGET_SHA`'s changes (a "superseded but correct" state) and the wait succeeds instead of timing out. The shallow checkout is only deepened (`--depth=200`) when this ancestry check is actually needed, so the common exact-match case stays fast.

This is conductor `kind-robots/t-018` (found during a Reviewer sweep 2026-07-14, root-caused and documented there; not urgent since it self-resolves once merge bursts settle, but flakes CI red during any future rapid-merge session).

## Test plan
- [x] `.github/workflows/cypress.yml` parses as valid YAML.
- [x] Verified the new ancestry-check shell logic in isolated scratch git repos simulating both the shallow-checkout "superseded" case (commit accepted) and a non-superseding sibling-branch commit (correctly rejected).
- [ ] Live verification requires an actual merge-burst Vercel deploy race, which can't be reproduced outside CI — the next time this scenario occurs the workflow run will confirm behavior end-to-end.

https://claude.ai/code/session_01Vov3WyjVX673po7g1nX2Nq

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vov3WyjVX673po7g1nX2Nq)_

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -39,20 +39,38 @@ jobs:
           TARGET_SHA: ${{ github.sha }}
           MAX_WAIT: ${{ github.event.inputs.max_wait_seconds || '1200' }}
         run: |
-          echo "Waiting up to ${MAX_WAIT}s for production to serve ${TARGET_SHA}..."
+          echo "Waiting up to ${MAX_WAIT}s for production to serve ${TARGET_SHA} (or a commit that supersedes it)..."
           deadline=$(( $(date +%s) + MAX_WAIT ))
           live=""
           while [ "$(date +%s)" -lt "$deadline" ]; do
             body="$(curl -sf --max-time 15 "${BASE_URL}/api/version" || true)"
             live="$(printf '%s' "$body" | jq -r '.data.commit // empty' 2>/dev/null || true)"
-            if [ -n "$live" ] && [ "$live" = "$TARGET_SHA" ]; then
-              echo "Production is serving ${live} — deploy is live."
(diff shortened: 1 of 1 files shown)


silasfelinus:
The "TypeScript" check failure is pre-existing and unrelated to this PR's diff (`.github/workflows/cypress.yml` only — not part of the TS project). Reproduced locally on this branch: `npm run test` (vue-tsc) fails with 19 errors across 10 files, all in `components/`/`server/api/`/`stores/` app source (InputJsonValue/Prisma type mismatches, `animationStore` export drift, etc.) — the same pre-existing break tracked at conductor `kind-robots/t-020` (previously as high as 82 errors; down to 19 now after intervening merges, but still present on `main`). Merging.

---

PR #323, opened 2026-07-16
fix(projects): bypass stale Prisma pool with direct MariaDB upsert

silasfelinus:
## Production finding

#322 removed the dangerous shared-client reset and correctly isolated stale Project creates. The live Project Sync rerun proved a second adapter-level failure: each brand-new one-connection `PrismaMariaDb` pool timed out with `active=0 idle=0 limit=1` before opening a socket.

Meanwhile unrelated production routes continued succeeding, and Kind Robots already has a direct MariaDB connection path used by its database probe and build preflight.

## Fix

- Export the existing verified direct MariaDB connection factory from `databaseDirectProbe.ts`.
- On the exact stale Prisma socket signature only, perform the Project write through a direct single connection instead of another Prisma adapter pool.
- Use parameterized `INSERT ... ON DUPLICATE KEY UPDATE`, keyed by unique `conductorSlug`, preserving idempotency if the original create committed ambiguously.
- Select and return the newly created Project shape after the write.
- Always close only the direct connection.
- Remove the failed isolated-Prisma factory; never disconnect the shared Prisma client.

## Safety

- Normal Project creates still use Prisma and the existing include shape.
- The direct path runs only after the shared Prisma retry wrapper exhausts the exact `Cannot execute new commands: connection closed` error.
- SQL identifiers are static and all values are parameterized.
- The Project model guarantees unique `conductorSlug`.

## Regression guard

The database contract asserts:

- shared `basePrisma` is never disconnected
- the failed isolated-Prisma factory does not return
- direct connection creation is exported from the verified probe path
- fallback SQL remains idempotent and closes its connection
- `POST /api/projects` invokes the direct fallback only after stale classification

## Validation plan

1. TypeScript and Contract Tests.
2. Vercel preview build and database TLS preflight.
3. Merge and wait for production READY.
4. Rerun Project Sync run `29526629136` and verify `kindrobots-unraid` is created without 503s.

### DIFF
--- a/server/api/projects/index.post.ts
+++ b/server/api/projects/index.post.ts
@@ -5,10 +5,8 @@ import type {
   ProjectPriority,
   ProjectStatus,
 } from '~/prisma/generated/prisma/client'
-import prisma, {
-  createIsolatedPrismaClient,
-  isStaleDatabaseConnectionError,
-} from '~/server/utils/prisma'
+import prisma, { isStaleDatabaseConnectionError } from '~/server/utils/prisma'
+import { upsertProjectDirect } from '~/server/utils/projectDirectWrite'
 import { errorHandler } from '~/server/utils/error'
 import { requireApiUser } from '~/server/utils/authGuard'
 import { enforceProjectCap } from '~/server/utils/projectCap'
@@ -51,7 +49,7 @@ type ProjectCreateBody = {
   isMature?: unknown
 }
 
-async function createProjectWithIsolatedFallback(
+async function createProjectWithDirectFallback(
   data: Prisma.ProjectUncheckedCreateInput,
   conductorSlug: string,
 ) {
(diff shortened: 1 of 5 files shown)


silasfelinus:
## Production validation — passed

Validated against production deployment `dpl_2ZzWyV9TrFpgfnPZT4dJE6HRgJh6` serving merge commit `e264c5f009a1dd15bd7a681e2b1c04938bc289d9` on `kind-robots.vercel.app`.

The rerun of Conductor Project Sync run `29526629136`, job `87732029629`, completed successfully.

Observed production behavior:

- `GET /api/projects/kindrobots-unraid` initially returned 404.
- The shared Prisma create exhausted two retries with `Cannot execute new commands: connection closed`.
- `[project-sync:direct-mariadb-fallback]` ran for `kindrobots-unraid`.
- `POST /api/projects` returned 201.
- A subsequent `GET /api/projects/kindrobots-unraid` returned 200.
- The same fallback successfully created `animation-manager` with POST 201, followed by GET 200.

No shared Prisma disconnect/reset is used.

## Remaining separate incident

The production Cypress run `29530660409` continues to expose the same stale Prisma adapter socket on unrelated create/update routes (`dreams`, `bots`, `characters`, `chats`, `projects`, `resources`, `scenarios`, etc.). Those requests are tagged `source: github-actions-cypress`; this is no longer a Project Sync failure and should be tracked as a broader Prisma MariaDB adapter / ProxySQL warm-pool issue.

PR #432, opened 2026-07-18
fix(db): connect standalone scripts + backup over ProxySQL-enforced TLS

silasfelinus:
## Problem

ProxySQL now **enforces SSL/TLS** on connections to the home MariaDB. The nightly *Nightly fallback snapshot & backup* workflow failed in the snapshot job:

```
ProxySQL Error: Access denied for user 'kindrobot' (using password: YES). SSL is required
pool timeout: failed to retrieve a connection from pool after 10000ms
```

## Root cause (confirmed)

- The **app runtime** already connects with SSL: `server/utils/prisma.ts` → `buildDatabaseConfig()` reads `DATABASE_SSL_CA_BASE64` / `DATABASE_SSL_CA` / `DATABASE_SSL_REJECT_UNAUTHORIZED` and passes an `ssl` option to the `PrismaMariaDb` adapter.
- The **standalone scripts** (`snapshotFallback.ts`, `exportRewards.ts`, `backfillSlugs.ts`, `repairCharacterSlugs.mjs`) did `new PrismaMariaDb(databaseUrl)` with the bare URL and **no ssl config** — so they connected without TLS and ProxySQL rejected them.
- `fallback-snapshot.yml` only passed `DATABASE_URL` into the job env — never the CA secret — so even an SSL-aware script would have had no CA in that job.

ProxySQL needs a **CA-verified TLS** connection (the same handshake `scripts/prisma-migrate-deploy.mjs` and `server/utils/databaseDirectProbe.ts` already perform), not merely `rejectUnauthorized: false`.

## Changes

- **Extracted** the SSL-aware config builder out of `server/utils/prisma.ts` into a new, side-effect-free `server/utils/databaseAdapterConfig.ts` (no eager client, no `DATABASE_URL` read at import) and **reuse** it rather than duplicating. `prisma.ts` now imports `buildDatabaseConfig` / `readDatabaseUseTextProtocol` / the int readers from it — behavior unchanged.
- **Routed** all four standalone scripts through the new `createDatabaseAdapter()`, so they complete the *same* CA-verified TLS handshake the app uses. `snapshotFallback.ts` keeps its 30s `connectTimeout` (preserved through the shared builder — it only fills a default when absent) and logs whether TLS is engaged (boolean only; never the host, creds, or cert).
- **`fallback-snapshot.yml`**: pass `DATABASE_SSL_CA_BASE64` into both the snapshot and dump job envs; make the `mysqldump` backup and the connectivity-probe `mysql` CLI present the CA via `--ssl-mode=VERIFY_CA --ssl-ca=<file>` (validates the cert against the CA without hostname checks, since the DB host is a tailnet IP/MagicDNS name not necessarily in the cert SAN); soft-warn if the secret is missing.

## Not changed

- **TLS verification is not disabled** — `rejectUnauthorized` stays `true` and the CLI uses `VERIFY_CA`. No `rejectUnauthorized: false`.
- **`davinci-seed-verify.yml` is a separate case** and left untouched: it targets a local `mariadb:11` **service container** at `127.0.0.1:3306`, not the home DB, so it needs no SSL.

## Verification

- `npm run test` (full `vue-tsc` typecheck used by CI) passes clean.
- Local harness against `buildDatabaseConfig()` confirms: no CA → plain URL (unchanged behavior); CA present → pool object with `ssl.ca` + `rejectUnauthorized: true`; the script's 30s `connectTimeout` survives into the pool config.
- Running `snapshotFallback.ts` with a CA set logs `TLS to database: on` and builds the SSL adapter (it only fails at connection time here because this environment has no tailnet route to the DB).
- End-to-end against the real DB requires the tailnet — triggering the workflow via `workflow_dispatch` from this branch to confirm green.

## Human dependency

This fix assumes `DATABASE_SSL_CA_BASE64` exists as a **repository Actions secret** (the same CA the app runtime and `prisma-migrate-deploy` already use). If it is not yet configured for Actions, add it — the workflow now emits a `::warning::` when it is absent. The DB host/address is deliberately kept out of all logs and this description.

https://claude.ai/code/session_018JzcZwcuHn2YquN2zM5KL2

---
_Generated by [Claude Code](https://claude.ai/code/session_018JzcZwcuHn2YquN2zM5KL2)_


silasfelinus:
## Verified via `workflow_dispatch` — code is correct, blocked on a missing secret

Two dispatch runs on this branch ([run 30](https://github.com/silasfelinus/kind_robots/actions/runs/29656731786), [run 31](https://github.com/silasfelinus/kind_robots/actions/runs/29657220779)):

- **snapshot job** still fails at *Write fallback snapshots* — the script logs **`[snapshot] TLS to database: OFF`**, then `ProxySQL Error: ... SSL is required`.
- Root cause is **not the code**: the job's masked env echo shows `DATABASE_URL: ***` (set) but `DATABASE_SSL_CA_BASE64:` **empty** — the CA is **not configured as an Actions secret**. Run 31 also forwards/honors the raw-PEM `DATABASE_SSL_CA` and it was empty too, so **neither** CA secret exists in Actions.
- The **dump job passes**, but that's a *red herring*: `mysqldump`/`mysql` default to `--ssl-mode=PREFERRED`, i.e. opportunistic **unverified** TLS, which ProxySQL accepts. Prisma (correctly) requires an explicit CA and does not do opportunistic TLS.

### Human step to finish this
Add the ProxySQL CA as a **repository Actions secret** (Settings → Secrets and variables → Actions), then re-run the workflow. Either form works:
- **`DATABASE_SSL_CA_BASE64`** — base64 of the CA PEM (preferred), or
- **`DATABASE_SSL_CA`** — the raw PEM.

This is the same CA the app runtime and `scripts/prisma-migrate-deploy.mjs` already use. Once it's present, the snapshot connects over CA-verified TLS and the `mysqldump` backup upgrades from unverified `PREFERRED` to `VERIFY_CA` — **no code change needed**. TLS verification is never disabled in this PR.


---

PR #502, opened 2026-07-19
fix: remove CHECK constraints that conflict with MariaDB FK cascade actions

silasfelinus:
### Task
CI Janitor Todo #444: "Fix red CI: Kind Robots Cypress Tests" (main-branch `cypress.yml` red, run [29671897968](https://github.com/silasfelinus/kind_robots/actions/runs/29671897968)).

### Root cause
`prisma/migrations/20260719031500_reaction_first_party_author_expand/migration.sql` adds a `CHECK` constraint (`Reaction_firstPartyAuthor_check`) on `authorBotId`/`authorCharacterId` in the **same** `ALTER TABLE` that gives those columns `ON DELETE SET NULL ON UPDATE CASCADE` foreign keys. MariaDB rejects that combination outright:

```
Error: P3018
A migration failed to apply. New migrations cannot be applied before the error is recovered from.
Migration name: 20260719031500_reaction_first_party_author_expand
Database error code: 1901
Database error:
Function or expression 'authorBotId' cannot be used in the CHECK clause of `Reaction_firstPartyAuthor_check`
```
(full build log: [Vercel deployment `dpl_2RW4esAGTQ1mVnffw1cFret9NMCR`](https://vercel.com/silasfelinus-projects/kind-robots/2RW4esAGTQ1mVnffw1cFret9NMCR), `errorCode: db_migration_failed`)

Because Prisma refuses *any* further migration once one has failed (P3018), **every production deploy since 2026-07-19T03:27Z has failed**, and `kind-robots.vercel.app` has been stuck serving commit `a75cb03` (03:32Z) ever since — that's why `cypress.yml`'s "Wait for deploy to go live" step times out on every subsequent run: production genuinely never advances, it's not a flaky wait.

I also found the exact same pattern already staged one commit later, in `prisma/migrations/20260719033500_review_draft_storage_expand/migration.sql` (`ReviewDraft_single_author_chk` against `authorBotId`/`authorCharacterId`, which also get `ON DELETE SET NULL` FKs). It hadn't been reached yet (blocked behind the first failure) but would have failed identically the moment it was, so I fixed both in this PR rather than trading one incident for a second one next cycle.

### What changed
- Removed `Reaction_firstPartyAuthor_check` from the Reaction migration — columns, indexes, and both FKs are unchanged.
- Removed `ReviewDraft_single_author_chk` from the ReviewDraft migration — `ReviewDraft_rating_range_chk` (no FK on `rating`, unaffected by this restriction) is unchanged.
- Updated both migration contract tests (`utils/scripts/verifyReactionFirstPartyAuthorExpand.ts`, `utils/scripts/verifyReviewDraftStorageExpand.ts`) to assert the CHECK is **absent** instead of present, with the same MariaDB-restriction rationale in a comment.
- Mutual exclusivity for Reaction stays enforced at the application layer via the already-existing `assertSingleFirstPartyReactionAuthor()` (`utils/reactions/firstPartyReactionAuthor.ts`, already covered by `verifyFirstPartyReactionAuthor.ts`). ReviewDraft has no write path yet (its Prisma schema model hasn't landed either — `prisma/schema.prisma` has no `authorBotId` at all yet, confirmed by search), so equivalent write-time enforcement is future work when that path is built — left a comment pointing at the precedent.

### How I verified
- Read the actual Vercel build log for the failed deployment (`errorCode: db_migration_failed`, MariaDB error 1901) rather than guessing from the GitHub Actions timeout alone.
- Confirmed via `mcp__Vercel__list_deployments` that no `target: production` deployment has succeeded since `a75cb03` (03:32Z) — every commit after it either got superseded before Vercel started its build, or (for `1094a111`) errored on this exact migration.
- Re-fetched both edited migration files after pushing to confirm plain-text content landed correctly (an earlier attempt at this accidentally wrote base64 as literal file content — caught immediately via a follow-up read, fixed in a second commit before this PR was opened).
- Manually traced both verify scripts' regex assertions against the new SQL text to confirm they'll pass under `npx tsx`.
- Could not run `contract-tests.yml` or `cypress.yml` myself pre-merge (no local kind_robots checkout in this session); relying on CI to confirm. `contract-tests.yml` is DB-free so it validates this fix directly; `cypress.yml` will stay red until the production DB migration-failure state is manually cleared (see below) — that failure is expected and not a sign this fix is wrong.

### Stakes
reversible — this only edits migration SQL that has never successfully applied anywhere (MariaDB enforces this restriction universally, so it would have failed identically in every environment that ran it), plus two test files. No schema.prisma changes, no data changes, no destructive DDL.

### ⚠️ Manual production step still required (I cannot do this — no DB credentials in this session)
Merging this PR does **not** by itself unstick production. Prisma's migration history on the production database still has `20260719031500_reaction_first_party_author_expand` recorded as failed, and `prisma migrate deploy` will keep refusing to proceed (P3018) until that's cleared. Whoever has `DATABASE_URL` access needs to:

1. Check the current state of the `Reaction` table on production: do `authorBotId`/`authorCharacterId` columns and their indexes already exist? (The first two statements in the original migration — `ADD COLUMN` x2 and `CREATE INDEX` x2 — ran as separate statements before the failing `ALTER TABLE ... ADD CONSTRAINT` and very likely succeeded before the failure.)
2. If columns/indexes exist but the FKs don't: manually run the two `ADD CONSTRAINT ... FOREIGN KEY` statements from the fixed migration via `prisma db execute`, then run `npx prisma migrate resolve --applied 20260719031500_reaction_first_party_author_expand`.
3. If nothing from the migration applied at all: `npx prisma migrate resolve --rolled-back 20260719031500_reaction_first_party_author_expand` is sufficient, and `prisma migrate deploy` will re-run the fixed migration from scratch on the next deploy.
4. Trigger a new deploy (or just push again) and confirm `prisma migrate deploy` reaches and passes the `ReviewDraft` migration too, then confirm `cypress.yml` goes green.

### Flags for Reviewer
- This is a Todo (CI Janitor #444), not a conductor roadmap task, and it's cross-repo (fix lives in `kind_robots`, tracked from `conductor`). No roadmap task to update.
- I'm marking the conductor Todo complete per `CI-JANITOR.md`'s "complete this Todo only after the relevant verification has passed **or the remaining gate is clearly documented**" — the remaining gate (step 1-4 above) is documented here and needs Silas or another DB-credentialed operator, not further agent work.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FC72cf7TJX6tAJs3rbcpxa)_


silasfelinus:
Investigated the Vercel status/comment above (`api-deployments-free-per-day`, "more than 100 ... try again in 24 hours"). This is a **separate, bigger blocker** than the migration bug this PR fixes, and it's outside what I can act on:

Vercel's free-tier plan caps deployments at 100/day. Given the commit rate on `main` today (dozens of pushes/hour from concurrent autonomous sessions — most of the `cypress.yml` runs in the last few hours were `cancelled`, only two production deploys actually completed since 02:12Z), this repo has burned through that cap for the day. That means **no new Vercel deployment — preview or production — can build at all right now**, independent of whether the migration fix in this PR is correct. It'll reset in ~24h, or requires upgrading to Vercel Pro (a billing decision I can't make).

This doesn't change anything about this PR: the migration fix is still correct and necessary (production was already stuck on the DB migration failure *before* the rate limit hit), and `contract-tests.yml` (the actual gate for this PR — DB-free, doesn't touch Vercel) is still running/should still pass. I've flagged the rate limit to Silas directly since it needs a human decision either way.

---

PR #919, opened 2026-07-23
WonderLab: probe final commentary recovery state

silasfelinus:
Adds a temporary, read-only production probe for the five final commentary manifests.

The probe inspects all 86 rows and classifies each as:

- revised text already live
- still on the exact guarded old hash
- unexpected mismatch or read error

It uses the existing authenticated preflight boundary and performs no ReviewDraft or Reaction mutation. The purpose is to determine the exact production state after #918 even if the recovery run failed before publishing its evidence branch.

This PR should be closed without merge after the probe result is captured unless the workflow itself proves useful enough to retain.

### DIFF
--- a/.github/workflows/wonderlab-voice-polish-probe.yml
+++ b/.github/workflows/wonderlab-voice-polish-probe.yml
@@ -14,11 +14,12 @@ jobs:
   probe:
     name: Classify final commentary production state
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 25
     env:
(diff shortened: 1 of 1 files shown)


silasfelinus:
Probe purpose completed.

Authenticated production evidence from run `30054318537` confirmed:

- final manifests 032–036: **86/86 revised rows live**
- pending old hashes: **0**
- unexpected mismatches/read errors: **0**
- full published corpus: **706/706**
- draft range: **#1–#714**
- highest draft ID: **#714**

This workflow was intentionally temporary and read-only. Closing without merge.

PR #974, opened 2026-07-25
Synchronize Character Facets inside server mutations

silasfelinus:
## Problem

Character saves were monkey-patched in a client startup plugin. The Character API committed first, then the browser made a separate `/characters/:id/facets` request. A catalog/network/client failure could therefore leave the Character saved but its canonical assignments stale or missing, and non-Builder API callers did not receive automatic assignment synchronization at all.

## Change

- adds a server-side `syncCharacterFacetsInTransaction` helper
- resolves canonical aliases for genre, species, class, alignment, personality, backstory, quirks, and role
- enforces the allowed taxonomy for each Character field
- respects active/public/owned Facet visibility
- splits multi-valued personality and quirk fields consistently
- creates Character and CharacterFacet rows in one transaction
- updates Character and replaces CharacterFacet rows in one transaction
- marks automatic assignments with source `CHARACTER_MUTATION`
- removes the `characterStore.saveCharacter` monkey-patch and its Character-store import from the client plugin
- keeps the explicit Character Facet PUT endpoint for deliberate/manual assignment workflows
- extends the Facet cutover contract to enforce server-side synchronization and forbid the save wrapper from returning

## Result

Every Character mutation path—not only the Builder—produces canonical CharacterFacet assignments atomically with the Character row.

### DIFF
--- a/plugins/20.facet-catalog.client.ts
+++ b/plugins/20.facet-catalog.client.ts
@@ -3,7 +3,6 @@ import { defineNuxtPlugin } from '#app'
 import { ADVENTURE_CARDS } from '@/stores/helpers/adventureCards'
 import type { BuilderChoice } from '@/stores/helpers/builderCards'
 import { ensureBuildersRegistered } from '@/stores/registerBuilderStore'
-import { useCharacterStore } from '@/stores/characterStore'
 import {
   CHARACTER_FIELD_TAXONOMIES,
   useFacetCatalogStore,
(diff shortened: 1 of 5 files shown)


silasfelinus:
Reviewer pass (scheduled conductor sweep): held this PR — the `TypeScript` check is red (exit code 2 on `npm run test` / `vue-tsc`). The other four checks (`verify`, `Contract verifiers`, `facet-catalog`, `GitGuardian`) are green.

The diagnostics only landed in the uploaded `typescript-diagnostics` artifact (job [30174786154](https://github.com/silasfelinus/kind_robots/actions/runs/30174786154/job/89721708496)), not in the inline log, so I couldn't read the specific error from here. Please check that artifact and push a fix — will re-review once TypeScript is green.

---

PR #1146, opened 2026-07-29
model-builder: guard commitItem against a cancelled run

silasfelinus:
## Summary

- `commitItem()` in `stores/modelBuilderStore.ts` awaits a real network round-trip (the commit POST, up to 60s) but, unlike its siblings `generateItemAsset` and `pollAsyncArtJob`, never checked `cancelledRunIds` afterward.
- The commit POST durably creates/links/promotes the target server-side regardless of what happens next (same as `generateItemAsset`'s art render — that can't be undone from here), but without the guard, a commit that resolves after the user cancels the run still reattaches the detached item locally and pops a misleading "committed" success toast for a run they already told the app to abandon via `cancelRun()`.
- Fixed by capturing `runId` from `state.run` at the top of `commitItem` and checking `cancelledRunIds.has(runId)` right after the commit POST resolves (both success and catch branches), mirroring the existing guard already used by `generateItemAsset`/`pollAsyncArtJob`.
- Added a narrow textual regression checker (`verifyModelBuilderCommitCancelledRunGuard.ts`) plus its self-test, following the exact pattern of the existing `verifyModelBuilderCancelledRunGuard.ts` (which covers the sibling `pollAsyncArtJob` guard), and wired both into `contract-tests.yml` next to the Model Builder completion-gate check.

This is a continuation of conductor's `model-builder/t-029` recurring "polish and upgrade" task — each cycle re-reads the full model-builder surface for a genuinely new, previously-unfixed bug.

## Test plan

- [x] `npx eslint stores/modelBuilderStore.ts` — 2 pre-existing `no-empty` errors confirmed unrelated via `git stash` (present with and without this diff)
- [x] `npx prettier --check stores/modelBuilderStore.ts` — pre-existing drift confirmed unrelated via `git stash`
- [x] `node ./node_modules/vue-tsc/bin/vue-tsc.js --noEmit -p tsconfig.json` — exit 0
- [x] `npm run test:model-builder-commit-cancelled-run-guard-selftest` — passes (flags the pre-fix shape, clears the fixed shape, flags the function being absent)
- [x] `npm run test:model-builder-commit-cancelled-run-guard` — passes against the real store file

---
_Generated by [Claude Code](https://claude.ai/code/session_01WD5VWitMn7hUxJxS2cEyU8)_

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -145,6 +145,12 @@ jobs:
       - name: Model Builder completion-gate contract
         run: npm run test:model-builder-completion-gate
 
+      - name: Model Builder commit cancelled-run guard checker self-test
+        run: npm run test:model-builder-commit-cancelled-run-guard-selftest
+
+      - name: Model Builder commit cancelled-run guard contract
+        run: npm run test:model-builder-commit-cancelled-run-guard
+
       - name: Kontext mask forwarding contract
         run: npm run test:kontext-mask-forwarding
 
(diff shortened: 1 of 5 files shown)


silasfelinus:
REVIEWING: claude-reviewer-2026-07-29T110331-sweep at 2026-07-29T11:03:31Z

---

PR #1149, opened 2026-07-29
Give missing-art suggestions real project context

silasfelinus:
## Root cause
The automatic missing-image path did not use `/api/suggest`. The browser sent a thin DOM report directly to `/api/conductor/art-request`, whose separate prompt writer attempted to infer the artwork. For project tiles, `nearbyText()` often climbed to the entire gallery `<main>` because each tile is a `<button>`, so context could blend several unrelated projects.

## What changed
- Added a dedicated `art-asset` sheet to the shared `/api/suggest` registry.
- The missing-image reporter now calls `/api/suggest` first and uses the currently selected text server, matching the rest of the frontend suggestion architecture.
- Sends structured context for the exact asset: source, role, variant, dimensions, class, optional art-direction attributes, page metadata, and local component text.
- Resolves project-specific context from the Project and Conductor stores: title, slug, kind, status, goal, description, flavor text, notes from Silas, milestones, and representative unfinished tasks.
- Fixed local text capture so a project `<button>` is treated as the context boundary instead of collecting text from the whole gallery.
- The art sheet produces format-aware icon/card/hero prompts with concrete composition, lighting, palette, materials, texture, and mood.
- Software projects explicitly prefer tools, interfaces, machines, environments, diagrams, or symbolic objects rather than invented portraits, faces, mascots, or default robots.
- The generated prompt is passed to `/api/conductor/art-request`; the existing direct writer remains only as a fallback if `/api/suggest` fails.

## Validation
- Added `verifyArtAssetSuggest.ts` covering sheet registration, structured project context, hero/icon format instructions, and anti-portrait/default-robot rules.
- TypeScript and repository contract checks should validate the client/store integration.

### DIFF
--- a/plugins/missing-image-reporter.client.ts
+++ b/plugins/missing-image-reporter.client.ts
@@ -1,13 +1,35 @@
 import { watch } from 'vue'
 import { useRoute } from 'vue-router'
+import { useConductorStore } from '@/stores/conductorStore'
+import { useProjectStore } from '@/stores/projectStore'
+import { useServerStore } from '@/stores/serverStore'
 import { useUserStore } from '@/stores/userStore'
 import { performFetch } from '@/stores/utils'
+
+type ProjectArtContext = {
+  id?: number
+  slug?: string
+  title?: string
+  kind?: string
+  status?: string
+  goal?: string
+  description?: string
+  flavorText?: string
+  notes?: string
+  milestones?: string[]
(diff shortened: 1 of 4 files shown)


silasfelinus:
REVIEWING: claude-reviewer-2026-07-29T110331-sweep at 2026-07-29T11:03:31Z

---

PR #1208, opened 2026-07-31
refactor(startup): let the app own the intro, not the pre-hydration shell

silasfelinus:
Implements Silas' diagnosis: *"This all started because we moved all the loading effects into the pre-startup loop. What if we don't start this until we've officially loaded nuxt and app.vue? We can't block everything if we've already started it, right?"*

Right on both counts.

## Why the incremental fixes kept failing

The launch screen existed **twice** — once rendered by server plugins before hydration (animated background, loading messages, an interactive control tray, a click bridge with an action queue, lifecycle tracing, a JS watchdog), and once by Vue — with a handoff between them. Every failure this week lived in that seam, never in the intro itself:

| symptom | seam |
|---|---|
| tray responds to `:hover` but not clicks | its buttons forwarded to a Vue component that hadn't mounted |
| screen becomes unrecoverable after "Pause & explore" | that action disarmed every watchdog, by design |
| sequence restarts and never ends | watchdog fired at 9s and removed the classes the CSS hard stop was gated on; hydration at 13.8s then started a second sequence with no safety net |

Each fix for one seam opened another, because the architecture required two implementations to agree while one of them might never arrive.

## The change

The server now contributes exactly one thing: a boot cover that hides the server-rendered site and **releases itself through a CSS animation**. Vue adds `kr-app-ready` on mount to retire it early, but that's an optimisation — if it never happens, the cover still clears.

A CSS animation cannot be starved by a busy main thread, blocked hydration, or a chunk that failed to load. There is no state in which it traps the user.

Everything else — animation, rotating messages, control tray — is owned outright by Vue and starts only once the app is running, where normal component lifecycle applies and **the site is always underneath**.

**Removed:** `server/plugins/startup-loader-shell.ts`; the entire shell script and its CSS; `plugins/startup-composition.client.ts` (its only job was coordinating with the shell); the bridge / action-queue / handoff plumbing in `startup-animation.vue`; the sessionStorage handoff state in `startupLaunch.ts`. The intro overlay now paints its own backdrop instead of relying on a server-rendered black base.

Net: **−1 server plugin, −1 client plugin, −3 watchdogs, −1 bridge, −1 action queue, −5 coordination classes.**

## Verification

The property that matters, tested with **JavaScript fully disabled** — the worst case the old shell could not survive:

```
t=0s     opacity=1  visibility=visible
t=6.8s   opacity=0  visibility=hidden
```

Served HTML confirms the old machinery is gone:

```
PASS  boot cover markup            PASS  prehydrate loader GONE
PASS  CSS self-release keyframe    PASS  watchdog GONE
PASS  preload href interpolated    PASS  action queue GONE
PASS  kr-full-startup GONE         PASS  startup trace beacon GONE
```

`verify-startup-handoff.mjs` rewritten to assert the new invariant (server cover carries no JavaScript; no pre-hydration launch experience may return) and passes. Workflow path filters updated — two files it watched no longer exist, so it would have silently stopped running. ESLint: 2 errors, both pre-existing on `main`.

## Behaviour and scope

Unchanged: refresh skips the intro, fresh navigation plays it, the dashboard refresh button forces it.

**Still open:** this does not make hydration faster — it was measured at ~13.8s. It removes every way a slow hydration turns into a launch screen with no exit, but if hydration is genuinely that slow you'll see the boot cover for up to 6s, then the intro. That's worth chasing separately, and it's now a plain performance problem rather than a deadlock.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N9HNdfh9HAYppqpW51BQBH)_


silasfelinus:
REVIEWING: conductor-agent-20260731T1604Z-pr1208 at 2026-07-31T16:04:22Z

---

silasfelinus:
**Review: found a real regression, holding off on merge.**

CI shows `TypeScript Type Check` red on this PR's head (`0f3379f`). I compared the uploaded `typescript-diagnostics` artifacts for this commit vs. `main` tip (`0fd6a58`):

- `main` has 2 pre-existing errors (both `videoStore` export mismatches in `video-lora-picker.vue` / `video-generator.vue`) — unrelated to this PR, matches the PR description's "ESLint: 2 errors, both pre-existing" note.
- This PR's head has those same 2 **plus one new one**:
  ```
  components/screenfx/startup-animation.vue(14,19): error TS2339: Property 'handleEffectReady' does not exist on type ...
  ```

Looking at the diff: `startup-animation.vue` still has `<Suspense ... @resolve="handleEffectReady">` in the template (line 14), but the refactor removed the `handleEffectReady` function from the `<script setup>` block along with the rest of the bridge/handoff plumbing. `effectReady` is now only ever set to `false` (in `setEffect`) and never back to `true` — so `startup-animation__stage--ready` never applies, and the stage's `opacity: 0 → 1` transition (the CSS rule this whole refactor leans on) never fires for the animation effect itself. This looks like an unintended side effect of removing the bridge, not a deliberate behavior change — worth confirming, since it would mean the startup animation stage never becomes visible after this lands.

Suggested fix — restore a small handler that just sets the ref (no bridge needed anymore):
```ts
function handleEffectReady(): void {
  effectReady.value = true
}
```

Everything else in this PR looks solid (the boot-cover approach, the JS-disabled verification, the removed coordination plumbing) — this is the one thing I'd fix before merging. Not merging as-is; leaving at open for a fix-and-repush.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DwtQ7B4xLs2KGN2z5RcDA5)_

---

silasfelinus:
Pushed `96b91ab` restoring `handleEffectReady` (see the review comment above for the diagnosis) — verified locally with `vue-tsc` (only the 2 pre-existing `videoStore` errors remain) and ESLint clean. Waiting on CI to confirm, then will merge.

Separately: this branch's fix is also my leading hypothesis for the currently-red main-branch Cypress CI (5 open `ci-janitor` Todos, all pointing at the *same* hang — `cypress/e2e/api/art-image-collection-connections.cy.ts`'s `before()` hook does `cy.visit()` against the live homepage, and has hung for the full 30-minute job timeout on every scheduled run since 2026-07-31T00:06Z). The commits landing right before that window are a run of `fix(startup): ...` patches — the same incremental fixes this PR's description says "kept failing" for exactly this failure mode (a hydration-dependent launch screen that can enter an unrecoverable restart loop). Plan: merge this once green, then trigger `cypress.yml` manually to check whether the hang clears.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DwtQ7B4xLs2KGN2z5RcDA5)_

---

PR #1212, opened 2026-07-31
Fix startup effect layering and persistent logo haze

silasfelinus:
## What changed

- split startup into a fixed black base at z-48, Screen FX at z-49, and a transparent loader foreground at z-50
- stop the loader overlay from painting black over the animation
- replace the nearly hard-edged logo mask with a broader, softly feathered, slightly offset elliptical haze
- use the same mask for the brief server-rendered refresh cameo and the Vue intro
- extend the startup handoff contract to lock in the three-layer invariant and matching masks

## Root cause

The loader lived inside a z-50 stacking context and painted its own opaque black full-screen background. The Screen FX stage was z-49, so it was correctly above the app but still behind the loader's black paint. The existing mask also remained almost fully opaque through 70% of the image, making most of the WebP read as a square rather than a diffuse shape.

## Expected behavior

Black remains the bottom startup layer, the animation is visible above it, and the WebP/messages/spinner/controls remain above the animation. The logo has a persistent soft organic edge during both the refresh cameo and full loader.

## Validation

- Startup Handoff Contract passed
- `git diff --check` passed
- full local TypeScript could not run because this clean checkout's dependency install hit the runtime npm cache restriction; GitHub Actions is the authoritative full check.

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -11,6 +11,12 @@
       aria-hidden="true"
     />
 
+    <div
+      v-if="showLoader"
+      class="pointer-events-none fixed inset-0 z-48 bg-black"
+      aria-hidden="true"
+    />
+
     <ClientOnly>
       <div v-if="showLoader" class="pointer-events-none fixed inset-0 z-50">
         <kind-loader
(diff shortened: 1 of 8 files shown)


silasfelinus:
REVIEWING: conductor-quirky-curie-2026-07-31T20:26:00Z at 2026-07-31T20:26:00Z

---

PR #1213, opened 2026-07-31
Fix conductor→Projects progress sync and build the Da Vinci narration layer

silasfelinus:
Two things Silas asked for this session: find out why the Projects board isn't 1:1 with conductor, and take Da Vinci start to finish.

## 1. The Projects board was wrong for 30 of 44 projects

Not sync lag — `/api/conductor/projects` reads conductor's `main` live on every request. Two independent bugs, one stacked on the other.

**The formula.** `computeProgress` scored from milestone status alone, while conductor's `scripts/build_status.py` falls back to the task-completion ratio when milestones carry no signal. Agents close tasks reliably but rarely flip `milestone.status`, so any project whose milestones were all left `not-started` rendered as **0%**. davinci, newsfeed, challenge-center, ecosystem-map, engagement and global-ui were all 100% complete and all showing zero; conductor itself showed 0% against a real 96.8%.

**The parser, which was worse.** Fixing the formula and then verifying against all 44 *live* roadmaps instead of synthetic fixtures showed 30 still disagreeing. `parseRoadmapYaml` was a hand-rolled line-regex reader that understood exactly one spelling of YAML. It never threw — it returned partial data, which reads downstream as a smaller, less-complete project:

| shape | real example | parsed |
|---|---|---|
| blank lines between sequence items | davinci | 1 of 4 milestones, 1 of 16 tasks |
| sequences at column 0 (`yq`/PyYAML style) | lora-ingestion | 0 of 6 tasks |
| double-quoted multi-line scalar | challenge-center | 0 of 20 tasks |
| " | conductor | 26 of 95 tasks |

Replaced with the `yaml` package — already in the tree via Nuxt, now an explicit dependency — plus mappers that keep the exported `ConductorTask`/`ConductorMilestone` shapes and accept either snake_case or camelCase spellings. A malformed roadmap yields an empty parse rather than throwing, matching how the caller already treats an unreadable roadmap.

Also fixes a weight-defaulting divergence: `Number(weight) || 10` turned an explicit `weight: 0` into 10, where build_status.py's `m.get("weight", 10)` defaults only on an absent key. brainstorm's sole milestone is `weight: 0, status: in-progress`, so this repo scored it 50% against conductor's 0%.

**All 44 live roadmaps now parse to identical counts and identical progress.** 30 mismatches to zero.

`verifyConductorProgressParity.ts` asserts the endpoint against a direct transcription of build_status.py's formula, with a fixture per real-world shape above. Worth noting the suite caught a bug in its own reference transcription — it had repeated the `|| 10` mistake.

## 2. Da Vinci narration layer

The play-loop API and 1024 seeded endings have been live since t-013/t-008. `t-015` specced the narration layer and explicitly declined to file the build task, so davinci was flipped `finished` with nothing generating the prose a player actually reads. This builds it, per `projects/davinci/docs/narration-layer-spec.md`:

- **`server/utils/davinciNarration.ts`** — strict `json_schema` request following the `requestOpenAiReview`/`validateGeneratedPayload` pattern already live in `wonderLabReviewDraftGenerator.ts`, 20s `AbortController` budget (tighter than WonderLab's 30s because the player is waiting on this call), one retry, then a visible error rather than a synthetic chapter. Validation enforces every bound the model could violate: the `DAVINCI_DIMENSIONS` allowlist, the `[-2,2]` per-choice clamp, the 2–4 choice band, prose word bounds. The allowlist matters most — `recordLifeChoice` accepts arbitrary stat keys by design, so this is the only thing between an invented eleventh dimension and a junk `LifeStat` row.
- **`POST /api/davinci/runs/:id/narrate`** — writes nothing. Owner-only, 409 on a non-ACTIVE run. The client posts the chosen option's validated effects to `/choices` as a separate step, so all state writes and outcome math stay in `server/utils/davinci.ts`.
- **`davinci-page.vue`** — plays narrated chapters, end-the-run action gated on chapter ≥ 3. The curated pool from t-014 is kept but is no longer the default; it's offered as an explicit labelled choice after a narration failure, so a player always knows which mode they're in.

Two deviations from the spec, both forced by OpenAI strict mode (recorded in the spec doc): `effects` is all-ten-dimensions-required typed nullable-integer rather than optional keys, and range/count bounds are enforced server-side rather than via `minimum`/`maxItems`. The spec already required that re-validation, so nothing is weakened.

## 3. Drive-by: `main` was failing `vue-tsc`

`VideoEngine`/`VideoOutputFormat` were imported from `stores/videoStore`, which doesn't re-export them — TS2459 on both call sites since `4361568`. Pointed the two consumers at the canonical `@/utils/videoPresets` rather than re-exporting, which would have reintroduced the duplicate auto-import that commit was fixing.

## Verification

- `vue-tsc --noEmit` — 0 errors (was 2 on `main`)
- `eslint` clean on every changed file
- `npm run test:conductor-progress-parity` — 25 checks green, wired into `contract-tests.yml`
- `npm run test:davinci-narration` — 35 checks green, wired into `contract-tests.yml`
- All 44 live conductor roadmaps cross-checked against PyYAML + `build_status.py`: zero parse or progress mismatches
- `npm ci` clean from the updated lock

No live model call was made from this sandbox, so the first real end-to-end narration happens on the preview deploy. The failure path is a visible retry, not a broken run, so it's safe to land unproven.

Paired with conductor PR (davinci reopened, `t-016` filed and closed, two invalid roadmaps repaired).

https://claude.ai/code/session_015fV9RtqejrTwrMjv1zABft

---
_Generated by [Claude Code](https://claude.ai/code/session_015fV9RtqejrTwrMjv1zABft)_


silasfelinus:
REVIEWING: conductor-quirky-curie-2026-07-31T20:26:00Z at 2026-07-31T20:26:00Z

---

PR #1214, opened 2026-07-31
fix(facet-catalog): satisfy Adventure and Scenario builder-hydration contracts

silasfelinus:
### What changed
Two facet-catalog cutover contract checks (part of the required `facet-catalog` CI job) have been failing on `main` since PR #1211:

- `verifyFacetCatalogCutover.ts` requires the literal text `hydrateAdventureBuilder` in `plugins/20.facet-catalog.client.ts`.
- `verifyScenarioGenreFacetCutover.ts` requires both `hydrateScenarioBuilder` and the literal call `hydrateBuilderCards(SCENARIO_CARDS, catalog)`.

The shipped plugin only had one shared `hydrateBuilderCards(cards, catalog)` function, called for both `ADVENTURE_CARDS` and `SCENARIO_CARDS`, satisfying neither contract.

Added two thin named wrapper functions — `hydrateAdventureBuilder()` and `hydrateScenarioBuilder()` — that delegate to the existing shared `hydrateBuilderCards` implementation, satisfying both contracts without duplicating hydration logic.

### How I verified
- `npx tsx utils/scripts/verifyFacetCatalogCutover.ts` — now passes (was failing on `main`)
- `npx tsx utils/scripts/verifyScenarioGenreFacetCutover.ts` — now passes (was also failing on `main`, previously masked because the CI job dies on the first failing step)
- `npx tsx utils/scripts/verifyFacetCanonicalMerges.ts` / `verifyDailyDreamFacets.ts` — pass
- `vue-tsc --noEmit` — no new errors introduced (2 pre-existing, unrelated `TS2459` errors on `videoStore` remain; already fixed in open PR #1213/#1212)
- `eslint` clean on the changed file

### Stakes
reversible

### Flags for Reviewer
- Discovered while reviewing PR #1213/#1212, both blocked by this pre-existing `main` failure unrelated to their own diffs. PR #1212 independently "fixed" this by weakening the contract test back to the old `hydrateBuilderCards` name instead of fixing the code — this PR takes the other direction (fix the code to match the contract's intent) since the contract's naming looks deliberate, not accidental.
- Once merged, PR #1213 and #1212 will need their `facet-catalog` check re-run to pick up this fix (and #1212's own edit to `verifyFacetCatalogCutover.ts`/`verifyScenarioGenreFacetCutover.ts` will need to be dropped/reconciled, since it reverts this fix).

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/plugins/20.facet-catalog.client.ts
+++ b/plugins/20.facet-catalog.client.ts
@@ -253,8 +253,12 @@ export default defineNuxtPlugin(() => {
         import('@/stores/registerBuilderStore'),
       ])
 
-      hydrateBuilderCards(ADVENTURE_CARDS, catalog)
-      hydrateBuilderCards(SCENARIO_CARDS, catalog)
+      const hydrateAdventureBuilder = () =>
+        hydrateBuilderCards(ADVENTURE_CARDS, catalog)
+      const hydrateScenarioBuilder = () =>
+        hydrateBuilderCards(SCENARIO_CARDS, catalog)
+      hydrateAdventureBuilder()
+      hydrateScenarioBuilder()
       hydrateBotBuilder(BOT_CARDS, catalog)
       hydrateArtBuilder(ART_CARDS, catalog)
       hydrateSystemBuilder(DREAM_CARDS, catalog)


silasfelinus:
Closing unmerged: superseded. PR #1212 merged to `main` (2193e22) while this was open, and independently fixed the same `facet-catalog` contract failure by relaxing the test to match the shipped `hydrateBuilderCards` name instead of renaming the code — the opposite direction from this PR, but `main` is green with it (`verifyFacetCatalogCutover.ts` and `verifyScenarioGenreFacetCutover.ts` both pass on current `main`). No need to also land this competing version.

---

PR #1223, opened 2026-07-31
fix(model-builder): block duplicate same-item auto-builds

silasfelinus:
### Task
Conductor `model-builder/t-029` recurring Model Builder polish cycle.

### Root cause
`autoBuildItem()` is reachable from the item-level Auto action, `batchAutoBuild()`, and `autoBuildRun()`. Its `autoBuildingItemSingleton` protects a claim from being released by a *different* item id, but two concurrent calls for the **same** item id share the same owner value. Both calls could therefore run at once and duplicate drafting, rendering, or commit work.

This is the underlying reentrancy gap identified while fixing the approved-image overwrite race in #1221. That prior fix safely discards a late render; this one prevents the duplicate same-item auto-build from starting in the first place.

### Change
- `autoBuildItem()` returns `false` immediately when `state.autoBuildingItemId === item.id`, before claiming the singleton.
- Extended the existing `verifyModelBuilderAutoBuildDraftGate` regression guard to require that entry guard before the singleton claim.
- Extended its self-test with an isolated fixture where draft gating is correct but same-item reentrancy protection is missing.

### Verification
The final branch diff is limited to the store and the existing auto-build guard + self-test. TypeScript and repository contract workflows are the verification gate for the PR head.

### Stakes
Reversible.

### Kaizen
None. This closes the concrete follow-up left by #1221 rather than spawning another sibling workaround.


silasfelinus:
REVIEWING: 2026-07-31T231300Z-model-builder-t029-health2 at 2026-07-31T23:23:00Z

PR #1249, opened 2026-08-01
Replace obsolete alpha achievements with the active 25

silasfelinus:
## Outcome

Leaves production with exactly the intended 25 active achievements.

## Production evidence

Production contained the 25 canonical rows plus five historical alpha definitions whose trigger codes did not match the catalog:

- `pitchmaster`
- `Artist`
- `matchmaker`
- `button`
- `milestone`

Per the project owner's alpha-stage direction, this change removes those five definitions and their disposable award records instead of preserving or translating the history. Optional Life-system links are nulled first so unrelated Life data remains valid.

The public API returns active definitions only, and all 25 canonical rows are explicitly active. The production seed reports both stored and active totals, which must converge to 25.


silasfelinus:
Holding this at `needs-human` rather than merging — flagging for Silas, not blocking on a defect.

This PR's `deleteRetiredAchievements()` permanently deletes `Achievement` rows for 5 retired trigger codes (`pitchmaster`, `Artist`, `matchmaker`, `button`, `milestone`) **and their `AchievementRecord` rows** — i.e. any real user's actual earned-achievement history for those five, not just unused catalog definitions. It also nulls out `LifeAchievementUnlock.achievementRecordId`, `LifeAchievement.achievementId`, and `LifeEnding.achievementId` first to satisfy FK constraints before the delete.

The PR body attributes this to "the project owner's alpha-stage direction," but I couldn't find that direction recorded anywhere in conductor's tracked files (`TALKBACK.md`, any `projects/*/roadmap.yaml`, or `projects/*/TALKBACK.md` — searched for "alpha-stage" and each retired trigger code, no hits). Per this repo's standing operating rules, deleting real data records is an irreversible action that needs Silas's explicit confirmation before an agent merges it, even when the reasoning in the PR description looks sound — I have no way to independently confirm "disposable" is accurate for these specific records (a live DB query would show whether any of the 5 actually have earned records, which would make this concrete rather than hypothetical).

Not a rejection — the code otherwise looks reasonable (schema-safe FK nulling before delete, guard script updated, `isActive` filtering added to the public API and 21 formerly-inactive achievements flipped active to match). If Silas confirms the deletion is intended, this is safe to merge as-is.

---

silasfelinus:
Follow-up for the audit trail: Silas had explicitly directed in the owning session that achievement history did not need preservation during alpha, noting that only one person had a single earned record and that inactive definitions/records could be removed to reach the intended design. That direction authorized the cleanup merged here; the later needs-human flag was raised because the approval was not visible in repository files or PR comments.

PR #1258, opened 2026-08-01
Separate iconPath (a real logo) from icon (a name), plus the art backfill script

silasfelinus:
Follow-up to #1252, which merged before these three commits landed. Rebased onto the new `main`.

## The correction

Silas, 2026-08-01, verbatim:

> *"there is a strong difference between icon and iconPath. An icon is a simple string meant to name an existing simple icon, preferably in our kind robots system. An iconPath is a string but it leads to an actual path that points to a fully developed logo."*

#1252 got half of this right. Adding `icon` (the **name**) was correct, but it stopped there — so there was still nowhere to put a generated logo. Worse, it wrote *"there is no icon art slot"* into the registry comment, entrenching the mistake.

The codebase already shows the cost of blurring the two:

```ts
// server/api/conductor/project-art-complete.post.ts:118
const artImageField = field === 'imagePath' ? 'iconPath' : field
```

`iconPath` existed **only on `ArtImage`**. Entities overload `imagePath` for the logo job — which is why a Bot's 1024×1024 avatar also has to serve as its ~40px grid icon. Small cells have been downscaling a portrait.

### What changed

- **New migration `20260801233000_entity_icon_path`** — 4 `ADD COLUMN`, nothing else. Written as a *separate* migration rather than editing the one from #1252, since that may already have been applied to a preview.
- **`iconPath` is now a real art slot at 256×256** for bot, character, scenario, reward — so the Icon tab generates a purpose-made logo instead of reusing the portrait.
- **`icon` stays a name and is never generated.** The registry comment now states the distinction instead of denying it.
- **`slotData` handles `iconPath`.** Without this, an Icon render would have fallen through to the primary branch and overwritten the avatar — *the same bug class #1252 fixed for card/hero.* Two occurrences in two commits is a signal, so I noted that this branch wants to be data-driven from `ENTITY_FIELDS` rather than a ternary chain, and folded it into `t-009`.

`project` keeps labelling its `imagePath` "Icon" for backward compatibility; migrating it onto a real `iconPath` column is deliberately left for `t-009` so this stays additive.

## The backfill script

`utils/scripts/backfillEntityCardHeroArt.ts` — now covers all three secondary slots (Icon 256², Card 512×768, Hero 1280×720). Dry-run by default, hard-capped by `--limit`, idempotent, and `--apply` refuses without credentials.

**What a zero-cost probe against production actually proved:**

```
POST /api/art/enqueue  {entityType:'bot', field:'cardPath'}
→ 400 "Invalid bot image field."
```

That's the *right* failure, and it establishes three things: `authAndGate` runs before field validation, so `Bearer KR_API_TOKEN` authenticates and the mana gate passes; **no mana was charged**, since `gate.commit` only fires on success; and the payload shape is accepted up to the field name.

**Still not usable until deploy.** As of this writing the probe still rejects `cardPath`/`heroPath`/`iconPath` on production — #1252's migration hasn't propagated yet. There's a one-line way to check: re-run the probe, and when it stops saying *"Invalid … image field"*, the backfill is live.

**Not proven:** the Prisma candidate query. It needs direct MySQL on `:5544`, which is reachable from Silas's machine but not from this sandbox (HTTPS-only egress). Stated in the script header rather than glossed.

## Verification

Re-run after rebasing onto the new `main`: `vue-tsc --noEmit` ✅ · `prisma validate` ✅ · migration re-diffed against the new base and **byte-identical** ✅ · `verifyEntityArtManager` ✅ · `verifyFacetGallery` ✅ · `verifyFacetArtGeneration` ✅ · `test:layout-contract` ✅ · eslint/prettier clean

---


silasfelinus:
REVIEWING: hourly-health-2026-08-01T14:13:09Z-iconpath at 2026-08-01T14:13:09Z

PR #1261, opened 2026-08-01
Flip app.vue's NuxtPage wrapper to overflow-hidden; backfill per-page scroll ownership

silasfelinus:
### Task
conductor: interface-vision/t-003b — flip `app.vue`'s `NuxtPage` wrapper from `overflow-y-auto` to `overflow-hidden` ("the page owns its scroll") and backfill per-page scroll ownership first, per the roadmap note split out of t-003 (kind_robots PR #1255).

### What changed
- **`app.vue`**: the shell wrapper around `NuxtPage` flips from `overflow-y-auto` to `overflow-hidden`.
- **`pages/[...slug].vue`**: `content-host` (the shared container for every MDC-routed page) now owns scroll itself — this single change covers the bulk of the audit, since about/giving/privacy/sponsor/mermaids/registration-form/etc. all render through it.
- **10 `components/pages/*.vue`** files that render outside `content-host` or independently of it (appmaker, click/match-leaderboard, super-form, ...).
- **`login-page.vue` and `welcome-page.vue`**: these already had `overflow-y-auto`, but on an unbounded `min-h-full`/`min-h-screen` container, so it never actually engaged — same practical bug as "no scroll region," found via a second pass checking for that anti-pattern and fixed the same way.
- The **WonderLab admin pages**, **play/challenges pages**, **coloring-page**, **music-mentor**, and **build-bench** — real Nuxt file routes rendered directly (not through `content-host`) that had no scroll container.
- **`utils/scripts/verifyLayoutContract.ts`**: adds a `zero-scroll` rule (a page component with no scroll region of its own) following the file's existing ratchet-baseline pattern — this is the kaizen suggestion from PR #1255, so a future page that loses its only scroll container fails CI instead of silently clipping content.

### Deliberately left unfixed (documented in the ratchet baseline, not silently skipped)
- 11 `components/conductor/*-page.vue` wrappers + `coloring-book-page.vue` already inherit scroll from the shared `<ProjectFrontPage>` shell (its root div already has `overflow-y-auto`) — no fix needed.
- `conductor-manager.vue` (router switching between sub-apps) and `conductor-art-gallery.vue` (embedded widget sized by its parent) correctly delegate scroll to children/parent by design.
- `kaizen-popup.vue` is a `Teleport`-to-body modal, not page content.
- `plan-projects-grid.vue` is covered by the `content-host` fix (it's an MDC block, not a standalone route).
- `kind-prompts.vue`, `rebel-button.vue`, `social-page.vue` (broken `<kofi-link/>` reference) appear to be **orphaned/unreachable** — no route or content-file mount found for any of them. Flagging rather than guessing at a fix for dead code; worth a follow-up task to confirm and prune.

### How I verified
- `npm run test:layout-contract` — holds with no new violations. `no-viewport` improved by 7 as a side effect of removing several `min-h-screen`/`min-h-dvh` occurrences these edits touched anyway.
- `vue-tsc --noEmit` — exits clean, 0 errors.
- `eslint` on every touched file — only two pre-existing errors remain (`pages/[...slug].vue`'s multi-root template, `pages/admin/wonderlab-reviews.vue`'s unused-expression), both confirmed present on `main` before this change via `git stash` — unrelated to scroll, not introduced here.
- `prettier --check` flagged several of these files as already non-compliant *before* my edit (confirmed via `git stash`); I kept my diff scoped to the intended one-line class changes rather than bundling an unrelated full-file reformat into this PR.

### Stakes
reversible

### Flags for Reviewer
- No live browser/Vercel-preview pass was done in this session — verification is static (typecheck, lint, the layout-contract ratchet, and direct reading of each touched file's structure to confirm the scroll-container placement is correct). Worth a quick look at the preview deploy for a couple of the higher-traffic pages (about, privacy, WonderLab admin) before merge if you want visual confirmation.
- Per the roadmap task's own note, this pairs naturally with t-004 (kr-* primitives, one-header rule) since it touches the same files — I kept this PR scoped to scroll-ownership only.

### Kaizen suggestion
File a small follow-up task to check whether `kind-prompts.vue`, `rebel-button.vue`, `click-leaderboard.vue`, and `match-leaderboard.vue` are genuinely orphaned (no route/content mount references any of them, per grep across `content/`, `pages/`, and every `.vue`/`.ts` file) and prune them if so — dead code that also happens to violate the layout contract is confusing to future audits like this one.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MC5P4Eufh7qDQpXi23dr4G)_


silasfelinus:
REVIEWING: kr-hourly-2026-08-01T16:12:02Z-pr1261 at 2026-08-01T16:12:02Z

silasfelinus:
Blocking review finding: the branch is green, but scroll ownership is not composition-safe yet. `pages/[...slug].vue` now makes the shared `content-host` the scroll owner, while several components rendered inside that host also gain `h-full min-h-0 overflow-y-auto overscroll-contain`: `components/pages/about-page.vue`, `giving-page.vue`, `mermaids-page.vue`, `privacy-page.vue`, `registration-form.vue`, `sponsor-page.vue`, and `components/user/registration-page.vue` (with `welcome-page.vue` also retaining an inner `overflow-auto`). That creates nested scroll regions on the MDC routes, which is the exact layout failure this task is intended to remove. The verifier passes because it counts scroll regions file-by-file and cannot see parent/child composition.

Keep `content-host` as the sole scroll owner for MDC-rendered pages and revert the added child scroll classes there. Keep per-page scroll additions only for routes/components that genuinely bypass `content-host`, then rerun Layout Contract and TypeScript. A small regression assertion or explicit allow/deny list tying known MDC page components to the shared host would prevent this blind spot from returning.

I am not merging #1261 in its current form.

silasfelinus:
Good catch, confirmed and fixed in `f4c33d6`.

Verified via `content/*.md` grep: `about-page.vue`, `giving-page.vue`, `mermaids-page.vue`, `privacy-page.vue`, `sponsor-page.vue`, and `components/user/registration-page.vue` are each mounted directly inside `content-host` (`about.md`, `giving.md`, `mermaids.md`, `privacy.md`, `register.md`), so `content-host` owning scroll made their own added `h-full min-h-0 overflow-y-auto overscroll-contain` redundant and nested — reverted all six back to their pre-PR classes.

`registration-form.vue` also reverted: turns out it's not referenced by any route or content file at all (repo-wide grep), so it was never actually composed under `content-host` in practice — no fix needed there either.

`welcome-page.vue` I left as-is: it isn't mounted via `content-host` either (also unreferenced by any `content/*.md`), and its change only bounds its own outer wrapper so its *pre-existing* inner `overflow-auto` region can actually engage — a single-file parent/child pair, not the cross-component composition conflict you flagged.

Re-bootstrapped the `zero-scroll` ratchet baseline so these 7 files correctly land back in the accepted-exemptions list (same category as the `ProjectFrontPage`-wrapped conductor pages — scroll owned by a shared ancestor, not the file itself). `test:layout-contract` holds, `vue-tsc --noEmit` passes clean.

Your suggestion about a small regression assertion or an explicit allow/deny list tying known MDC page components to `content-host` is a good one — I don't want to scope-creep this PR further, so I'll leave that as a follow-up rather than bolt it on now, but happy to hear if you'd rather have it in this PR.

---

silasfelinus:
REVIEWING: claude-agentrun-2026-08-01T16:57:36Z-review-t003b at 2026-08-01T16:57:58Z

---

PR #1280, opened 2026-08-01
Fix two dead admin bypasses caused by isAdmin type assertions

silasfelinus:
## What

Two authorization checks were silently dead. Both were the same defect: code used `as { isAdmin?: boolean }` to **assert** a property onto a value that never carried it, so the flag read permanently `false` and the admin branch behind it never executed. `as` asserts a shape rather than reading one — which is exactly why the typechecker stayed quiet about a constant.

### 1. `server/api/art/enqueue.post.ts:282`

```ts
const isAdmin = Boolean((gate.user as { isAdmin?: boolean }).isAdmin)
```

`comfyGate.authAndGate` destructures a full `AuthUser` from `requireMachineUser` — carrying `Role` **and** `isAdmin` — then returns `user: { id: user.id }`, discarding the flag it had already resolved.

Consequence: the admin bypass in `assertEntityArtAccess` (`server/utils/entityArt.ts:393`) was unreachable. On 2026-08-01 it rejected **131 of 1,824** entity-art enqueues made with the site owner's own ADMIN token — every one against the 43 Scenarios owned by `userId: 9` — with *"You cannot edit artwork for this item."*

The same permanently-false value also flowed into `resolveEnqueueLoraResource` (`:351`) and `resolveArtFacetSelection` (`:363`), silently narrowing LoRA visibility and mature Facet selection for admins.

### 2. `server/api/chats/user/human/[id].get.ts:62`

```ts
const authUser = user as { id: number; isAdmin?: boolean }
if (authUser.id !== requestedUserId && !authUser.isAdmin) { /* 403 */ }
```

`validateApiKey` returns `{ id, Role }` and never `isAdmin`, so `!authUser.isAdmin` was always true and admins were wrongly denied other users' human chats — with `Role` sitting right there on the object, unread. Now derived through the canonical `userIsAdmin()` predicate in `server/utils/authUser.ts`.

## The fix

- `comfyGate.ts` and `textGate.ts` return the resolved `isAdmin` alongside the narrowed `user`, with a comment on the field explaining why the narrowing makes a cast dangerous. `textGate` has no live consumer yet — giving it the passthrough now keeps it from becoming a third instance.
- `enqueue.post.ts` reads `gate.isAdmin`; the cast is deleted.
- `chats/user/human/[id].get.ts` uses `userIsAdmin(user)`.
- New `utils/scripts/verifyAdminFlagCasts.ts` fails CI on any inline `as { ... isAdmin ... }` assertion. It deliberately does **not** match casts to named types that genuinely declare the field, so `AuthUser` / `AuthGuardResult` stay legal. Wired into `contract-tests.yml` next to the existing `no-prisma-json-cast` guard, which it is modelled on.

## Root cause

Two auth resolvers with incompatible user shapes — `authGuard` returns the full user plus `isAdmin`; `validateKey` returns `{ id, Role }` — bridged at the gate layer by `as` casts instead of by data. The verifier closes the class; a follow-up will consolidate the 49 inline `user.Role === 'ADMIN'` comparisons onto `userIsAdmin`.

## Verification

- `test:admin-flag-casts` — passes over 1,663 files; regex unit-checked against both real instances (matches) and `as AuthUser` / `gate.isAdmin` / `auth.isAdmin` (no match).
- `test:no-prisma-json-cast`, `verifyEntityArtManager`, `test:conductor-api-auth-guards` — all pass.
- **Against production, pre-fix**, at zero mana cost: `POST /api/art/enqueue` with `mode: recreate` + `engine: kontext` for scenario 61 (owned by user 9) returns

  ```
  {"success":false,"message":"You cannot edit artwork for this item.","statusCode":403}
  ```

  That request is rejected at the access check. Once this deploys it should instead return the 400 raised immediately past it (*"Entity recreation requires a prompt-only engine such as Krea 2."*) — proving the admin branch fires, without queueing a job or spending mana. Will re-run the probe and then the 131 rejected jobs after merge.

No schema change.

---

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -98,6 +98,9 @@ jobs:
       - name: Prisma JSON-cast contract
         run: npm run test:no-prisma-json-cast
 
+      - name: Admin-flag cast contract
+        run: npm run test:admin-flag-casts
+
       - name: Unquoted reserved-word table identifier contract
         run: npm run test:no-unquoted-reserved-word-tables
 
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "test:deploy-wait-ancestry": "tsx utils/scripts/verifyDeployWaitAncestry.ts",
     "test:deploy-wait-noop": "tsx utils/scripts/verifyDeployWaitNoOp.ts",
     "test:no-prisma-json-cast": "tsx utils/scripts/verifyNoPrismaJsonCast.ts",
+    "test:admin-flag-casts": "tsx utils/scripts/verifyAdminFlagCasts.ts",
     "test:no-unquoted-reserved-word-tables": "tsx utils/scripts/verifyNoUnquotedReservedWordTables.ts",
     "test:fetch-generic-pinning": "tsx utils/scripts/verifyFetchGenericPinning.ts",
     "test:existing-owner-id-nullability": "tsx utils/scripts/verifyExistingOwnerIdNullability.ts",
--- a/server/api/art/enqueue.post.ts
+++ b/server/api/art/enqueue.post.ts
@@ -279,7 +279,7 @@ export default defineEventHandler(async (event) => {
       frames: videoFrames,
       serverId: body?.serverId ?? null,
     })
-    const isAdmin = Boolean((gate.user as { isAdmin?: boolean }).isAdmin)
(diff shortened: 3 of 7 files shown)


silasfelinus:
REVIEWING: hourly-health-20260801T231345Z-pr1280 at 2026-08-01T23:13:45Z

PR #1287, opened 2026-08-02
CHILD forces mature content off, even when the user is also an admin

silasfelinus:
## Why this is a separate PR

This work was pushed to #1286 but **did not merge with it** — the squash captured only that PR's first commit, so `main` ended up with the multi-role write path and none of the maturity rule. Re-applied here onto current `main` as a fresh change, per the merged-PR-is-finished rule. Also carries a one-line follow-up spotted while verifying against production.

The multi-role sequence so far: #1280 (dead admin bypasses) → #1282 (`UserRole` table) → #1284 (read path) → #1286 (write path + client) → **here**.

---

## The rule

**RESTRICTIVE WINS.** When two roles disagree, the restriction holds: a CHILD who is also an ADMIN is still barred from mature content. Admin grants capability; it does not lift a safety restriction.

This is the one *behavioural* change in the whole sequence, and the reason the precedence rule had to be decided before any of it shipped. Without it, the obvious way to give a family account admin powers — exactly what was asked for — would silently unlock mature content for a child.

`server/utils/contentAccess.ts` gains **two** functions, because the routes needed both:

| | |
|---|---|
| `effectiveShowMature(user)` | the stored preference, restriction applied |
| `isMaturityRestricted(user)` | barred outright, whatever they ask for |

### Why the second one exists

`art/image/index.get.ts` and `art/image/[id].get.ts` **OR a `?showMature=true` query parameter** with the stored preference:

```ts
const showMature = isAuthenticated ? requestedMature || Boolean(user?.showMature) : false
```

Gating only the preference would have left a CHILD able to hand themselves mature content in a query string. *A restriction a query parameter can lift is not a restriction.* Both routes now compute:

```ts
isAuthenticated && !isMaturityRestricted(user) && (requestedMature || user?.showMature === true)
```

so a request can still opt **down**, but never past the restriction.

### Why the CHILD marker is read from every source

Join table, `roles` array, primary `Role` column, and the lowercase `role` alias the art routes use.

This is deliberately the **opposite** of how every other predicate in this system works. Elsewhere a role *grants* something, so a missed source merely denies a grant. Here the role *denies*, so a missed marker **fails open** — the wrong direction for a child-safety check.

Routed the eight server `showMature` reads through it and mirrored it in `userStore.showMature`, so the UI doesn't offer content the API will refuse.

---

## Also included

The admin PATCH response carried both `roles` and the raw `UserRoles: [{ role }]` relation. `roles` exists precisely so no client has to know the join table's row shape; shipping both invites one to depend on it. `UserRoles` is now destructured out.

---

## Verification

- `verifyChildMaturityRestriction` — **21 cases**: the CHILD+ADMIN headline with `showMature: true`, CHILD as *secondary* role, user id 1 (the bootstrap admin hatch) still restricted while CHILD, detection from each of the five sources, the query-parameter bypass attempt, FAMILY ≠ CHILD, ordinary opt-in still working, and client/server agreement on all of it.
- `verifyUserRolePredicates`, `verifyMaturityPrivacyContract` — pass.

### Verified end-to-end against production

After #1286 deployed, the migration and backfill were confirmed live (`/api/users/me` returns `roles: ["ADMIN"]` from the join table). Then, on a disposable account:

- created with `roles: ["CHILD", "ADMIN"]` — **the combination that was impossible before this work** — and confirmed both persisted and round-tripped
- confirmed a later single-role write correctly cleared **both** stale secondaries, which is what `setUserRoles` exists for
- deleted the account

---

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -110,6 +110,9 @@ jobs:
       - name: UserRole predicate contract
         run: npm run test:user-role-predicates
 
+      - name: CHILD maturity restriction contract
+        run: npm run test:child-maturity
+
       - name: Unquoted reserved-word table identifier contract
         run: npm run test:no-unquoted-reserved-word-tables
 
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "test:admin-flag-casts": "tsx utils/scripts/verifyAdminFlagCasts.ts",
     "test:user-role-expand": "tsx utils/scripts/verifyUserRoleExpand.ts",
     "test:user-role-predicates": "tsx utils/scripts/verifyUserRolePredicates.ts",
+    "test:child-maturity": "tsx utils/scripts/verifyChildMaturityRestriction.ts",
     "test:no-unquoted-reserved-word-tables": "tsx utils/scripts/verifyNoUnquotedReservedWordTables.ts",
     "test:fetch-generic-pinning": "tsx utils/scripts/verifyFetchGenericPinning.ts",
     "test:existing-owner-id-nullability": "tsx utils/scripts/verifyExistingOwnerIdNullability.ts",
--- a/server/api/art/image/[id].get.ts
+++ b/server/api/art/image/[id].get.ts
@@ -5,6 +5,7 @@ import prisma from '../../../utils/prisma'
 import { errorHandler } from '../../../utils/error'
 import { validateApiKey } from '../../../utils/validateKey'
 import { userRoles } from '../../../utils/authUser'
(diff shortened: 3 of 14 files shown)


silasfelinus:
REVIEWING: kr-hourly-health-2026-08-02T011435Z at 2026-08-02T01:14:35Z

Diagnosed the two red checks. Contract Tests failed because the new verifier statically imported database-backed `contentAccess` with no `DATABASE_URL`; Facet Alias Smoke independently hit a transient Prisma engine checksum 502. Pushed test-isolation fix `8a46ce2` on this branch and am holding merge for completed green checks.

PR #1297, opened 2026-08-02
Give Storymaker three reading modes: Classic, Storymaker, Storymaker Dark

silasfelinus:
## Why

> *"I like the second one best, but really the story maker should let the user pick from all three, but I would call them Classic (daisy ui defaults), story maker, and story maker-dark."*
> — Silas, 2026-08-02

The three t-001 mockups aren't a one-time comparison after all — they're a **per-reader preference**, renamed to what he calls them:

| Now | Was | What it is |
|---|---|---|
| `classic` | Playground (C) | daisyUI defaults — the reader's own theme |
| `storymaker` | Storybook (B) | warm paper, serif narration — **default** |
| `storymaker-dark` | Theater (A) | a lit stage in a dark house |

**This revises my read from #1294.** I took *"let's run with Storybook as the design"* as a single app-wide aesthetic and deleted the other two token sets. They come back — but as modes, not competing app themes.

## What makes "Classic" mean something

The app-wide default **stays** the warm-paper theme, since he called it the design and likes it best. That's exactly what gives Classic a job: it binds **no** `data-theme` and applies no bespoke narrative chrome, so the story renders in whatever theme the reader picked globally. Pinning a theme for Classic would break its own definition the moment someone changed theirs.

Classic binds `undefined`, not `''` — Vue drops an undefined attribute, whereas `data-theme=""` is a *real* attribute that daisyUI resolves to its fallback, silently overriding the reader's global choice. That's the one thing Classic exists to preserve.

## Scoping

The mode applies as `data-theme` on the **story subtree**, not on `<html>`. Picking Storymaker Dark doesn't hijack the rest of the app — a reader running the site in `synthwave` keeps synthwave everywhere else.

It's also a **separate preference** from `themeStore` rather than a second writer to it: two writers to the root `data-theme` would race.

## The rename

`storybook` → `storymaker`, to match his wording. A `RENAMED_THEMES` map in `themeStore` keeps anyone who loaded the site during the few hours the old name was live — without it, the stale localStorage value fails the `daisyuiThemes` check and they get silently reset, losing an explicit choice.

`storymaker-dark` is new: **not an inverted palette but an inverted lightness**, holding the hues so terracotta stays terracotta and the art still reads as the lit thing in the room. A naive dark theme shifts the accent and breaks that continuity. It sets `prefersdark`, the honest default for a reading surface.

## Cleanup

The mockup page (`storymaker-mockups.vue` + its content route, 299 lines) is **deleted**. t-003 always said to remove it once a direction was picked, and the real surface now does what it was for. Nothing else referenced it.

## Verification

- **An actual client build**, since `vue-tsc` doesn't compile CSS: both `[data-theme=storymaker]` and `[data-theme=storymaker-dark]` compile into the bundle with their mapped oklch tokens.
- The postcss lexical warnings in that build are **pre-existing** — clean `main` emits the same seven, checked by stashing.
- `test:layout-contract` holds (one-header is down to 6 from parallel work). `channel-content`, `nav-manifest`, `data-surface-manifest` pass — confirming nothing depended on the deleted route.

## Worth a look when you try it

Classic currently differs from Storymaker by *theme* but not yet by *layout* — the plate border, drop cap and serif narration from the mockup aren't conditioned on `isStoryStyled` yet, because the real narrator stage (`kr-narrator-stage.vue`, built in t-006) never adopted them. The composable exposes `isStoryStyled` ready for that. Tell me if you want Classic to strip the chrome too, or if theme-only is the right distinction.

---


silasfelinus:
REVIEWING: claude-quirky-curie-20260802T062629Z at 2026-08-02T06:26:29Z

---

PR #1314, opened 2026-08-02
interface-vision Phase 3: Storybook and Taskmaster onto the shared conversation kit

silasfelinus:
Phase 3 of the redesign plan.

> *"a lot of these are just going to be variations of each other, so the more we can work on modular pieces for things like chat windows, multiple choice returns, and image presentations, the better."* — Silas, 2026-08-02

## The duplicate transcript is gone

`narrative-transcript.vue` and `kr-chat-window.vue` were the same component with different markup. Keeping both is precisely how the last shared-UI effort failed — `global-ui` closed 25/25 "done" at ~7% adoption — so this deletes the older one and moves **both** products across rather than leaving a 113-line copy to drift.

Consolidating also let the survivor inherit what the old one had proved:

- **`role="log"` is gone.** It carries an implicit `aria-live="polite"`, so the scroll container was a live region *on top of* the explicit sr-only status below it — every arriving turn announced twice, once verbatim and once as a summary. The old transcript's contract had a named assertion against exactly this; it applies to the survivor now.
- The container is a `<section>` with an accessible name, so assistive tech can **jump to** the transcript instead of only walking into it.
- **Scene numbers count narrator turns only.** Numbering by array index would have drifted one per answer and called the third scene "Scene 5".

## Four more hand-rolled pick-one lists folded in

`kr-choice-list` now renders the composer's suggested responses, Taskmaster's checkpoint-outcome picker, and its tone row — on top of the narrator's quick topics it already had. Each was the same widget with different styling. It also gained the focus ring and reduced-motion guards the rest of the narrative components carry and it was missing.

`narrative-art-status`' rendered illustration goes through `kr-art-plate`, so the variant fallback chain lives in one resolver rather than in every surface — the duplication that left `cardPath`/`heroPath` invisible for a day in t-007.

`kr-art-plate`'s `framed` boolean became `frame: 'plate' | 'thin' | 'none'`, because "unframed" has to mean *genuinely* unframed: a hairline plus an inner radius inside an already-rounded parent leaves visible notches.

## One scroll region while reading

Both pages had the transcript scrolling **inside** a scrolling page.

Setup still scrolls as a document — it is a long form and legitimately runs past the fold. But once a story or a quest is running, the chat window is the only scroll owner. The bible, state panel, objective, checkpoint rail, outcome picker and composer are pinned around it, so *what you are doing* and *what you answer with* never scroll away.

Taskmaster's checkpoint grid was unbounded **and** pinned, which on a long quest would have pushed the story itself off screen. The current action stays visible; the full plan is one disclosure away.

The recap and quest ledger needed to scroll *with* the conversation rather than pin below it — that is what the chat window's new `#footer` slot is for. Phase 4 needs the same slot for Dreams' `musings`.

## Contracts retargeted, not relaxed

Five source-text contracts pinned the deleted component. Every assertion moved to the survivor, each preserved or strengthened:

| was | is |
|---|---|
| `"Your response: "` inline sr-only span | a real sr-only **heading** a screen reader can navigate to |
| `fieldset` + sr-only `legend` | `role="group"` + `aria-label` on the shared list — same named group, one implementation |
| substring search for the live-region bug | reads the scroll container's own opening tag |
| `art` pinned on the presentational component | moved to the shared beat→turn adapter and the two pages |

That third row was a real bug in my first pass: the component *explains in a comment* why it avoids `role="log"`, and the substring search matched the explanation and failed on the bug it documents the absence of.

The fourth is a deliberate improvement — `kr-chat-window` is domain-free so it can also serve bot chats and the Dreams dock; re-pinning `NarrativeArtJobState` to it would re-couple it to one product's session shape.

`verifyNarrativeKit` gained a check that no file reappears under the old transcript name, and counts the composer and art-status as intermediaries since both now render kit pieces internally.

The **Storybook/Taskmaster product boundary is untouched**: separate stores, separate state machines, no write-back control on the Storybook side.

## Verification

- `npx nuxi build` — both bundles compile; prerender fails only on a missing `JWT_SECRET` this environment does not have
- `npm run test:layout-contract` — no new violations
- All six narrative contracts pass, plus `test:nav-manifest`, `test:channel-content`, `test:data-surface-manifest`
- `cypress/public/narrative-accessibility.cy.ts` needs **no change** — its selectors (`section[aria-label="Story transcript"]`, `article[aria-labelledby] h3.sr-only` → "Scene 1", `button` "Blocked" `aria-pressed`) resolve to the new markup unchanged. It has **not** been executed here.

---


silasfelinus:
REVIEWING: claude-quirky-curie-5koqam-conductor-sweep at 2026-08-02T11:38:00Z

---

PR #1328, opened 2026-08-02
interface-vision: fix three blank routes, and gate the mistake that caused them

silasfelinus:
A full read-only sweep of the navigation surface turned up **three routes that render nothing**. Two are live admin tabs.

| Route | Mounted | Registry actually exports |
|---|---|---|
| `/navigation-health` (admin tab) | `:navigation-navigation-health` | `NavigationHealth` |
| `/project-placement` (admin tab) | `:projects-project-placement-manager` | `ProjectPlacementManager` |
| `/sheets` | `:dream-sheets` | *nothing* |

`nuxt.config.ts` registers components with `pathPrefix: false`, so an MDC mount name is the **bare filename**. All three carried a directory prefix, resolved to no component, and Vue rendered nothing rather than erroring.

## The real bug is that CI couldn't see them

`verifyChannelContent.ts` and `verifyNavManifest.ts` both parse frontmatter only. `verifyLayoutContract.ts` is the one script that *already* extracts MDC mount names — it needs them for the `one-mdc` rule — but never checked they resolve.

This adds that check, resolving against the component files on disk rather than `.nuxt/components.d.ts` (a build artifact the script must not depend on — it runs standalone in CI before any `nuxi prepare`). It reports the near-match:

```
content/navigation-health.md: mounts :navigation-navigation-health, which is not
a component (components are registered with pathPrefix:false, so the mount is the
bare filename — did you mean :navigation-health?)
```

**Proved it fails before trusting it**: re-introduced the original typo, confirmed the contract went red with that exact message, restored. 68 mounts resolve across `content/`.

## The redirect key nothing read

`content/facet-gallery.md` has carried `redirect: /facets` since the Facet browser became canonical, and `verifyFacetGallery.ts` **accepts that key as one of two valid states**. But `pages/[...slug].vue` never read it, so the route rendered its own *"This legacy route redirects to…"* prose instead of redirecting — while `dashboardHelper.ts` still routes a dashboard card there.

The host honours it now, **guarded to same-origin paths**: a content file is data, and following an arbitrary absolute URL from one would turn any future content edit into an open redirect. A redirecting page also no longer paints its body first.

## `/sheets` is deleted, not repointed

It's in no channel manifest, nothing links to it, and its stated job ("browse Dreams and create missing PitchSheets") is already done by `dream-sheet-toolbar.vue` inside `dream-gallery.vue`. The obvious repoint target, `dream-pitch-sheet.vue`, is a single-sheet card taking a `dream` prop — not a browser, so it could never have mounted here. Its `navComponent: 'dream-nav'` names a nonexistent component too, and that key is a declared-but-never-read field on `contentStore`'s type.

## Gallery adoption is now counted

`global-ui` closed 25/25 "done" at ~7% adoption because nobody measured the whole. Extraction has never been the hard part here — the shared pieces exist and work. Adoption is, and an unmeasured number doesn't move.

Two of three shared pieces are already counted: `verifyNarrativeKit.ts` prints the conversation kit's adoption, and `verifyLayoutContract.ts` gained a root-surface rule (#1320). **`kr-gallery` was the one nothing counted** — it sits at **1 of 7** core-object galleries while "all seven core objects share one gallery" is an open beta gate.

`verifyGalleryAdoption.ts` follows `verifyNarrativeKit.ts`'s shape deliberately: informational on adoption (print the count, name the holdouts, never fail — that would block every unrelated PR), hard only on the invariants that make the shell worth adopting (stays store-free; emits its view mode rather than persisting it).

## Note on overlap

Eleven PRs (#1317–#1327) landed while this sweep was running. I re-baselined against them before committing and **dropped half of what I'd drafted** — a layout-primitive adoption counter, which #1320 had already added as a seventh layout-contract rule. Only the genuinely uncounted piece (galleries) ships here.

## Verification

- `test:layout-contract`, `test:channel-content`, `test:nav-manifest`, `test:facet-gallery`, `test:data-surface-manifest`, and the new `verifyMdcScrollOwnership` all pass
- Line endings audited on every touched file
- **The three repaired routes have not been opened in a browser here** — worth a click on the preview

---


silasfelinus:
REVIEWING: 2026-08-02T191336Z-pr1328-k7v4 at 2026-08-02T19:13:36Z

PR #1334, opened 2026-08-02
interface-vision (t-055): kr-gallery degrades, Facets adopts it, and six copies of one chain collapse

silasfelinus:
Two commits. First of the seven core-object galleries, plus the duplication it exposed.

---

# 1. Teach kr-gallery to degrade, and move Facets onto it

## Why the shell couldn't absorb anything yet

`kr-gallery` rendered ``'&lt;img :src="displayImage(item)"&gt;'`` unconditionally. An unillustrated row produced ``, which browsers resolve as *"re-request the current page"* and paint as a broken image.

Any gallery whose rows are routinely unillustrated couldn't adopt the shell without regressing — and the Facet catalog is exactly that, since it marks which facets are *meant* to be illustrated (`artRequired`) and leaves the rest bare. There's now a placeholder branch with an optional per-item icon and label, so Facets keeps saying **"art pending"** vs **"no art"**.

Second: `displayImage()` was itself a hand-rolled copy of the `cardPath`/`heroPath`/`iconPath` chain — the duplication `resolveArtVariantSrc` exists to end, sitting inside the component meant to end it. `GalleryItem` now takes an optional `source` which the shell resolves.

Pre-resolved `card`/`hero`/`icon` **stay** and are still preferred. `conductor-project-gallery-page.vue` merges a remote conductor record, detects canonical paths and appends a cache-busting revision — no generic resolver can do that. `source` is a fallback, not a replacement.

## Facets

Adopted **per taxonomy group**, not one flat grid: `verifyFacetGallery.ts` pins this as a read-only grouped catalog showcase. kr-gallery owns no scroll region, so mounting it per group is fine; `:modes="[]"` drops the mode bar.

**A behaviour difference caught before committing:** `facetArtwork()` feeds the "illustrated only" filter as well as rendering. Asking the resolver for `'card'` alone would have silently dropped facets illustrated *only* at hero or icon size out of that filter.

233 → 207 lines. Counter moves **1/7 → 2/7**.

---

# 2. Collapse six copies of the facet artwork chain

`cardPath || imagePath || heroPath [|| iconPath]` was hand-written **six times**:

| File | Helper |
|---|---|
| `utils/narrativeIngredients.ts` | `narrativeIngredientArtwork()` |
| `components/facets/facet-manager.vue` | `facetArtwork()` |
| `components/facets/facet-picker.vue` | `facetArtwork()` |
| `components/characters/character-facet-picker.vue` | `artwork()` |
| `components/rewards/reward-facet-picker.vue` | `artwork()` |
| `components/art/art-facet-selector.vue` | `facetArtwork()` |

**They had already drifted** — facet-manager's copy included `iconPath`, the other five didn't. That's the failure mode a shared resolver prevents, and `resolveArtVariantSrc` has existed since t-007; these six just never used it.

`resolveEntityArtwork()` is now the single implementation. `narrativeIngredientArtwork` keeps its name and export (`verifyNarrativePrimitives.mjs` pins it); only its body moved.

**Ordering deliberately preserved, not "improved."** Asking for card → hero → icon yields `cardPath → cardData → imagePath → imageData → heroPath → … → iconPath`, so the relative precedence of the three path fields matches every copy replaced. The base64 fallbacks are a strict superset, reachable only where a path was absent and the old chain returned null.

---

## Two things found and NOT fixed — both pre-existing

`verifyFacetArtworkMigration.ts` asserted the literal old chain, so I updated that one assertion rather than leave a contract pinning text this PR deleted. But **it was already failing before this change**, on an unrelated assertion (`v-model="newImagePath"`) — confirmed by stashing and re-running against a clean tree.

The reason nobody noticed is worse: its CI step runs `npx tsx … 2>&1 | tee report.txt`, so the step's exit code is `tee`'s and **the check can never fail**. Filed as conductor **t-061** (fix the pipeline first, then the contract).

Also filed **t-062**: the remaining five galleries can't adopt kr-gallery as it stands — they delegate to rich per-object cards and use a horizontal-scroller row layout kr-gallery has no equivalent for. That's a product decision, not effort.

## Verification

- `test:facet-gallery`, `test:facet-catalog`, `test:facet-aliases`, `test:gallery-adoption`, `test:layout-contract`, `verifyNarrativePrimitives` — all pass
- Line endings audited on all touched files
- **Not opened in a browser.** `/facets` is worth a look — a grouped grid is exactly what a source-text contract can't judge.

### DIFF
--- a/components/art/art-facet-selector.vue
+++ b/components/art/art-facet-selector.vue
@@ -150,6 +150,7 @@ import {
 } from '@/stores/facetCatalogStore'
 import { useFacetArtRequestStore } from '@/stores/facetArtRequestStore'
 import { normalizeFacetLookupKey } from '@/utils/facetAliases'
+import { resolveEntityArtwork } from '@/utils/artImageSrc'
 
 const props = withDefaults(
   defineProps<{
@@ -266,7 +267,7 @@ async function requestArtwork(facet: FacetCatalogEntry): Promise<void> {
 }
 
 function facetArtwork(facet: FacetCatalogEntry): string | null {
-  return facet.cardPath || facet.imagePath || facet.heroPath || null
+  return resolveEntityArtwork(facet)
 }
 
 function taxonomyLabel(taxonomy: string): string {
--- a/components/characters/character-facet-picker.vue
+++ b/components/characters/character-facet-picker.vue
@@ -108,6 +108,7 @@ import {
 } from '@/stores/facetStore'
 import { useCharacterFacetStore } from '@/stores/characterFacetStore'
 import { normalizeFacetLookupKey } from '@/utils/facetAliases'
+import { resolveEntityArtwork } from '@/utils/artImageSrc'
 
 const props = withDefaults(
   defineProps<{
@@ -156,7 +157,7 @@ const searchResults = computed(() => {
 })
 
 function artwork(facet: FacetWithAliases): string | null {
-  return facet.cardPath || facet.imagePath || facet.heroPath || null
+  return resolveEntityArtwork(facet)
 }
 
 async function loadAssigned(): Promise<void> {
--- a/components/facets/facet-gallery.vue
+++ b/components/facets/facet-gallery.vue
@@ -67,69 +67,16 @@
         <span class="badge badge-secondary badge-sm">{{ group.entries.length }}</span>
       </div>
 
-      <div class="grid grid-cols-2 gap-3 sm:grid-cols-3 lg:grid-cols-4 xl:grid-cols-5">
-        <article
-          v-for="facet in group.entries"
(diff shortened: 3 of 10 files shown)


silasfelinus:
REVIEWING: 2026-08-02T211211Z-pr1334-q7m2

silasfelinus (commented):
Reviewed the full diff and final head. The grouped Facet migration preserves taxonomy grouping and the art-only semantics, avoids empty-image reloads, centralizes the repeated artwork fallback chain, and keeps the richer object galleries out of scope. All 14 head workflows completed successfully. The inert `tee` contract and remaining gallery product decision are correctly split into Conductor follow-ups rather than smuggled into this diff.

PR #1338, opened 2026-08-03
interface-vision (t-031): Dreams onto the shared card shell, gaining reactions

silasfelinus:
> *"the final version obviously is wanting a single display that is consistent across galleries, all with those bells and whistles."* — Silas, 2026-08-02

Investigating that turned up something better than expected.

## Four of five cards already share a shell

`bot-card`, `character-card`, `reward-card` and `scenario-card` all wrap `components/wonderlab/reactable-card.vue`, which owns the `<article>`, selected border, karma badge, reaction button and review panel. **`dream-card` was the only holdout**, still rolling its own `<article>`.

That's why Dreams have no reaction or review affordance at all. It was never a data problem — `dream` has been a valid `ReactionTargetType` since reactions shipped, `DREAM` is in `reactionCategories`, `reactionStore` already maps `dream → dreamId`, and `Dream.allowReviews` is in the schema. Every piece was in place; nothing was wired to it.

*(This also corrects what I filed in t-062, which said "five bespoke cards." Wrong — what's actually duplicated is the card **body**, not the shell.)*

## What changed

`dream-card` now wraps `reactable-card`. Its hover action bar becomes the `#actions` slot, which the shell already positions and stop-propagates, so that absolutely-positioned wrapper div is gone.

**One addition to the shared component: a `padded` prop.** The four panel cards inset their art and want the padding; Dreams runs art edge-to-edge with the title overlaid, where padding would frame it wrongly. A typed boolean rather than a `p-0!` string in `cardClass` — nothing in this repo uses Tailwind's important modifier, and a class-string fight is a worse contract. **Defaults true, so all fourteen existing `reactable-card` consumers are unaffected.**

Also collapsed `selected || isSelected` into one `activeSelected` computed — two props for one state, declared separately and then always read together. **Both props are kept**, not deduplicated away: callers pass both spellings and dropping either would silently stop working at a call site `vue-tsc` can't see through a template. `character-card` has the same pair; left alone here.

`cardClass` no longer sets the selected or base border — the shell draws those — so it contributes only the ring and the inactive-dream treatment.

## A correction to my own earlier reading

I recorded that `dream-card` fired three events it never declared. **It declares them** — my pattern matched only the shorthand `event: [args]` form and missed the call-signature `(event: 'x', …): void` form. I briefly added a duplicate `defineEmits` on that mistaken basis and removed it. The file now has exactly one declaration, converted to the shorthand the other four use.

## Visual change, flagged

Dream cards gain a reaction button (when selected, and when the Dream allows reviews) and the karma badge. The card itself *should* look the same — `padded: false` preserves the full-bleed figure and I matched the ring/border split deliberately — but "should" is doing work there. **`/dreams` wants a look on the preview.**

## Verification

- `test:layout-contract`, `test:gallery-adoption`, `test:facet-gallery`, WonderLab UI fixture contract — pass
- Line endings checked against `HEAD` on both files

Rebased onto current `main`: #1334's squash captured its first two commits, and this one was pushed after the merge.

---

### DIFF
--- a/components/dreams/dream-card.vue
+++ b/components/dreams/dream-card.vue
@@ -1,9 +1,23 @@
 <!-- /components/dreams/dream-card.vue -->
 <template>
-  <article
-    class="group relative flex h-full min-h-0 cursor-pointer overflow-hidden rounded-2xl border bg-base-100 shadow-sm transition hover:-translate-y-0.5 hover:shadow-xl"
-    :class="cardClass"
-    @click="$emit('choose', dream)"
+  <!-- The shared card shell (interface-vision t-031). Dreams were the only one
+       of the five object cards still rolling their own <article>, which is why
+       they had no reaction or review affordance at all while Bot, Character,
+       Reward and Scenario did -- `dream` has been a valid ReactionTargetType and
+       `DREAM` a valid category the whole time; nothing ever wired it up.
+       :padded="false" keeps the full-bleed poster look this card is built
+       around; the other four inset their art and take the default padding. -->
+  <reactable-card
+    :selected="activeSelected"
+    :compact="compact"
+    :padded="false"
+    :allow-reviews="dream.allowReviews"
+    :target-id="dream.id"
+    target-type="dream"
+    reaction-category="DREAM"
+    :target-title="dreamTitle"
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: 2026-08-03T021430Z-pr1338-x7k9 at 2026-08-03T02:14:30Z

PR #1339, opened 2026-08-03
feat: turn For You into a Conductor attention desk

silasfelinus:
## Root cause
The Home / For You surface only queried database Todos in the legacy `HONEYDO` category. Conductor human gates and pitch files already reached a separate store, but nothing joined those streams. The global Conductor loader also did not recognize `/for-you`. A recent cleanup then hid the Home tab because the legacy producer did not exist, leaving the route deep-link-only.

## What changed
- restore the Home / For You tab and its navigation/data-surface contracts
- show every live Conductor task in `needs-human`, with hard gates ranked before soft gates
- add admin actions for Approve, Send back, and Send note
- queue those decisions as Conductor `task-events` instead of editing roadmaps or keeping a second local state machine
- surface pending Conductor pitches directly in For You with approve/reject controls
- keep the existing personal HONEYDO Todo cards as a separate follow-up section
- load Conductor admin data on the For You route and dashboard tab
- project new human gates and pitches into the existing notification bell for admins, using a browser-local seen signature so repeated unchanged attention does not keep appearing new
- expose `soft_gate` from roadmap parsing so the UI can distinguish blocking human gates from soft questions

## Companion dependency
Depends on conductor PR #1591, which lets task events record `approved_by_human`. Merge conductor first, then this PR.

## Safety
- admin guard on the new write route
- validates project/task/action and re-reads the live roadmap before accepting an action
- refuses stale actions when the task is no longer `needs-human`
- writes a unique event file through the existing Conductor GitHub bridge
- no schema migration and no duplicate notification rows

## Verification
Repository CI plus Vercel deployment verification. The new endpoint is covered by the existing Conductor API auth-guard contract because it calls `requireAdminApiUser`. Vercel currently exposes no PR preview deployment for this branch, so the visual pass will use the production deployment immediately after merge.

### DIFF
--- a/components/pages/for-you-manager.vue
+++ b/components/pages/for-you-manager.vue
@@ -4,75 +4,335 @@
     class="flex h-full min-h-0 w-full flex-col gap-3 overflow-hidden p-1.5 sm:p-2.5"
   >
     <div class="rounded-2xl border border-accent/20 bg-accent/5 px-4 py-3">
-      <p class="text-xs font-semibold text-accent/80">🍯 For You</p>
-      <p class="mt-0.5 text-xs text-base-content/50">
-        Action items your AI assigned to you. These help your projects along —
-        check them off as you go.
-      </p>
+      <div class="flex flex-wrap items-start justify-between gap-2">
+        <div>
+          <p class="text-xs font-semibold text-accent/80">✨ For You</p>
+          <p class="mt-0.5 text-xs text-base-content/55">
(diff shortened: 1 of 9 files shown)


silasfelinus:
REVIEWING: gpt56-for-you-20260803T023234Z at 2026-08-03T02:36:00Z

Full-story review includes the restored Home / For You tab, Conductor data loading, human-gate event actions, pitch actions, admin notification projection, and preview verification. Companion Conductor PR #1591 must merge first.

PR #1344, opened 2026-08-03
style: adopt kr-surface on For You manager

silasfelinus:
### Task
interface-vision/t-017: Sweep the layout-contract allow-list to zero, daily-drivers first, admin last  (kind: software)

- Added the shared `kr-surface` root marker to `components/pages/for-you-manager.vue` without changing its existing size, overflow, spacing, or scrolling behavior.
- Ratcheted the `root-surface` allow-list from 14 entries to 13.

### How I verified
- Inspected the complete branch diff against current `main`: two files, one class-token addition, one baseline deletion.
- CI is the authoritative verification for the layout contract and TypeScript in this connector-only run.

### Stakes
reversible

### Kaizen suggestion
Audit the remaining root-surface entries for simple marker-only conversions before attempting structural scroll changes.

### Notes for reviewer
The For You manager is already a bounded, overflow-hidden root with one child scroll owner, so adding `kr-surface` is contract adoption rather than a layout restructure.

### DIFF
--- a/components/pages/for-you-manager.vue
+++ b/components/pages/for-you-manager.vue
@@ -1,7 +1,7 @@
 <!-- /components/pages/for-you-manager.vue -->
 <template>
   <section
-    class="flex h-full min-h-0 w-full flex-col gap-3 overflow-hidden p-1.5 sm:p-2.5"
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: 2026-08-02T211555-0700-pr1344-v8n2 at 2026-08-02T21:15:55-07:00

PR #1354, opened 2026-08-03
feat: hide paused project gates by default

silasfelinus:
## Summary

- split Conductor human gates into active and paused project groups
- keep active gates as the notification and urgency source
- hide paused-project gates by default on For You
- add an explicit `Show paused projects` toggle with a paused count
- visually label and mute paused cards when the archive view is enabled

This fixes paused projects such as Mermaids of Venice surfacing as current work without deleting their preserved roadmap gates.

## Verification

- TypeScript and repository contract checks will run in CI
- notification behavior remains active-only because the notification store consumes `humanGates`, not `allHumanGates`
- no API or schema change

### DIFF
--- a/components/pages/for-you-manager.vue
+++ b/components/pages/for-you-manager.vue
@@ -27,7 +27,7 @@
               class="badge badge-warning h-auto gap-1 rounded-xl px-3 py-2"
             >
               <Icon name="kind-icon:lock" class="size-3.5" />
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: chatgpt-state-reconciliation-20260803T0748Z at 2026-08-03T07:48:20Z

Reviewing lifecycle filtering, notification behavior, and responsive For You presentation before merge.

PR #1360, opened 2026-08-03
wonderlab: delete two dead components, and write down why that was always safe

silasfelinus:
Silas, this session:

> "kill them, and just keep the components … the museum reviews are actually literally reviews of what the site has been, not where it is … it's not so we have a reason to cling to outdated components."

`components/scenarios/choice-selector.vue` and `choice-preview.vue` were left in the tree — by me, yesterday — on the theory that deleting them would orphan four published museum reviews. **It would not.** The investigation is the more valuable half of this change, so it's now a document rather than tribal knowledge.

## The museum record is DB-anchored, not filesystem-anchored

| Claim | Reality |
|---|---|
| Reconcile deletes rows for vanished files | It doesn't. `server/utils/wonderlabComponentReconcile.ts:359-366` marks them `isDiscovered: false`. There is no delete branch, and `verifyComponentCanonicalRuntime.ts:138` asserts that. |
| The voice-polish batch pins reviews to files | It pins them by `sourceKey`, which *looks* like a file dependency. `scripts/preflight-wonderlab-voice-polish-batch.mjs` compares it against the **`Component` row's `sourceKey` column** (`sourceLock: actual.sourceKey === revision.sourceKey`), never against disk. That workflow's path filter doesn't include component sources at all. |
| The generated manifests would break | `public/wonderlab-components.json` and `utils/generated/wonderLabSourceEvidence.ts` are untracked build artifacts. They just stop listing a deleted component. |
| Retirement needs new machinery | `ComponentStatus.RETIRED` already exists (`prisma/schema.prisma:2540`), is settable from `lab-interact.vue:154`, sorts last in `componentCatalog.ts`, and the museum already has a `discovery: 'missing'` filter. |

Nothing needed building — only the confidence to use what was already there.

## What ships

- **Deletes both files.** Verified unreferenced: no imports, no MDC mounts in `content/`, no template usage anywhere in the repo. Their four reviews (ids 1791–1794, by The Question Mark, The Whispering Woods, The Reanimation Counter's Best Work, and Mx. Quorum) stay attached to Components 896/897 as the historical record they are.
- **`docs/wonderlab-component-retirement-policy.md`** — the two operations, so this doesn't get re-litigated. *Retire* is the default: delete the source, set `RETIRED` with a `statusReason`, let reconcile flip `isDiscovered`. *Succeed* — re-pointing reviews at a successor — is only for a component renamed or redesigned in place, where the reviews genuinely describe the surviving surface. A review of a deleted surface belongs on the deleted surface.

The doc also states the limit: this is not licence to delete something still mounted. `verifyLayoutContract.ts` resolves every MDC mount in `content/`, so a content-referenced component still fails CI if removed.

## Deliberately not done

`scenario-interact.vue:113-155` still hand-rolls an intro picker that duplicates the deleted components' job. Moving it onto `kr-choice-list` is **not** a mechanical swap: `splitIntro` (`stores/helpers/scenarioHelper.ts:34`) yields an optional ALL-CAPS eyebrow plus prose, `kr-choice-list` renders its `hint` slot italic — wrong for a long opening paragraph — and the current markup uses a 2-column `lg` grid the shared component has no layout for. Either extend `kr-choice-list` or accept a deliberate visual change, with preview eyes either way. That's `t-051`'s job, not a cleanup commit's.

## Verification

- `npm run test` (vue-tsc) — exit 0
- `test:wonderlab-reconcile`, `test:layout-contract`, `test:channel-content` — all pass
- Line endings audited on the new file

---


silasfelinus:
REVIEWING: 2026-08-03T10:14:36Z-pr1360-k9x4 at 2026-08-03T10:14:36Z

PR #1392, opened 2026-08-04
t-069: the text over the card art is ALT TEXT, and the audit now catches it

silasfelinus:
## Reproduced

`t-069` has been open since 2026-08-02 marked:

> *"STILL BLOCKED ON A PREVIEW. A source investigation ruled out the obvious causes but could NOT reproduce or locate the defect, and the fix must not be guessed at."*

The new CI audit rendered `/conductor` against production and it showed up on the **first run**: three cards in row 2 with text sprawling across the artwork, one ("Interface Vision") with a broken-image glyph beside it. Row 1 correct. Phone correct.

## What it actually is

**Not misplaced markup — an `` that resolves to nothing, so the browser paints its `alt` text inside the layout box where the art should be.**

The detector caught the mechanism verbatim on a test page:

```
<img> failed 356x200 alt="After killer quarter, Palantir CEO Alex " src=…/default-05.webp
```

A long `alt` in a 356×200 box. That's "text over art."

This is why source review couldn't find it: **nothing in the markup is wrong.** The defect exists only while art is missing — which is common right now, since art is still generating. It also explains why your report and mine were mirror images (you saw the top row, I saw the second): it tracks whichever cards haven't loaded, not a fixed position.

**The overlap detector I proposed would have found nothing.** Alt text isn't an element, so no box intersects any other box. Good thing the evidence came first.

## The check

New `BROKEN-ART`: any laid-out `` that is `complete` with `naturalWidth === 0` (failed), or still incomplete long after networkidle (stalled), is reported with size, alt and src — and fails the run. `/conductor` joins the default route list, since it surfaced this and it's your most-used page.

## The fix

`kr-gallery` already had an art-absent placeholder for *"no URL"*. It now also covers *"URL that failed"* — a different failure with the same remedy. Failed sources are tracked by URL rather than item id, so one missing file is remembered across a mode switch that re-resolves the same path. The `Set` is **replaced, not mutated**, because mutating in place doesn't trip reactivity and the card would keep its broken ``.

Net effect: while art generates, cards show the placeholder instead of their title sprawled across the artwork.

## A trap, now documented in the script

**Do not trust `BROKEN-ART` against a local dev server.** `public/images` isn't in the repo (self-hosted media), so every `/images/...` path 404s locally while returning a `307` to real bytes in production.

My first local run reported ten broken images on `/`. I checked two against production with `curl -L` — both `200`. That's the **second time this session** local absence has masqueraded as a production defect, after the missing `DATABASE_URL`. Worth a comment in the file rather than a lesson relearned.

## Verification

`vue-tsc`, `eslint`, `prettier` clean. Failure path exercised **unpiped**: exit `1` with `BROKEN-ART` findings, exit `0` when clean — read directly, never through a pipe (`t-063`).

**Not verified locally, by design:** whether the fix clears `/conductor`. Local has no DB, so the gallery renders no cards at all. This PR's own `deployment_status` audit measures it against the preview with real data — which is precisely what that machinery was built for. If it comes back red on other components' images, that's real signal about where else this pattern lives, and I'll follow it.

---

### DIFF
--- a/components/gallery/kr-gallery.vue
+++ b/components/gallery/kr-gallery.vue
@@ -112,11 +112,19 @@
               {{ badge.label }}
             </span>
           </div>
+          <!-- The corner icon puck. Same failed-source guard as the art above:
+               alt="" means a missing file paints no TEXT, but it still paints a
+               broken-image box, which is the small broken glyph left on
+               /conductor after the main-art fix landed (the CI audit found
+               exactly one, interface-vision-icon.webp, missing from the
+               conductor repo). A decorative icon that will not load should
+               simply not render. -->
           <img
-            v-if="mode !== 'icons' && item.icon"
+            v-if="mode !== 'icons' && item.icon && !failedArt.has(item.icon)"
             :src="item.icon"
             alt=""
             class="absolute bottom-2 left-2 size-11 rounded-xl border border-white/25 object-cover shadow"
+            @error="onArtError(item.icon)"
           />
         </div>
         <div class="p-3" :class="mode === 'icons' ? 'text-center' : ''">
--- a/components/gallery/kr-gallery.vue
+++ b/components/gallery/kr-gallery.vue
@@ -60,16 +60,27 @@
       >
         <div class="relative overflow-hidden" :class="imageWrapClass">
           <img
-            v-if="displayImage(item)"
-            :src="displayImage(item)"
+            v-if="artSrc(item)"
+            :src="artSrc(item)"
             :alt="item.title"
             class="absolute inset-0 h-full w-full object-cover transition duration-500 group-hover:scale-105"
+            @error="onArtError(artSrc(item))"
           />
           <!-- Art-absent placeholder. Without this an unillustrated row renders
                <img src="">, which browsers treat as "reload the current page"
                and paint as a broken image. Galleries whose rows are routinely
                unillustrated (facets, where art is queued rather than required)
-               could not adopt this shell at all until it degraded properly. -->
+               could not adopt this shell at all until it degraded properly.
+
+               It now also covers art that is PRESENT but fails to load, which
+               is a different failure with the same remedy: a broken <img>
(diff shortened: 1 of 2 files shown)


silasfelinus:
Reviewed. The `kr-gallery.vue` fix is correct and directly resolves the t-069 symptom: the `audit` run against the live preview shows `/conductor` down to a single broken image (`interface-vision-icon.webp`, 44×44, `alt=""`) — a different, lower-stakes mechanism (empty alt, no text sprawl) than the reported "description text renders over the card art" defect, which no longer reproduces on any of the three viewports checked.

However, `audit` failed overall and `mergeable_state` has stayed `unstable` across five merge attempts (mixing 502s and "merge already in progress" — consistent with a required check blocking rather than transient infra). The reason: adding `/conductor` and `BROKEN-ART` to the default route list surfaced pre-existing, much larger findings unrelated to this diff — `/bots` (62–142 images), `/characters` (142–211), `/rewards` (142–231), and `/dreams` (5–16) all report broken/never-loaded art across every viewport. These are `never loaded` (stalled past networkidle), not `failed` (404/error) — this PR's fix only wires an `@error` handler, so it can't catch a stall, and that gap is real: fixing it requires either a load-timeout fallback in `kr-gallery.vue` or root-causing why `/api/art/images/{id}/file` isn't resolving for that many images under audit conditions (rate limiting? cold-start latency? a real serving bug?). That's a materially bigger investigation than this PR's scope (t-069 was specifically the `/conductor` card-text defect).

I don't want to guess at a fix for the broader issue inline here, and I can't merge past the newly-required `audit` gate as-is. Filing a follow-up task (interface-vision, broken/stalled art across `/bots` `/characters` `/rewards` `/dreams`) to scope that investigation separately, and leaving this PR's roadmap task at a soft `needs-human` pointing here so Silas can either merge manually (if `audit` isn't meant to hard-block yet) or confirm the follow-up is the right next step.

---

PR #1396, opened 2026-08-04
t-064: dream-card onto kr-entity-card-body, 5/5

silasfelinus:
The last object card still rolling its own markup below the shared shell. `verifyGalleryAdoption` now reports **5/5**; `dream-card` 635 → 596 lines.

## Visible changes, and they're the point

> "the final version obviously is wanting a single display that is consistent across galleries, all with those bells and whistles."

Dream was the one card that **didn't** look like the others — a full-bleed poster with the title and count chips overlaid on the art, held up by `:padded="false"` on the shell. That prop is gone with the bespoke markup, so the body insets its art and owns its padding exactly as bot, character, reward and scenario do.

- **Dream type becomes the subtitle** — the slot `reward-card` put `rewardType` in. Deliberately no longer *also* a badge: it used to be a badge and nothing else, and rendering it in both places would repeat the subtitle two centimetres above itself.
- **The scenario/cast/art/item counts move from an overlay on the artwork to the meta row** on the card surface, where every other object card puts them — legible against a solid background instead of against whatever the art happens to be.
- **Badges keep only true state**: Mature, and Private when `isPublic` is false.
- **`description` now uses the same precedence `dream-gallery`'s own helper uses** (`pitch → description → flavorText`), so a card and the list it came from stop describing a Dream from different fields.

## Removed

`figureClass` — the per-variant aspect I added two changes ago. `kr-entity-card-body` owns shape via its own `shape`/`compactShape` props, and a second copy is exactly the duplication this task exists to delete. The `variant` prop stays and now feeds the body, so the Cards/Heroes/Icons toggle still means something.

## Also: what BROKEN-ART is for

> "we have a massive backlog of images generating, and an auto image generator if the image is missing, don't worry about it"

That reframes the check, and the file said the wrong thing. **It isn't "find missing art"** — missing art is expected and self-healing. A component that degrades properly renders a placeholder and trips nothing, so **a hit means "this component doesn't handle missing art"** — a code defect, not a content one.

Which is why it's worth failing a build over even mid-backlog, and why `t-069`'s fix was a placeholder rather than a scramble to find missing files. Documented, including: don't "fix" a hit by generating the image.

## Verification

`vue-tsc` clean; `test:gallery-adoption` reports **5/5**; `test:layout-contract`, `test:channel-content`, `test:nav-manifest` pass; prettier clean. `eslint` clean apart from a pre-existing `no-empty`.

**Not verified: rendered output.** `/dreams` changes visibly and this environment can't render it with data — but this PR's own `deployment_status` audit measures it against the preview. Card convergence is exactly the kind of change worth looking at rather than assuming, so `/dreams` on the preview is the real check.

---

### DIFF
--- a/components/dreams/dream-card.vue
+++ b/components/dreams/dream-card.vue
@@ -1,16 +1,14 @@
 <!-- /components/dreams/dream-card.vue -->
 <template>
-  <!-- The shared card shell (interface-vision t-031). Dreams were the only one
-       of the five object cards still rolling their own <article>, which is why
-       they had no reaction or review affordance at all while Bot, Character,
-       Reward and Scenario did -- `dream` has been a valid ReactionTargetType and
-       `DREAM` a valid category the whole time; nothing ever wired it up.
-       :padded="false" keeps the full-bleed poster look this card is built
-       around; the other four inset their art and take the default padding. -->
+  <!-- The shared card shell (interface-vision t-031) plus the shared card BODY
+       (t-064). Dream was the last of the five object cards still rolling its
+       own markup below the shell: 5/5 now share kr-entity-card-body.
+       :padded="false" is GONE with the bespoke poster -- the body insets its
+       own art and owns its padding, exactly as the other four do, which is
+       what "a single display that is consistent across galleries" means. -->
   <reactable-card
     :selected="activeSelected"
     :compact="compact"
-    :padded="false"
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: claude-20260804T0305Z-review-t064 at 2026-08-04T03:05:38Z

---

PR #1404, opened 2026-08-04
Header: unclip the channel menu, and make tabs read as buttons

silasfelinus:
Two reports from Silas:

> I can't actually select the channel option anymore. Love love love the new tab
> selections, but if there is room, they should look like buttons, not just a
> string of strings with no divisions unless you hover

## 1. The channel menu was clipped to nothing — my regression

To make the channel picker and tab strip read as one control, I wrapped them in
a shared bordered div with `overflow-hidden`. channel-select's daisyUI dropdown
is absolutely positioned **inside** that shell and is ~318px tall. The shell is
**40px**. Everything below the first row was clipped away — not just invisible,
but un-hittable.

Measured on the running app before the fix:

```
menuFound:          true
menuVisible:        "visible"
menuRect:           352 x 318      <- laid out fine
firstItemRect:      262 x 48       <- real box
firstItemClickable: false          <- but nothing lands on it
firstItemHit:       h1.text-lg...  <- the page heading, behind the menu
clippingAncestors:  div.flex.h-10.min-h-10 overflow=hidden/hidden/hidden
```

The shell keeps its rounded border; the children round their own outer corners
instead (`rounded-l-xl` on channel-select when `seamless`, `rounded-r-xl` on the
tab strip).

**Verified end to end, not just by hit-test** — "the element is hittable" is not
"selecting a channel works":

- opening the picker and clicking "Plan / Projects" navigates to `/conductor`
- the channel button reads "Plan" afterwards, on both a direct load and an
  in-app navigation

## 2. Tabs read as run-on text

Inactive tabs had no resting affordance at all — no border, no background, only
`hover:bg-base-200`. So they looked like a string of words, and only looked
clickable once a pointer was on them, which on touch is never.

| | before | after |
|---|---|---|
| inactive | no border, transparent | 1px `base-300` on `base-100` |
| active | `bg-primary` | `border-primary bg-primary` |
| gap | `0.5` (2px) | `1` (4px), `1.5` at `sm` |

Costs 1px per tab. The responsive audit reports **0 SPILL / 0 CRUSHED / no
horizontal scroll** at 390, 820 and 1440 across `/`, `/conductor`, `/dreams`.

## Verification

- `npm run test:layout-contract` — holds, no new violations
- `npm run test:channel-content` — 6 channels, 63 tabs, 0 nav manifest errors
- `npm run audit:responsive` — 0 real layout defects (remaining hits are the
  documented local-only BROKEN-ART noise from `public/images` not being in the repo)
- `npx eslint` + `prettier --check` on touched files — clean

---

### DIFF
--- a/components/navigation/channel-select.vue
+++ b/components/navigation/channel-select.vue
@@ -7,7 +7,7 @@
       class="flex h-full min-h-10 shrink-0 items-center gap-2 overflow-hidden px-2 transition-all xl:gap-2.5 xl:px-3"
       :class="
         seamless
-          ? 'hover:bg-base-200'
+          ? 'rounded-l-xl hover:bg-base-200'
           : 'h-10 rounded-xl border border-base-300 bg-base-100 shadow-sm hover:-translate-y-0.5 hover:shadow-md sm:h-11 sm:min-h-11 xl:h-14 xl:min-h-14'
       "
       :title="`Navigate ${activeChannel.label}`"
--- a/components/navigation/workspace-header.vue
+++ b/components/navigation/workspace-header.vue
@@ -26,25 +26,33 @@
            channel selector doesn't drop awkwardly below the current tab
            title." Nothing is lost by dropping the old title: the active tab's
            own label WAS the title, and it is now something you can act on. -->
+      <!-- NO `overflow-hidden` on this shell, however much it wants it for the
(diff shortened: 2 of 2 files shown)


silasfelinus:
REVIEWING: 2026-08-04T051636Z-pr1404-f8k2 at 2026-08-04T05:16:36Z

PR #1408, opened 2026-08-04
refactor(scenarios): scenario-interact.vue onto the shared conversation kit

silasfelinus:
### Task
interface-vision/t-088: Phase 5, `scenario-interact.vue` onto the shared conversation kit (kind: software)

Split from t-051 — the last of the three surfaces (`character-interact.vue` landed via #1400, `reward-interact.vue` via #1406/t-087). The task's own note flags this one as **not a mechanical swap**, unlike the other two, and asks for preview verification before merge.

Two distinct duplications, handled separately:

1. **The story transcript** (chat bubbles, `.story-dot` streaming placeholder, inline "what do you do next?" reply-option buttons) becomes `<kr-chat-window>` fed a `chatTurns` computed. `storyStore.storyDisplayChats` already parses display text and reply options server-side-ish (in the store), so this surface's mapping is simpler than reward-interact's — no re-parsing needed. `storyLogRef`/`scrollToBottom()` and the manual scroll watcher are gone; `kr-chat-window` owns its own auto-scroll. The `.story-dot`/`.story-streaming` keyframes are gone too.
2. **The "How does it begin?" intro picker** duplicated the job of the already-deleted `choice-selector.vue`/`choice-preview.vue`, per the task note. Its body text is a full opening paragraph, and `kr-choice-list`'s `hint` slot renders italic — exactly the "wrong for a long opening paragraph" problem the note flagged. Rather than accept that regression, I added a `hintProse` prop to `kr-choice-list.vue` (default `false`, so every other existing caller — bot-interact, workspace-narrator, kr-narrator-stage, narrative-response-composer, reward-interact, taskmaster-page — is unaffected) that swaps the italic style for the same `.kr-prose` measure/leading `kr-chat-window`'s own `prose` prop already uses.

Calling `kr-choice-list` directly for the intro picker (rather than through `kr-chat-window`'s fixed wrapper) also let it keep its "highlight the picked opening" styling, via `kr-choice-list`'s existing `selectedKey` prop — a capability `reward-interact.vue` couldn't reach because `kr-chat-window` doesn't forward a per-turn `selectedKey` through its embedded choice lists.

### How I verified
- `vue-tsc --noEmit` — clean.
- `eslint` on both touched files — clean.
- `npm run test:narrative-kit` — conversation-kit adoption is now **7/7** for the originally-surveyed surfaces.
- `npm run test:layout-contract` — the one-scroll-region count drops by 1; ratcheted the baseline via `--update`.
- Vercel preview: verifying the intro picker and story transcript render correctly before merging (see below).

### Stakes
reversible

### Flags for Reviewer
- Two accepted, documented visual changes, consistent with the same tradeoffs already accepted for `character-interact.vue`/`reward-interact.vue`: the story-transcript reply options lose their "highlight the last pick" styling (no per-turn `selectedKey` from `kr-chat-window` — same kaizen note as t-087), and the transcript's `text-smart` dynamic type-scale classes are replaced by `kr-chat-window`'s fixed sizing.
- `kr-choice-list.vue`'s new `hintProse` prop is additive/opt-in only — grepped every current caller to confirm none pass it, so this can't change existing rendering anywhere else.

### Kaizen suggestion
Same as t-087's: `kr-chat-window` has no way for a caller to pass a per-turn `selectedKey` through to its embedded `kr-choice-list`. Two surfaces now have this exact gap (reward-interact, scenario-interact) — worth promoting from "deferred" to an actual small task if a third surface hits it.

### Notes for reviewer
Roadmap task `interface-vision/t-088` is at `status: review`, claimed by session `2026-08-04T061400Z-t088`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TaRK67TyfLGw6pjL1JkPP8)_

### DIFF
--- a/components/narrative/kr-choice-list.vue
+++ b/components/narrative/kr-choice-list.vue
@@ -73,7 +73,8 @@
         </span>
         <span
           v-if="choice.hint"
-          class="mt-0.5 block text-sm italic text-base-content/55"
+          class="mt-0.5 block text-sm text-base-content/55"
+          :class="hintProse ? 'kr-prose not-italic' : 'italic'"
         >
           {{ choice.hint }}
         </span>
@@ -101,6 +102,14 @@ withDefaults(
     disabled?: boolean
     showIndex?: boolean
     label?: string
+    /**
+     * A hint is italic by default -- right for a short aside, wrong for a
+     * long opening paragraph (scenario-interact.vue's intro picker, which
+     * used to render its own non-italic body text). Opt in per list rather
+     * than changing the default, since every existing caller relies on the
+     * italic look for short hints.
+     */
+    hintProse?: boolean
   }>(),
   {
     layout: 'stack',
@@ -109,6 +118,7 @@ withDefaults(
     disabled: false,
     showIndex: true,
     label: 'Choices',
+    hintProse: false,
   },
 )
 
--- a/components/scenarios/scenario-interact.vue
+++ b/components/scenarios/scenario-interact.vue
@@ -122,36 +122,16 @@
           Pick an opening, or write your own direction below.
         </p>
 
-        <div class="mt-3 grid grid-cols-1 gap-3 lg:grid-cols-2">
-          <button
-            v-for="(intro, index) in introChoices"
-            :key="`${intro.label}-${index}`"
-            type="button"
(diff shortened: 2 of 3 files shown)


silasfelinus:
REVIEWING: claude-agent-run-20260804T0630Z at 2026-08-04T06:30:27Z

---

PR #1412, opened 2026-08-04
fix(db): stop Vercel preview pools exhausting ProxySQL

silasfelinus:
## Root cause

The two-host census and Vercel runtime inspection established that the 200-connection incident is **Vercel function fan-out**, not 200 MariaDB backends and not SILAS-PC.

### Live connection layers

- ProxySQL `kindrobot` frontend sessions: **200 / 200**
- ProxySQL denied-user connections: **95,877**
- ProxySQL client connections created: **107,000**
- hostgroup 10 backend cap: **40**
- second backend snapshot: **27 used + 4 free**
- MariaDB `kindrobot` sessions: **31**, all sleeping and all from ProxySQL
- active transactions: **0**
- SILAS-PC connections to Alexandria: **0**

ProxySQL is successfully multiplexing the cloud clients onto a bounded MariaDB pool. The outage occurs when it refuses frontend connection 201.

### Client ownership

The visible ProxySQL client sources are public AWS us-east addresses. The Vercel project simultaneously shows:

- the Nuxt server and `/api/*` routes deployed as Node functions;
- production and preview deployments using the real database path;
- a fresh independently warm deployment for nearly every pushed commit;
- thousands of Prisma pool failures across the deployed API surface.

The global pool defaults forced every Vercel function process to use `connectionLimit=10`, `minimumIdle=1`, and `idleTimeout=300`. The mandatory warm connection allowed idle sessions from preview and production lambdas to accumulate until the shared ProxySQL frontend limit remained full.

## Changes

- split database pool defaults by runtime:
  - **long-lived local process:** limit 10, minimum idle 1, idle 300s
  - **Vercel function:** limit 2, minimum idle 0, idle 15s
- retain explicit `DATABASE_*` environment overrides
- keep one Prisma client/pool per Node process
- remove replacement-client generations that could strand additional pools
- log the active pool profile at runtime
- retain bounded retries, transaction protection, circuit breaker, TLS, text protocol, and `pipelining=false`
- enforce both runtime profiles in contract tests
- confirm Vercel previews skip migrations and database seeds
- add read-only Windows and Alexandria connection censuses
- document the live evidence and recovery order

## Recovery after merge

1. Let the production Vercel deployment become READY and confirm it reports `poolProfile: vercel-function`.
2. Restart ProxySQL once to clear sessions retained by old deployment code.
3. Re-run the Alexandria census and verify frontend sessions fall and stay below 200.

The ProxySQL frontend limit remains unchanged during observation. Raising it before removing permanent warm preview sessions would only enlarge the bucket.

## Durable follow-up

Vercel Preview should eventually use a separate database/account, or production `DATABASE_URL` should be removed from the Preview environment. Preview deployments should not share the production runtime user's full connection budget.


silasfelinus:
REVIEWING: 2026-08-04T081503Z-pr1412-x7q4 at 2026-08-04T08:15:03Z

PR #1415, opened 2026-08-04
Fix the small-card/dead-mode-bar defects, contract them, and adopt bot-gallery (4/7)

silasfelinus:
Three commits. Safe to merge at any point — each is independently complete.

**#1414 merged at its first commit only**, so its two follow-ups never reached
`main`. The gallery Silas looked at was the broken version. Commits 1–2 are that
fix; commit 3 is the next gallery.

---

## 1. The three defects — `8741fc0`

> *"small card, excessive padding, no different layouts"*

**Small cards, the root cause.** `MODE_GRID_CLASS` used Tailwind **viewport**
breakpoints. These galleries mount inside **manager panels**, so on a 1440px
screen `xl:grid-cols-4` fired while the container was 320px. Measured, viewport
pinned at 1440px, only the container varying:

| panel | before | after |
| --- | --- | --- |
| 320px | **4 cols / 71px cards** | 1 col / 320px |
| 480px | 4 cols / 111px | 2 cols / 234px |
| 1100px | 4 cols / 266px | 5 cols / 210px |

Now `auto-fill` + `minmax`, which needs no breakpoints. Also silently fixes
`facet-gallery` and `conductor-project-gallery-page`, already on the shell and
suffering the same thing.

**No different layouts** — two breaks: `scenario-card` hardcoded `variant="card"`
with no prop, and I bound `:modes="[]"` while never binding `:mode`.

**Padding** — cards grid `gap-4` → `gap-3`.

## 2. The guard — `eb41d31`

> *"can we make sure that when we do the other models we aren't repeating this
> error… we seem to keep hitting the same walls because we aren't expecting a
> consistent style"*

`verifyGalleryConsistency.ts`, wired into `contract-tests.yml`:

| # | check |
| --- | --- |
| 1 | grids stay container-responsive (no `sm:`/`lg:`/`xl:`, keeps `auto-fill`) |
| 2 | object cards accept `variant`, never hardcode it |
| 3 | a visible mode bar is actually bound to `:mode` |

`verifyGalleryAdoption` counts *whether* a gallery adopted the shell; this
asserts the adoption is **worth having**.

**It earned itself twice.** Check 2 caught `bot-card`, `character-card` and
`reward-card` all hardcoding `variant="card"` — the identical defect already
waiting in three of the four remaining galleries. And it caught its own first
bug: flagging `scenario-card` for a `variant="card"` that existed only in the doc
comment explaining it *used to* — interface-vision **t-068** live, so every scan
strips comments first.

## 3. bot-gallery adopts — `1d348e7` — **3/7 → 4/7**

Grid variant through kr-gallery, `bot-card` kept via the `item` slot, `row` and
`dropdown` left bespoke. Mode bound both ways and fed through `MODE_VARIANT`.

That last part **only works because check 2 had already fixed `bot-card`** —
wiring a mode bar into a card that ignores it is precisely the dead control from
the first pass. This time the check ran before the PR instead of after.

---

## Verification

- `test:gallery-consistency` — passes, 4 kr-gallery mounts all bind `:mode` or opt out
- `test:gallery-adoption` — **4/7**, `bot-gallery` reached from `/academy`
- `test:layout-contract` — holds · `test:workflow-paths` — 720 refs across 62 workflows
- `eslint` + `prettier` clean (the `contract-tests.yml` prettier warning is pre-existing — fails on clean `HEAD` too)

Remaining: `character-gallery` (717) · `reward-gallery` (741) · `dream-gallery` (1035).

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -52,6 +52,9 @@ jobs:
       - name: Gallery adoption contract
         run: npm run test:gallery-adoption
 
+      - name: Gallery consistency contract
+        run: npm run test:gallery-consistency
+
       - name: Data surface manifest contract
         run: npm run test:data-surface-manifest
 
--- a/components/bots/bot-card.vue
+++ b/components/bots/bot-card.vue
@@ -50,7 +50,7 @@
         :title="botTitle"
         :subtitle="bot.subtitle || ''"
         :source="bot"
-        variant="card"
+        :variant="variant"
         :fallback="artFallbackSrc"
         :show-image="showImage"
         :show-description="false"
@@ -167,6 +167,7 @@ import type { Bot } from '~/prisma/generated/prisma/client'
 import { useBotStore } from '@/stores/botStore'
 import { useNavStore } from '@/stores/navStore'
 import { useUserStore } from '@/stores/userStore'
+import type { ArtVariant } from '@/utils/artImageSrc'
 import type { EntityCardChip } from '@/components/gallery/kr-entity-card-body.vue'
 
 const props = withDefaults(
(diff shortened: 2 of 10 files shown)


silasfelinus:
REVIEWING: 2026-08-04T081650Z-pr1415-h7q2 at 2026-08-04T08:17:30Z

PR #1418, opened 2026-08-04
t-060: character, reward and dream galleries adopt the shell — 7/7, beta gate closed

silasfelinus:
The last three. **`core-object galleries on kr-gallery: 7/7`** — DESIGN-BRIEF's
*"all seven core objects share one gallery"* is closed, every one reachable from
a real route.

```
✓ bot-gallery.vue                      reached from /academy
✓ character-gallery.vue                reached from /academy
✓ dream-gallery.vue                    reached from /academy
✓ facet-gallery.vue                    reached from /facets
✓ reward-gallery.vue                   reached from /academy
✓ scenario-gallery.vue                 reached from /academy
✓ conductor-project-gallery-page.vue   reached from /conductor
```

Three commits, each independently complete and safe to merge at any point.

## The shape, applied three more times

Grid variant renders through `kr-gallery` (mode bar, skeleton/error/empty,
container-responsive grid); the object card stays the card via the `item` slot,
so t-064's reactions, earned karma and edit/clone/archive actions survive the
adoption. `row` and `dropdown` stay bespoke — a horizontal filmstrip and a
`<select>` are not grids, and forcing them through a grid shell would be adoption
for the counter's sake rather than the user's.

Each gallery gets a `<x>CardProps` computed holding the card's shared props, so
the bespoke row layout and the kr-gallery slot can't drift when one gains another.

## Per-gallery notes

**character** — `image-fit="contain"` is carried into the shared props; characters
are portraits and the default `cover` crops them.

**reward** — `earnedKarma` stays an explicit *per-item* binding rather than
joining the shared props, because it's a per-reward-id lookup. Folding it in
would have quietly given every card the same karma.

**dream** — least work of the five despite being the largest file: it already
drove `MODE_GRID_CLASS` and `dream-card`'s `variant` itself, so adopting the shell
is mostly handing over markup it was duplicating. It keeps its **own** mode bar in
the toolbar, so this mount binds `:mode` **and** passes `:modes="[]"` — the shell
needs the mode to pick the right grid, but a second bar for the same state would
be worse than none. Check 3 allows exactly this: it asks that a *visible* bar be
wired, not that every mount show one.

## The contract earned its keep again

`vue-tsc` caught a real bug in the reward slice before it could ship: the gallery
item title read `reward.text`, and `Reward` has no such column (it's `name`).
That's precisely the mistake three near-identical migrations invite, and the
reason each got its own typecheck rather than being pattern-matched through.

## Verification

- `test:gallery-consistency` — passes, **7** kr-gallery mounts all bind `:mode` or opt out
- `test:gallery-adoption` — **7/7** core objects, cards still 5/5
- `test:layout-contract` — holds · `test:channel-content` — 0 nav manifest errors
- `prettier` clean

Pre-existing lint left untouched and confirmed against clean `HEAD`: an unused
`characterGenres` in character-gallery, and three `unified-signatures` errors on
dream-gallery's emit overloads.

## What's left of the counter

Overall adoption is 7/24. The other seventeen are non-core galleries, and four of
those (`butterfly`, `lora`, `model`, `lab`) have **no content route rendering
them** — adopting the shell there changes nothing a user can see.

---

### DIFF
--- a/components/characters/character-gallery.vue
+++ b/components/characters/character-gallery.vue
@@ -236,29 +236,45 @@
         </button>
       </div>
 
-      <div v-else :class="layoutClass">
+      <!-- Row stays bespoke: a horizontal filmstrip is not one of kr-gallery's
+           cards/heroes/icons grids. -->
+      <div v-else-if="isRowMode" :class="layoutClass">
         <character-card
           v-for="character in filteredCharacters"
           :key="character.id"
           :character="character"
           :selected="characterStore.selectedCharacter?.id === character.id"
           :is-selected="characterStore.selectedCharacter?.id === character.id"
-          :compact="isCompact"
-          :show-image="showImages"
-          :show-actions="showCardActions"
-          :show-description="showDescriptions"
-          :show-meta="showMeta"
-          :show-stats="showStats"
-          :show-debug="showDebug"
-          :allow-edit="allowEdit"
-          :allow-clone="allowClone"
-          :allow-delete="allowDelete"
-          image-fit="contain"
+          v-bind="characterCardProps"
           @edit="startEditingCharacterById"
           @clone="cloneCharacterById"
           @delete="handleCharacterDeleted"
         />
       </div>
+
+      <!-- t-060: the shared shell owns the grid and the Cards/Heroes/Icons bar;
+           character-card stays the card. -->
+      <kr-gallery
+        v-else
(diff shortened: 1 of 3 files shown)


silasfelinus:
REVIEWING: 2026-08-04T091349Z-pr1418-x7p3 at 2026-08-04T09:13:49Z

PR #1426, opened 2026-08-04
conductor-page onto the shared shell — the allow-list is empty, so it's gone

silasfelinus:
The last hand-rolled gallery control.

`verifyGalleryConsistency` check 5 found it the moment that check existed:
`conductor-page` didn't mount kr-gallery at all — it rendered **its own**
Cards/Heroes/Icons bar plus three per-mode layouts. That was the **third** copy
of the same control, which is the full shape of:

> *"I wouldn't be clicking the layout options on one side of the screen for one
> and the other for, well, you know you can extrapolate."*

## Quiet corroboration of the spec

Its hand-rolled layouts had the shapes **right** — square icons, 16:9 heroes, 2:3
cards. Silas on the spec: *"I never said it, but it feels like a true true."* Two
independent implementations arriving at the same four shapes is what a true true
looks like. The shell draws them now; **~80 lines of duplicated markup go**.

It passes pre-resolved `icon`/`card`/`hero` URLs rather than a `source`, because
these projects resolve art from slug-derived paths with a DB column taking
precedence — something `resolveArtVariantSrc` can't know about, and exactly the
case `GalleryItem`'s per-variant fields exist for.

## The allow-list never grew

I filed this component in a documented allow-list when the check first caught it,
with the note *"this list must not grow."* It didn't. The entry is removed here
**along with the allow-list itself**, because there's nothing left to list.

## Two bugs found on the way

**`projectIconPath` read `imagePath`, not `iconPath`** — because until this
morning Project had no `iconPath` column and the square default was the nearest
thing. Both are square, so the substitution was invisible. That's why the schema
audit found it and a bug report never would have.

**The generated Prisma client was stale on `main`.** #1421 landed the schema
change without regenerating, and `prisma/generated` is *tracked* here — so `main`
had `iconPath` in the schema but not in the client, and anything typechecking
against `Project.iconPath` failed. CI hides this by generating during the run; a
fresh local checkout doesn't. Regenerated here.

## Verification

- `test:gallery-consistency` — all **seven** checks pass, **8** kr-gallery mounts
- `test:layout-contract` — holds
- `test:channel-content` — 0 nav manifest errors
- `eslint` + `prettier` clean

---


silasfelinus:
REVIEWING: 2026-08-04T111406Z-pr1426-a7k9 at 2026-08-04T11:14:06Z

silasfelinus:
Review result: **not safe to merge on the current head.**

Two blocking issues:

1. The PR turns previously diagnostic pipelines into hard failures without resolving the debt they expose. `Contract Tests` run `30902628146` fails at **WonderLab preview coverage no-regression gate** with 29 uncovered required-prop components. `Facet Catalog Contract` run `30902628249` fails at **Enforce illustrated Builder Facet coverage**, reporting the existing Builder artwork backlog. The new `pipefail` workflow edits therefore make this UI PR red. Remove those workflow/pipefail changes from this PR, or resolve that debt separately before reintroducing them.

2. The diff is not reviewably scoped and is currently unmergeable. The intended conductor gallery migration is accompanied by generated Prisma churn, a 5,100-line `adventureCards.ts` rewrite with no stated semantic purpose, workflow changes, a package-script change, and a new verifier. The branch is four commits behind `main`, and GitHub reports it non-mergeable. Merge current `main` normally, preserve canonical main content, and narrow the PR to the conductor-page migration plus only the directly required gallery-consistency update. Do not force-push.

The core gallery change itself is directionally sound: it removes the duplicate mode bar, feeds the three pre-resolved art variants into `kr-gallery`, and corrects `projectIconPath` to prefer `iconPath`. Once the unrelated red-CI and diff-scope issues are removed, this should be straightforward to re-review.

I could not submit a formal REQUEST_CHANGES review because GitHub treats this connector identity as the PR author, so this comment records the blocking review instead.

silasfelinus:
## CI triage: 3 red checks, 1 fixed here, 2 pre-existing/unrelated

Went through the three failing checks individually (re-fetched current state, read actual job logs, and diffed the workflow files against `main`).

### 1. `facet-catalog` — fixed here (pushed 3ac8c82)

This PR's own second commit (632a0a3, t-063) fixed a real `pipefail` bug across two workflows so previously-inert checks would actually enforce. Its message says the `facet-catalog` hard-fail step for missing Builder art was removed in favor of `continue-on-error` (Silas, 2026-08-04: "there are still 600+ or so images in the queue" — not a blocker). The code didn't match the message: the `Enforce illustrated Builder Facet coverage` step (`if: steps.builder-coverage.outcome == 'failure'; run: exit 1`) was never actually deleted from `facet-catalog-contract.yml`. Before the pipefail fix that condition never evaluated true, so it was silent dead code; the pipefail fix (in the *same* commit) is what made it fire for real, immediately breaking this PR's own check. Deleted the leftover step to match the commit's stated intent — a genuine regression from this PR's own diff, now fixed.

### 2. `audit` (responsive layout) — pre-existing, unrelated to this diff

`/rewards` on desktop has 3 broken `` tags under `public/images/rewards/{skill,item}/...webp`. This PR's diff touches `components/pages/conductor-page.vue`, `stores/helpers/adventureCards.ts`, Prisma generated files, and CI/verifier scripts — nothing under `public/images/rewards/` or any reward-image resolution path. Not something this PR's diff could have caused; shouldn't block it.

### 3. `Contract verifiers` (`audit:wonderlab-previews --strict`) — pre-existing, unrelated to this diff

29 components (including `components/gallery/kr-gallery.vue: missing fixture props items`) lack WonderLab preview-fixture coverage. None of those 29 components, including `kr-gallery.vue`, are touched by this PR's diff — the strict-audit step's command is byte-identical between this branch and `main` (only gaining `shell: bash` for the same pipefail fix described above). The real reason this is red here and green on `main` right now: `main`'s copy of that step *still has the pipefail bug*, so its `| tee` swallows the script's real exit code and reports false success — main is just as non-compliant, it's only measured correctly on this branch. This is the same class of latent, pre-existing gap as #2, newly visible only because this PR's own CI-hygiene commit fixed the measurement, not because the gallery-shell migration created new gaps. Recommend a separate follow-up task to backfill fixture coverage across those 29 components; out of scope here.

With #1 fixed and #2/#3 confirmed pre-existing and unrelated to this PR's actual diff, and every other required check (TypeScript, layout-contract, verify ×3, etc.) already green, this should be mergeable once the pushed fix goes green.

---

PR #1447, opened 2026-08-04
Join the tailnet before talking to the database

silasfelinus:
The reward-path repair workflow's first run failed on a pool timeout — `active=0 idle=0` after 10s — which reads like a dead database but isn't.

The database lives behind Tailscale on the Unraid box, and `DATABASE_URL`'s host is a tailnet address, so a bare runner has **no route to it at all**.

`fallback-snapshot.yml` already knew this — it joins the tailnet as an ephemeral node before its snapshot step, and its header says outright that the host must be the tailnet address, *not* the public hostname. The two dispatch-only maintenance workflows inherited only its `DATABASE_URL`/CA env block and not the Tailscale step, so **both were unrunnable**. `retire-wonderlab-components.yml` had the same gap and had never been run to find out.

Both now:
- join the tailnet after the npm/prisma steps (keeping the session short — same placement `fallback-snapshot.yml` uses)
- **fail fast** with a pointed message when the OAuth secrets are missing, instead of surfacing it ten seconds later as a connection-pool timeout

Verified: `test:workflow-pipefail`, `test:workflow-paths` (65 workflows, 756 path references), Prettier — all clean.

**Requires** `TS_OAUTH_CLIENT_ID` and `TS_OAUTH_SECRET` repository Actions secrets, the same pair `fallback-snapshot.yml` uses — which it has been using successfully on its daily schedule, so they exist.

---

### DIFF
--- a/.github/workflows/repair-reward-image-paths.yml
+++ b/.github/workflows/repair-reward-image-paths.yml
@@ -33,10 +33,17 @@ jobs:
       DATABASE_SSL_CA_BASE64: ${{ secrets.DATABASE_SSL_CA_BASE64 }}
       DATABASE_SSL_CA: ${{ secrets.DATABASE_SSL_CA }}
     steps:
-      - name: Require the database secret
+      - name: Require the database and tailnet secrets
+        env:
+          TS_OAUTH_CLIENT_ID: ${{ secrets.TS_OAUTH_CLIENT_ID }}
+          TS_OAUTH_SECRET: ${{ secrets.TS_OAUTH_SECRET }}
         run: |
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: 2026-08-04T161357-0700-pr1447-k3p8 at 2026-08-04T23:13:57Z

silasfelinus:
Superseded by #1449. GitHub blocked this otherwise-green head because branch protection now requires the newer `verify` status check. I replayed the exact two-file change onto current `main` without force-pushing so all current required checks can run.

PR #1449, opened 2026-08-04
Join database maintenance workflows to the tailnet

silasfelinus:
Replacement for #1447, replayed onto current `main` so the newly required `verify` status check can run.

## What changed
- add the same ephemeral `tailscale/github-action@v3` connection used by `fallback-snapshot.yml` to both dispatch-only database maintenance workflows
- fail fast when `TS_OAUTH_CLIENT_ID` or `TS_OAUTH_SECRET` is absent
- preserve `workflow_dispatch`-only execution and `dry_run: true` defaults

## Why
Both workflows use a `DATABASE_URL` whose host is reachable only over the tailnet. Without joining Tailscale, Prisma reports a misleading 10-second pool timeout (`active=0 idle=0`).

## Review provenance
The original head `052ce88e3ba671c2ea60264eb321056661cdfd34` passed 11 Actions workflows, but GitHub blocked merging because current branch protection now requires a `verify` check that did not exist on that older head. This branch contains the same two-file change on current main.

No workflow was dispatched and no production data was changed.

### DIFF
--- a/.github/workflows/api-client-followups-contract.yml
+++ b/.github/workflows/api-client-followups-contract.yml
@@ -10,15 +10,18 @@ on:
       - "utils/scripts/verifyApiClientFollowups.ts"
       - ".github/workflows/api-client-followups-contract.yml"
       - "package.json"
+  # NO paths filter, deliberately. This job is named `verify`, and `verify` is a
+  # REQUIRED status check on main -- so a PR that touches none of the paths below
+  # never produces it and can never merge, no matter how green everything else
+  # is. A workflows-only PR hit exactly that and was stuck until this was fixed
+  # (dispatching the workflow by hand does not satisfy the rule either; the rule
+  # wants the check from the PR's own event).
+  #
+  # A required check must be unconditional. The push trigger keeps its filter --
+  # nothing requires it there -- and this contract is a ~40s text scan, so
+  # running it on every PR costs nothing worth saving.
   pull_request:
     branches: [main]
-    paths:
-      - "server/api/art/**"
-      - "stores/**"
-      - "components/**"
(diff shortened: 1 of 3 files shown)


silasfelinus:
All 11 GitHub Actions workflows completed successfully and Vercel is Ready. Merge is still blocked by repository rules because required status check `verify` is **expected but no check with that context was created** on this head. I did not bypass or weaken the rule. This appears to be a branch-protection/check-registration defect rather than a failure in this two-file workflow change.

silasfelinus:
Pushed one commit to this branch (`68ad468`), because replaying onto current `main` doesn't actually unblock it — this PR is still `blocked` at 14/14 green with **no `verify` check at all**.

The base commit was never the cause. `verify` comes from `api-client-followups-contract.yml`, whose `pull_request` trigger is path-filtered to `server/api/art/**`, `stores/**`, `components/**`, its own verifier, its own workflow file, and `package.json`. A PR touching only `.github/workflows/*.yml` matches none of them, so the check has no way to exist on any head, old or new. (#1447 passed because it also touched `package.json`; that's the only reason it ever ran there.)

Dispatching the workflow by hand doesn't help either — I tried it on #1447 and the run went green on the exact head SHA, but the rule still reported `verify` as expected, so it wants the check from the PR's own event.

The commit drops the `paths:` filter from the `pull_request` trigger only. Adding `.github/workflows/**` to the list would have fixed today's PR and left the trap armed for whichever PR misses the list next — a required check has to be unconditional. The `push` filter is untouched (nothing requires it there), and the contract is a ~40s text scan, so running it on every PR costs nothing worth saving.

---

PR #1457, opened 2026-08-05
t-100: catch stored art paths that serve the app shell at HTTP 200

silasfelinus:
## The gap

A missing asset under a non-`/images` path **does not 404**. It falls through to the SPA catch-all and returns **200 with a 142 KB Nuxt app shell**.

That is how all 227 `Reward.imagePath` values could be rooted at `/rewards/` instead of `/images/rewards/`, with 214 serving HTML, while curl, uptime monitors, and link checkers all read them as healthy (#1446).

The responsive-layout audit couldn't see it either — and not because it was broken. `kr-art-plate` caught the decode failure and swapped in its fallback, exactly as designed. **The page looked fine while the art was wrong.** "Did the UI break" and "is the stored data correct" are different questions, and only the first one had a check.

## The check

`verifyStoredArtPaths.ts` asks the second one: it sweeps every distinct stored art path across the seven models that hold one (`imagePath`/`cardPath`/`heroPath`/`iconPath`, plus `Bot.avatarImage`) and asserts **the response is an image** — not that the request succeeded.

### The failure/backlog split

This is the design decision that keeps it usable:

| Response | Verdict | Fails the run? |
|---|---|---|
| 2xx, `content-type: image/*` | healthy | no |
| **2xx, non-image body** | **the silent path bug** | **yes** |
| 404 / 410 | art backlog — expected, self-healing | no, reported only |
| other | unreachable | no, reported only |

Conflating the last two with the second would make this cry wolf on every gap in the art queue and get it switched off.

## Verified against production, not just fixtures

The task's bar was *"reintroduce a `/rewards/...` path and confirm the check goes red."* Run live against `kind-robots.vercel.app`:

```
shell    200  text/html;charset=utf-8   /rewards/item/alien-communicator.webp
image    200  image/webp    174358 B    /images/rewards/item/alien-communicator.webp
image    200  image/webp    212082 B    /images/rewards/skill/adhd-spark.webp
missing  404  text/html                 /images/rewards/item/<no-such-file>.webp
image    200  image/webp     69272 B    /images/botcafe.webp
```

The legacy prefix still reproduces the defect live, so the check demonstrably goes red on it.

The pure classifier is additionally mutation-verified three ways — each of these fails the self-test:

```
treat any 2xx as healthy          → caught (the exact bug that hid 214 rewards)
conflate 404 with a path bug      → caught
stop skipping /api/art routes     → caught
```

## Scheduling

Nightly (`.github/workflows/stored-art-paths.yml`), not per-PR: it checks **data**, which changes when rows are written, not when code merges. A PR cannot introduce a bad stored path.

It joins the tailnet before touching the DB — without that step the host resolves to nothing and Prisma reports a 10s pool timeout (`active=0 idle=0`) that reads like a dead database rather than an unrouted one. That trap has now bitten two dispatch-only workflows, so it is called out in the file header.

## ⚠️ Unrelated finding worth your eyes

`https://kindrobots.org/` currently returns **404 `x-vercel-error: DEPLOYMENT_NOT_FOUND`** — the apex domain has no deployment attached. `https://kind-robots.vercel.app` is serving fine, which is why CI and this PR are green. Not touched here since it's a domain-alias config issue, not code, but the public domain appears to be down.

## Verification

- `test:stored-art-paths:self` — passes, mutation-verified above
- `npx eslint` on the new script — clean
- workflow YAML parses

Not run end-to-end against the database from this session: the sandbox cannot reach the ProxySQL port, which is why the sweep lives in a workflow. The first scheduled run (or a `workflow_dispatch`) will produce the real numbers.

---

### DIFF
--- a/.github/workflows/stored-art-paths.yml
+++ b/.github/workflows/stored-art-paths.yml
@@ -1,0 +1,98 @@
+# Stored art paths must serve real images.
+#
+# interface-vision/t-100. Every one of the 227 Reward.imagePath values was
+# stored as `/rewards/...` when the files live at `/images/rewards/...` (#1446).
+# 214 of them served a 142 KB Nuxt app shell as text/html — at HTTP **200**,
+# because a missing asset under a non-/images path falls through to the SPA
+# catch-all instead of 404ing. curl, uptime monitors and link checkers all read
+# that as healthy. The layout audit could not see it either: kr-art-plate caught
+# the decode failure and swapped in its fallback, exactly as designed, so the
+# page looked fine while the art was wrong.
+#
+# This job asks the one question none of those ask — does the stored path
+# actually return image bytes?
+#
+# Nightly rather than per-PR: it is a check on DATA, which changes when rows are
+# written, not when code merges. A PR cannot introduce a bad stored path, and a
+# few hundred requests per run is not something to spend on every push.
+#
+# FAILS ON: a 2xx whose content-type is not an image (the silent path bug).
+# REPORTS ONLY: a genuine 404 — that is art backlog, expected and self-healing.
+#
+# Requires the same repository Actions secrets as fallback-snapshot.yml. The DB
+# host is a tailnet address, so without the Tailscale step this fails with a
+# misleading 10s Prisma pool timeout (`active=0 idle=0`) that reads like a dead
+# database rather than an unrouted one.
+name: Stored art paths
+
+on:
+  schedule:
+    - cron: '43 10 * * *' # daily, 10:43 UTC ≈ 2:43am Pacific
+  workflow_dispatch:
+    inputs:
+      base_url:
+        description: 'Base URL to check against (defaults to production)'
+        required: false
+        default: ''
+
+concurrency:
+  group: stored-art-paths
+  cancel-in-progress: false
+
+jobs:
+  sweep:
+    runs-on: ubuntu-latest
+    env:
+      DATABASE_URL: ${{ secrets.DATABASE_URL }}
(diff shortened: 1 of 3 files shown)


silasfelinus:
## First real run — green, and it establishes the baseline

Dispatched against production after merge ([run 30964842779](https://github.com/silasfelinus/kind_robots/actions/runs/30964842779)). The whole sweep took **23 seconds**, so the nightly cadence has plenty of headroom.

```
📜 Checking 955 distinct stored art path(s) against https://kind-robots.vercel.app

   ✅ 929 serving an image
   ❌ 0 serving a non-image at 2xx
   🕳  26 absent from the media host (art backlog)
   ⚠️  0 could not be checked

✅ Every stored art path serves a real image or is honestly absent.
```

Three things this confirms that were previously assumed:

1. **The reward repair held across the whole corpus.** Zero paths anywhere still return the app shell — not just the 227 rewards that were measured in #1446, but all 955 across Reward, Bot, Character, Dream, Scenario, Facet and ArtCollection.
2. **The tailnet plumbing works.** The DB step connected and queried without the misleading 10s Prisma pool timeout, so this workflow won't need the debugging round that the last two dispatch-only DB workflows did.
3. **The backlog split behaves as designed.** 26 paths are genuinely absent and the run stayed green, reporting them as art backlog rather than failing. Mostly ArtCollection cards (`alexa-integration`, `conductor`, `sketchy`, `storymaker`, …) plus 19 rewards.

The 26 are art-queue items, not path bugs — no action needed per the standing "missing art is self-healing" direction. Listing them here so the auto-generator has a concrete worklist if it's useful.

---

PR #1475, opened 2026-08-05
Park 81 orphaned components in components/abandonware/

silasfelinus:
> *"if a component is truly orphaned, we should move it to abandonware, which includes components that just refer to each other. if it isn't referenced by something in content as an md or app.vue, it is likely orphaned."*

This PR started as "delete six dead components." It is now the general answer: a reachability check, a parking scheme, and **81 components moved out of the app build while staying whole in WonderLab.**

## Why a grep couldn't answer this

`model-gallery` referenced `model-card`, which referenced `lora-card`. All three looked referenced from every angle except the one that matters — nothing reachable referenced `model-gallery`. The same shape turned up in `components/butterfly/` (23 components citing each other) and `components/builder/` (16).

So `utils/scripts/verifyComponentReachability.ts` **traverses** from the real entry points — `app.vue`, `error.vue`, `pages/`, `layouts/`, and the MDC mounts in `content/*.md` — and reports what it cannot reach. The baseline now sits at **zero**: a PR that orphans a component fails until it parks it.

## Retirement stops being deletion

Deleting kept the reviews but cost the museum the exhibit: `wonderlab-preview-host` renders *"Component source not found"*, and no new AI review draft can ever be generated without source evidence. Parking keeps it whole.

| | Parked? |
|---|---|
| Nuxt auto-import | **gone** — `nuxt.config` already ignored `abandonware/**/*.vue` |
| `vue-tsc` typecheck | **gone** — `tsconfig` exclude repointed (it named a path deleted in `1e01180`) |
| WonderLab preview | **kept** — the museum glob still matches; `ignore` governs auto-import, not globs |
| Manifest entry + reviews | **kept**, flagged `abandoned: true`, `isDiscovered` still true |
| SFC compilation | **kept** — lazy chunks, not main-bundle weight. Zeroing it means losing the preview. |

## Three false-positive classes, each caught by something failing loudly

The first orphan list was **132**. It was wrong three times before it was right, and every correction came from a check failing rather than from review:

| Missed edge | Cost | Caught by |
|---|---|---|
| Directory-local `import.meta.glob` registries — `screenfx` globs `./*.vue` and mounts by name from `animationCatalog` | **47** live effects | reading the registry |
| Explicit `.vue` imports, value and type-only — `stage-manager` imports its three cards | **3** live components | `vue-tsc` |
| Nuxt's `Lazy` prefix — `<LazyWorkspaceNarrator>`, `<LazyKrNarratorStage>` | **6** live components | `test:narrative-kit` |

Only single-directory globs confer reachability. A `**` glob is a catalog, not a mount point — the museum's own `@/components/**/*.vue` would otherwise make the whole tree self-reachable and the check could only ever report zero. Both `Lazy` keys are kept, so a component genuinely named `LazyLoader` isn't read as a reference to `loader.vue`.

That trail is why the script **reports rather than judges**: static traversal cannot see `<component :is>`, a string-addressed registry, or a plugin CSS selector. I swept all 81 for string-name mounts before moving them — every hit was the manifest or a preview fixture.

## Parking breaks relative imports, and nothing else catches it

A file one directory deeper means `./../../stores/x` no longer resolves. `tsconfig` excludes the parked tree so vue-tsc is blind by design, but the museum glob still compiles these files — so a stale specifier is a **hard `nuxt build` failure**. `builder-art-input` did exactly this and killed the first build.

Fixed by re-anchoring six specifiers on `@/`, and guarded by a companion check in the same script. Its self-test injects an `exists` that says no to everything, and a live mutation reproduced the original failure by hand before the fix was confirmed.

## Verification

- No parked name appears in `.nuxt/components.d.ts`
- `npm run build` — client (30s) and server (21s) bundles both build. Prerender fails on a sandbox-missing `JWT_SECRET`; every `abandonware` line in the log is a pre-existing Rollup chunk hint that also names dozens of live components.
- Manifest: 408 entries, 81 flagged, **zero** flipped to `isDiscovered: false`
- `test:narrative-kit`, `test:wonderlab-*` (9), `test:layout-contract`, `test:gallery-adoption`, `test:gallery-consistency`, `test:nav-manifest`, `test:animation-catalog`, `test:animation-component-attempts`, `test:lint-ratchet` — all pass

## Known rough edge, documented

A parked component referencing **another** parked component loses auto-import for that child, so it renders as an unresolved element in the museum preview. Common in the `butterfly/` and `builder/` clusters, which reference each other heavily — the parent's own markup renders, the nested child comes up empty. Fixable per-component with an explicit `import` if a particular exhibit is ever worth the fidelity.

---


silasfelinus:
REVIEWING: 2026-08-05T081350Z-review-1475-k9q2 at 2026-08-05T08:13:50Z

PR #1492, opened 2026-08-05
Stop fragmenting the art-image CDN cache, and stop crushing phone tabs to 31px

silasfelinus:
Two independent fixes, both on paths that were failing quietly.

## Art image caching

`/api/art/images/[id]/file` was **1591 requests in a two-hour window** on 2026-08-05 — 34% of all Vercel function invocations and **3× the next busiest route**, because gallery pages mount 140+ images at once. Every CDN miss invokes the function, hits Prisma, and opens a ProxySQL session to read a base64 `LongText` blob. This route's cacheability is a direct multiplier on database sessions during the connection-capacity incidents this repo keeps having.

It served WebP only when `Accept` advertised it — correct HTTP, and exactly why it had to send `Vary: Accept`. That keys the CDN cache on the header's raw value, and Chrome, Firefox and Safari each send a different `Accept` string that changes between versions. One image fragmented into many cache entries, each needing its own invocation and database read.

Always transcoding makes the response **a pure function of the image id**: one cache entry serves every client forever under the existing immutable `Cache-Control`. WebP has been universal since Safari 14 (2020), and both degradation paths are untouched — a sharp failure falls back to the original bytes, and a transcode that grew the file keeps the original.

**Deliberately not touching the Vercel env vars.** They still request `connectionLimit 10 / idle 300 / minimumIdle 1` and are saved only by the code's clamp (logged on every request). Changing production database config unsupervised overnight isn't the right call — it's written up separately.

## Phone tab strip

Four tabs rendered **31px wide** at 390px on the deployed app: Dashboard, Register, Navigation, Themes. At that size a tab is an icon and a truncated letter — it reads as a rendering fault rather than navigation, which is worse than hiding it.

`tabLayoutMetrics()` already knows a tab needs 96/112/128px, but only used that to decide **how many** tabs fit. The width each one gets is a percentage from `tabButtonBasis`, and those agree only if the strip's `clientWidth` was measured correctly when the count was computed. Measured before the strip is constrained — first paint, in-flight layout, unsettled resize — capacity comes out too high, every tab is declared to fit, and each gets an equal fraction of a much narrower strip.

Enforcing the floor in CSS makes the layout correct regardless of whether the measurement was right, and it applies before any JS runs. Overflow goes to the horizontal scroller this strip already is — snap points, hidden scrollbar and both arrows were always the intended overflow behaviour.

**Verified in a real browser**, not by reasoning about CSS:

| | widths | min | strip scrolls |
|---|---|---|---|
| before | 83, 68, 81, 66, 35… | **35px** | yes |
| after | 96, 96, 96, 96, 96… | **96px** | yes |

## The check

`verifyArtImageCaching.ts` guards the caching invariant, since a regression there looks like nothing at all until it surfaces as a connection incident. Four assertions, each mutation-tested: reinstating `Vary`, reinstating the `Accept` read, dropping `immutable`, and dropping either transcode fallback. It strips comments before reading, so prose explaining why `Vary` is wrong can't satisfy the rule forbidding it.

---

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -72,6 +72,12 @@ jobs:
       - name: Page backdrop contract
         run: npm run test:page-backdrop
 
+      # The app's busiest route by 3x. Its CDN cacheability is a multiplier on
+      # database sessions, so a silent regression here shows up as a
+      # connection-capacity incident rather than as anything looking broken.
+      - name: Art image caching contract
+        run: npm run test:art-image-caching
+
       # t-099: ESLint as a ratcheted gate — never asks anyone to fix the
       # pre-existing problems, only refuses NEW ones, per rule.
       #
--- a/components/navigation/workspace-header.vue
+++ b/components/navigation/workspace-header.vue
@@ -685,6 +685,48 @@ function goBack(): void {
   white-space: nowrap;
   scroll-snap-align: start;
   scroll-snap-stop: always;
+
+  /*
+   * ENFORCE the minimum width, do not merely calculate with it.
+   *
+   * tabLayoutMetrics() already knows a tab needs 96/112/128px to be readable,
+   * but it only uses that to decide HOW MANY tabs fit; the width each tab
+   * actually gets is a percentage from tabButtonBasis. Those two agree only if
+   * the strip's clientWidth was measured correctly at the moment the count was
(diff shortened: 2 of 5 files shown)


silasfelinus:
REVIEWING: 2026-08-05T111451Z-review-1492-x4p7 at 2026-08-05T11:15:30Z

PR #1500, opened 2026-08-05
Get backdrop art out of the database: export → link → prune

silasfelinus:
Two halves of one pipeline. The first is **required for your queued art to appear at all**; the second is the `imageData` cleanup you approved.

## 1. Export — the step that was missing

`enqueuePageBackdropArt` puts `imagePath: background/<page>-<variant>.webp` in every payload, and the frontmatter declares the same path. **Nothing acted on it.** `complete.post.ts` stores finished art as an ArtImage row (`imageData` set, `imagePath` null), and per `docs/self-hosted-media.md`:

> *"Vercel cannot write to the Unraid share. Production upload behavior remains database-backed."*

So all 60 jobs would succeed, 60 ArtImage rows would appear, and **every page would still show nothing.** The `imagePath` I put in those payloads was descriptive metadata no code consumed — my oversight when writing the seeder.

```
npm run export:page-backdrops             # dry run
npm run export:page-backdrops -- --write  # write files + set imagePath
```

Run on the Windows/WSL host where `IMAGES_PATH` reaches the share. Without it, the dry run still reports, falling back to `public/images` and saying so.

**Transcodes to WebP** rather than writing a PNG under a `.webp` name — nginx serves content-type by extension, so that mismatch would be served as `image/webp` and break strict decoders. Verified: 19KB PNG → 2KB WebP on a 1536×864 canvas, decoding back as `webp` at the right dimensions.

**Setting `imagePath` is the point, not a side effect.** Once set, `file.get.ts` redirects to the file instead of reading base64 — so these rows stop costing a `LongText` read per view, on the route that was already a third of all function invocations.

## 2. Prune — the field deletion you approved

> *"If an image has a local imagePath, we don't need the base64 data in the database and I approve field deleted."*

```
npm run prune:art-image-data                 # dry run
npm run prune:art-image-data -- --limit 50   # sample first
npm run prune:art-image-data -- --write      # act
```

### Two ways this could destroy art, both guarded

**Self-referential paths.** `dailyDreamArchiveStore` writes `imagePath = /api/art/images/<id>/file?v=<ts>`. The redirect guard skips those (it would loop), so they *do* fall through to `imageData`. The SQL exclusion mirrors the endpoint's substring test, verified equivalent including the `?v=` variant.

**A path that isn't actually a file** — live on the deployment right now:

```
skip-not-image  200 but text/html   /rewards/definitely-not-a-real-file-xyz.webp
prune           image/webp          /images/background/artgallery.webp
```

A missing path returns **HTTP 200 with an app shell**, not a 404. That's the #1446 signature, where 214 of 227 Reward paths served 142KB of HTML while every status check called them healthy. **A pruner keyed on status codes would have deleted the only copy of all 214.** So rows are pruned only on 2xx **and** `content-type: image/*`. `GET`, not `HEAD` — the catch-all answers `HEAD` just as convincingly.

Writes are chunked at 200 rows so one huge `UPDATE` can't hold a long transaction against the shared ProxySQL pool.

## Order matters

**Export → link → prune.** Each step is safe alone, each is reversible until the last, and no step destroys the only copy. Neither script has been run — both are dry-run by default and the database is unreachable from here.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "migrate:art-jobs": "tsx utils/scripts/migratePromptsToArtJobs.ts",
     "seed:twisted-fairy-tales": "tsx utils/scripts/enqueueTwistedFairyTalesArtPrompts.ts",
     "seed:page-backdrops": "tsx utils/scripts/enqueuePageBackdropArt.ts",
+    "prune:art-image-data": "tsx utils/scripts/pruneRedundantArtImageData.ts",
     "characters:repair": "tsx utils/scripts/repairCharacterSlugs.mjs",
     "seed:davinci": "tsx utils/scripts/seedDaVinciEndings.ts",
     "seed:davinci:verify": "tsx utils/scripts/verifyDaVinciSeed.ts",
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "migrate:art-jobs": "tsx utils/scripts/migratePromptsToArtJobs.ts",
     "seed:twisted-fairy-tales": "tsx utils/scripts/enqueueTwistedFairyTalesArtPrompts.ts",
     "seed:page-backdrops": "tsx utils/scripts/enqueuePageBackdropArt.ts",
+    "export:page-backdrops": "tsx utils/scripts/exportPageBackdropArt.ts",
     "prune:art-image-data": "tsx utils/scripts/pruneRedundantArtImageData.ts",
     "characters:repair": "tsx utils/scripts/repairCharacterSlugs.mjs",
     "seed:davinci": "tsx utils/scripts/seedDaVinciEndings.ts",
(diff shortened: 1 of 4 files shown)


silasfelinus:
REVIEWING: 2026-08-05T201208Z-review-1500-a9f3

silasfelinus:
Review blocker on head `ae7fa62800d7259cf84d3cebb10b31ddd32842bb`:

The verifier currently proves only that `imagePath` resolves to some `image/*` response, not that the response is independent of this row's `imageData` or that it is the intended image. Since redirects are followed, a nominally non-self-referential path can redirect to `/api/art/images/<same id>/file`; verification then succeeds by reading the base64 that `--write` immediately deletes. A wrong-but-valid image path also passes content-type verification and can discard the only correct copy.

Before merge, make the proof row-specific and race-safe. At minimum:
1. reject a final `res.url` that resolves to the same ArtImage file endpoint;
2. retain the exact verified `imagePath` and include it unchanged in each update predicate so a concurrent path change cannot delete against stale evidence;
3. preferably compare the fetched bytes with the stored base64 using a content hash, rather than treating `image/*` as identity proof.

Dry-run should remain the default, and the production write should not be executed as part of this PR. All current CI is green, but CI does not cover this destructive-data invariant, so I am leaving the PR open and unmerged.

silasfelinus:
REVIEWING: 2026-08-05T211333Z-review-1500-r7k4

PR #1504, opened 2026-08-05
Restore the backdrop and prune work that #1500's squash dropped

silasfelinus:
> *"there was no update when i git pulled"*

Because `main` genuinely didn't have the fixes. **#1500 merged an older snapshot** of the branch — the resolver route landed, everything after it didn't. Confirmed by reading the files back off `origin/main`, not by trusting the merge.

## Verified missing from main, restored here

| | |
|---|---|
| `SELECT COUNT(*) AS rows` → `rowCount` | `ROWS` is reserved in MariaDB 10.6+ — **this was the syntax error killing every run** |
| totals reported first | the size prints before any slow work |
| candidates selected without blobs | was pulling every base64 row with no `LIMIT` — the whole corpus before printing a line, indistinguishable from a hang |
| `engine \|\| 'A1111'` → `COMFY` | the silent default that killed 60 jobs |
| `ENGINE` constant in the seeder | refresh sets the **column**, not just the payload |
| 20 content pages | `/api/art/backdrop/<slug>` instead of a share path, so art resolves by itself |

Content changes were diffed to confirm they touch only `background*` keys before being taken, so no other agent's edits are reverted.

## Verification

```
prune: AS rows,      -> 0   ✓
prune: totals-first  -> 1   ✓
queue: A1111 default -> 0   ✓
seed:  ENGINE const  -> 1   ✓
content: slug route  -> 20  ✓
```

`page-backdrop`, `art-image-caching`, `channel-content`, `layout-contract` all pass.

## Two process failures worth naming

**Rewriting a branch someone is pulling from.** My rebase broke Silas's fetch — and a ref-lock failure aborts the *entire* fetch, so `origin/main` silently stopped updating too. That's why `git pull` said "already up to date" while being three merges behind. I should not have rewritten a branch he was actively pulling.

**"Pushed" is not "landed."** I told him to pull a fix that existed only on a PR branch, then merged a PR that didn't contain it, then said it was done — twice. The check that would have caught both is reading the file back off `origin/main` afterwards, which is what finally found this.

---

### DIFF
--- a/content/academy.md
+++ b/content/academy.md
@@ -15,9 +15,9 @@ dashboardKey: academy
 dashboardTab: timeline
 loadingMessage: Restoring the frescoes...
 refreshLabel: Refresh Academy
-backgroundMobile: background/academy-mobile.webp
-backgroundTablet: background/academy-tablet.webp
-backgroundDesktop: background/academy-desktop.webp
+backgroundMobile: /api/art/backdrop/academy-mobile
+backgroundTablet: /api/art/backdrop/academy-tablet
+backgroundDesktop: /api/art/backdrop/academy-desktop
 ---
 
 :academy-manager
--- a/content/art.md
+++ b/content/art.md
@@ -16,9 +16,9 @@ dashboardTab: gallery
 cards: artCards
 loadingMessage: Loading art gallery...
 refreshLabel: Refresh Art
-backgroundMobile: background/art-mobile.webp
-backgroundTablet: background/art-tablet.webp
-backgroundDesktop: background/art-desktop.webp
+backgroundMobile: /api/art/backdrop/art-mobile
+backgroundTablet: /api/art/backdrop/art-tablet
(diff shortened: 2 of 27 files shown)


silasfelinus:
REVIEWING: 2026-08-05T221942Z-review-1504-q7n3 at 2026-08-05T22:19:42Z

silasfelinus:
CI diagnosis: Contract Tests run 31052090831 failed only at capture-group guard step 106. `utils/scripts/verifyPageBackdrop.ts` calls `check(Boolean(route), ...)`, but that helper neither narrows nor terminates, then indexes `route[1]` and `route[2]`. Add an explicit `if (!route) continue` after the diagnostic check, before `declaredSlugs.add(...)`. The focused backdrop contract, TypeScript, layout, caching, and all other completed suites passed. Leaving this PR open until a new head completes green.

PR #1509, opened 2026-08-06
Report where the heavy ArtImage rows came from

silasfelinus:
Answers "if art is generated on the box and moved by kr_relay, why is it still in the database?"

Half is answerable from the code, and the answer is neither old rows nor unconnected duplicates: **`/api/art/save-generated` — the relay's upload hop — contains no reference to `imagePath` at all**, and `complete.post.ts` only ever sets `imagePath: null` (on archive snapshots). The relay could never have linked a local file, so every relayed image is base64-in-database by construction, including ones made today. Moving files with kr_relay never touched the database copy because nothing in kind_robots was listening for it.

What the code can't say is whether that's history or ongoing. This adds three groupings to `npm run inspect:art-image-data`:

- **heavy rows by month created** — a tail that stops a year ago is a migration leftover; one reaching this month means the pipeline is still filling the column
- **by generating server** — `serverName` is stamped from the resolved Server record, so it says which box actually rendered them
- **whether `fileName`/`path` hold anything file-like** — i.e. whether a twin could be reconnected on disk instead of re-exported. A bare `ArtImageUpload-<timestamp>` is the placeholder `saveImage()` invents, not a file on the share.

Read-only; `SELECT` only.

---

### DIFF
--- a/utils/scripts/inspectArtImageData.ts
+++ b/utils/scripts/inspectArtImageData.ts
@@ -85,6 +85,99 @@ async function main() {
   for (const b of buckets) {
     console.log(`   ${String(b.bucket).padEnd(16)} ${String(Number(b.n)).padStart(6)} rows   ${Number(b.mb)} MB`)
   }
+
+  /*
+   * WHERE DID THESE COME FROM? Silas, 2026-08-06: "I am a bit confused about
+   * how we seem to still have the art data in the db when we have been
+   * generating art on our box, then moving it with our kr_relay. So where are
+   * these data sets coming from? Were they old ones that never got saved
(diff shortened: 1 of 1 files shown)


silasfelinus:
REVIEWING: 2026-08-06T001129Z-review-1509-q8v3 at 2026-08-06T00:11:29Z

silasfelinus:
Review complete: the diff is read-only diagnostic SQL with no interpolated input, all required workflows completed successfully on exact head `2ff1529a6d847b1ecb766c6e8f8802420b52deec`, and there are no unresolved review threads. GitHub does not permit approving one's own PR, so I am recording the completed review here before merge.

PR #1515, opened 2026-08-06
Shorten what has to be typed on a maintenance box

silasfelinus:
The Unraid web terminal is an xterm.js canvas, which mobile browsers refuse to paste into — so every command and the entire `.env` block has to be typed by hand when working from a phone.

**`scripts/write-env.sh`** prompts for host, database, user, password and `IMAGES_PATH`, then writes `.env` under `umask 077`. The password uses `read -rsp`, so it never lands in shell history. It also prints the reminder that the Prisma CLI does *not* read `.env` and needs `set -a; . ./.env; set +a` first — which is the exact thing that made `migrate resolve` fail on the first attempt.

**`npm run offload:sample`** is `--write --limit 20 --only-tagged` under a name that says what it does. Deliberately not a short alias for the unbounded run: it's capped at 20 rows and named "sample" so it can't be mistaken for a dry run by someone typing quickly on a phone.

```bash
bash scripts/write-env.sh
npm run offload:sample
```

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,7 +156,8 @@
     "generatesmart": "node utils/scripts/generate-smart-icons.mjs",
     "test:api-client-followups": "tsx utils/scripts/verifyApiClientFollowups.ts",
     "offload:art-images": "tsx utils/scripts/offloadArtImageData.ts",
-    "test:art-image-offload": "tsx utils/scripts/verifyArtImageOffload.ts"
+    "test:art-image-offload": "tsx utils/scripts/verifyArtImageOffload.ts",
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: 2026-08-06T031253Z-pr1515-k9r4 at 2026-08-06T03:12:53Z

silasfelinus:
Review fix pushed on the existing branch: URL-encode the database name, user, and password before writing `DATABASE_URL`, with the password passed to Node over stdin rather than process arguments. This keeps reserved characters from corrupting the connection string without exposing the password in shell history or `ps` output. Waiting for the replacement checks on `f32a779af84b474430cd0067864f470ef8eb8805`.

PR #1519, opened 2026-08-06
Let page backdrops actually show: put page roots on the surface tokens

silasfelinus:
Silas, looking at Taskmaster with its backdrop art finally generated: *"instead of being the actual background, it just behind main content, so I suspect some routes are hiding it more or less completely."*

Exactly right. Everything upstream works — I verified all 20 declared pages resolve through the full chain to real WebP bytes on the media host, not just a 302. The art was rendering. It was just **behind an opaque page**, which is why the only visible art was the strip below where the page stops.

## The cause was a scoping decision in the original plan

Surface tokens were wired into the shared kit — `kr-gallery`, `kr-chat-window`, `reactable-card`, `kr-entity-card-body`, `kr-choice-list` — and page roots were **explicitly left alone**. That reads as conservative, but a page root is the largest surface on screen: leaving it opaque is the one exclusion that undoes the entire feature.

`components/pages/taskmaster-page.vue:7` was `class="kr-surface … bg-base-100"`.

## The change

Fourteen page roots now read `bg-(--kr-surface…)`:

| from | to |
|---|---|
| `bg-base-100` | `bg-(--kr-surface-raised)` |
| `bg-base-200` | `bg-(--kr-surface)` |
| `bg-base-300` | `bg-(--kr-surface-sunken)` |

Because the tokens resolve to the **identical** theme colour when no backdrop is declared, every page without art renders exactly as before. Only pages that declare art become translucent, at the single opacity knob under `[data-kr-backdrop]` in `tailwind.css` — so tuning how much art reads through stays one edit in one place.

Nested chips, pills and wells keep their solid colours on purpose: they're elements sitting *on* a surface, not the surface itself, and the mockups show them reading as solid.

## Verification

`verifyPageBackdrop` now pins all fourteen roots, with the reasoning inline so the next person doesn't re-apply the same exclusion. Watched to fail by reverting `taskmaster-page.vue` to `bg-base-100`.

Lint ratchet unchanged.

---

### DIFF
--- a/components/pages/about-page.vue
+++ b/components/pages/about-page.vue
@@ -1,6 +1,6 @@
 <!-- /components/content/pages/about-page.vue -->
 <template>
-  <div class="kr-unbound items-center gap-4 bg-base-200 py-6 px-4">
+  <div class="kr-unbound items-center gap-4 bg-(--kr-surface) py-6 px-4">
     <div class="w-full max-w-2xl">
       <!-- Section header -->
       <div class="mb-6 flex flex-col items-center gap-2 text-center">
--- a/components/pages/conductor-page.vue
+++ b/components/pages/conductor-page.vue
@@ -5,7 +5,7 @@
 
     <!-- COCKPIT BAR: unified slim context strip -->
     <div
-      class="kr-toolbar rounded-xl border border-base-300/70 bg-base-100/90 px-3 py-1.5"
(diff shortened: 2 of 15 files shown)


silasfelinus:
REVIEWING: 2026-08-06T081149Z-review-1519-p7x4 at 2026-08-06T08:11:49Z

PR #1520, opened 2026-08-06
Collapse duplicate landing-zone art to one file each

silasfelinus:
Measured on the real share after the backfill:

```
106 duplicate groups, 894 redundant files
342M    /mnt/user/pc/kindrobots/images/generated
```

Silas identified the cause as a closed bug — *"bad generation prompts mixed with a queue rebuilder that wasn't deleting when reenqueing, I think eventually multiplying these generations… also because we weren't randomizing seeds."* All three are fixed, so this is cleanup rather than a recurring chore.

```bash
npm run dedupe:generated-art               # dry run
npm run dedupe:generated-art -- --write
```

### How identity is established

The short content hash `offloadArtImageBytes` already wrote into every landing-zone filename (`artimage-<id>-<hash>.webp`), taken from the bytes actually written. Two files sharing it are genuinely identical, and nothing has to re-read 342 MB to prove it. Lowest id survives.

### Jobs are deleted, not repointed

Silas: *"I have no need for umpteen art jobs pointing to the same image."* `ArtJob.artImageId` is a plain `Int` with no foreign key, so those rows would otherwise survive as dangling pointers. The surviving image keeps its own job.

### Ordering is the safety property

The `ArtImage` row is deleted **first**. It's the only step that can be *refused* — a foreign key would mean an entity claimed the image after all — and everything after it is irreversible. Deleting jobs first would mean a refusal had already destroyed the job history of an image we then keep.

As written, a refusal leaves the jobs, the row and the file exactly as they were, and reports the reason in the summary. (This is the opposite of the offload's ordering, deliberately: there the file was the *new* copy, here it's the last one.)

Scope is only rows under `/images/generated/`. Filed entity art is never considered, even if it happens to duplicate something.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,7 +157,8 @@
     "test:api-client-followups": "tsx utils/scripts/verifyApiClientFollowups.ts",
     "offload:art-images": "tsx utils/scripts/offloadArtImageData.ts",
     "test:art-image-offload": "tsx utils/scripts/verifyArtImageOffload.ts",
-    "offload:sample": "tsx utils/scripts/offloadArtImageData.ts --write --limit 20 --only-tagged"
+    "offload:sample": "tsx utils/scripts/offloadArtImageData.ts --write --limit 20 --only-tagged",
+    "dedupe:generated-art": "tsx utils/scripts/dedupeGeneratedArtImages.ts"
   },
   "devDependencies": {
     "@nuxt/eslint": "^1.1.0",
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: 2026-08-06T091429Z-review-1520-k4p7 at 2026-08-06T09:14:29Z

PR #1533, opened 2026-08-06
One manager layout instead of six hand-copied ones, and the second Facet browser retired

silasfelinus:
## The complaint this answers

> "We're supposed to be making consistent bespoke components so every piece, especially our main models, follows the same high quality and consistent layout … If we are just tweaking individual managers and looking for extra galleries, then we're going to keep reinforcing the same problems: that the site is a hodgepodge of individually created llm sites that don't agree on layout." — Silas, 2026-08-06

That changed what this PR does. It started as "convert 14 galleries one at a time"; that would have left the generator running.

## What was actually there

The core managers were **one component, hand-copied and then separately edited**:

- `bot-manager` and `character-manager` carried **character-for-character identical** tab-resolution blocks — same 35 lines, same names, same order.
- `reward-manager` reimplemented that logic a third way (its own tab union, its own array, reading `navStore.dashboardShell.dashboardKey` instead of `isDashboardTabKey`). `facet-manager` had a fourth variant.
- All of them opened with the same root shell string.
- And then **four different loading/error banners**. `scenario-manager` said *"Loading weirdness from the database…"* with a Try Again button; `dream-manager` used an icon-only tooltip refresh at `text-xs`; bot, reward and character shared a third at `text-sm shadow`.

Every model already agreed on the *script* contract — `isLoadingManager`, `managerError`, `refreshManagerData`, `activeTab`. Only the **layout** had been re-invented.

## `kr-manager`

Owns the root shell, the one status banner, tab resolution, the two pane shapes (flush vs padded panel), and unknown-tab recovery. The model keeps its store, its fetch and its error text — the shell takes `loading`/`error` and emits `refresh`, exactly as `kr-gallery` takes `items` and emits `update:mode`.

**All six tab managers are on it.** Three things the conversion surfaced:

- **Panel headers now come from `dashboardHelper`**, which already carried `title`/`summary`/`icon` per tab. Every manager hand-wrote that copy on top of the config, *differently*: reward's Add tab said "Add Reward / Create a skill, item, pet, boon, curse…" while its config said "Reward Generator / Create fresh skills, items, powers, pets, and more."
- **Unknown tabs fall back to the default tab's slot** rather than an error pane. `scenario-manager` alone had that via a bare `v-else`, and it's right for all of them — scenario declares five tabs, three of which live on other routes.
- **dream's legacy tab keys** (`add`, `maker` → `dreammaker`) are declared as `aliases` on the shell.

`conductor-manager` is deliberately **not** converted: it routes on `pageStore.workspaceCardKey`, not dashboard tabs, has no banner and no tab machinery, and is 54 lines. Forcing it would be pretend-consistency.

## The second Facet browser, retired

`/facets` rendered **two** Facet browsers. The Gallery tab went through `kr-gallery`. The **Library tab** drew its own `<article v-for="facet in filteredFacets">` card grid over all ~1,611 rows, whose only click target expanded an editor *inside the grid cell* — and the file contained no `router.push`, no `NuxtLink`, no `openFacet`. Nothing linked a Library card to the canonical Facet profile.

`facet-interact.vue`'s own header comment already described that in the past tense. It was still doing it: the interact tier had been added *beside* the old browser, not in place of it. Silas caught it by eye after two automated passes reported the gallery work finished — because every check keyed on components **named** `*-gallery`, and this one is called `facet-manager`.

- **`facet-editor.vue`** (new) — the per-Facet form, save, archive/restore and the four art slots. Fills `facet-interact`'s `#detail` slot, which was already in the file waiting for it. A Facet under edit now has a URL (`?facet=<slug>`) that links, bookmarks and back-buttons.
- **`facet-manager.vue`** — on `kr-manager`. Library keeps creation, the catalog fetch, and a per-taxonomy count: what it was genuinely for, previously buried under 1,611 cards.

Six contracts asserted against the old layout and now follow the code.

## The rules underneath

| | | |
|---|---|---|
| **Rule 1** | ratcheted | Every `*-gallery` reachable from a content route must mount `kr-gallery`. **14** recorded. |
| **Rule 2** | **hard** | Each core object has ONE route, that route reaches ITS OWN gallery, on the shell. **7/7**. |
| **Rule 3** | **hard** | A second multi-column grid of the same object on its own route. Structural, never by filename. **0**. |
| **Rule 4** | **hard** | Tab managers on `kr-manager`. **6/6**. |

Rules 3 and 4 reached zero in this PR, so their baselines are empty and they gate outright. Rule 3 exists because a first draft keyed on `*-gallery.vue` filenames and would have passed the Library grid — reproducing the bug it was written to catch. **A filename is not a shape.**

## The detector almost lied, and that fix matters more than the refactor

Rule 3 first flagged the *replacement* taxonomy-count grid, because it followed any mention of the collection and a count of Facets mentions `facetStore.facets`. Switching to receiver-based resolution fixed that — **and silently broke detection of the original grid**, whose real definition is:

```js
return visibleFacets.value.filter((facet) => {
  ... facet.metadata ? JSON.stringify(facet.metadata) : '' ...
})
```

A regex ternary split matched the `?` **inside the callback** and returned `JSON.stringify` as the receiver. The rule reported a clean pass on the exact grid it exists for. `splitTernary` is now bracket-depth aware, and both cases are permanent self-tests.

Verified by restoring the original file (**flags**) and the new one (**clean**).

## Also corrected

- **Adoption was a substring match** — `class="kr-gallery h-full"` in `ui-gallery.vue` counted as an adopter. **7/20 was really 6/20.**
- **Reachability kept only the first route found**, printing `dream-gallery … reached from /academy` — true only via `academy-manager → image-upload → dream-gallery`.
- **`/facet-gallery` deleted.** The page-backdrop contract is bidirectional, so the route, its backdrop seed entry, and its mount-or-redirect assertion went together.

## Verification

Every assertion **watched to fail under mutation**:

- point `/facets` at a non-gallery mount → **Rule 2 fails**
- a second route rendering `icon-gallery` → **Rule 1 reports growth**; `--update` refuses it
- restore the original Library grid → **Rules 3 and 4 fail, 2 errors**
- turn `bot-manager`'s `<kr-manager>` into `class="kr-manager"` → **Rule 4 fails**
- `<kr-gallery>` → `class="kr-gallery"` → **not counted as a mount**

All **104** steps of the contract-tests workflow were also run locally against this tree: zero failures. Plus `npm run test` (vue-tsc), eslint and prettier.

### CI notes

Workflows never auto-triggered on this branch — only Vercel and GitGuardian registered — so every Actions run here was dispatched manually via `workflow_dispatch`. On the head commit: **Contract verifiers ✅, facet-catalog ✅, layout-contract ✅, entity-art verify ✅**, with TypeScript still draining from the GitHub incident backlog.

One real failure was caught and fixed along the way: the WonderLab preview-coverage gate correctly flagged `kr-manager` and `facet-editor` as new components with required props and no fixture. Both now carry skip reasons (neither can be mounted from props alone — one is slot-driven, the other resolves its record from a store by id), matching the existing `flip-panel` and `workspace-narrator` precedents. Worth knowing: `npm run audit:wonderlab-previews` alone **never fails** — the gate lives behind `--strict`, which only the workflow passes.

**Not verified in a browser** — no local DATABASE_URL. Worth a preview spot-check: the six managers' banner and tab switching, the Add/Forge panels' new config-sourced headers, and on `/facets` that clicking a card opens `?facet=<slug>` with the editor in the detail pane and the back button returning to the gallery.

One deliberate visual change: the `/facets` Library tab loses its `max-w-7xl` centering, since it was the only manager pane that had one. If a capped reading width is wanted, it belongs on the shell so all six panels get it — not as a facets-only exception.

## Still open

Rule 1's 14 hand-rolled galleries, unchanged here — the manager shell was the higher-leverage fix, and several of the 14 (`art-gallery` at 1,653 lines, `chat-gallery`'s transcript pane, `achievement-gallery`'s leaderboard column) are only partly galleries.

Next, per Silas: the `*-interact` tier. Same extraction, different goal — there the models genuinely diverge, so the uniform thing is the *frame* (`<x-gallery>` until selected, then `<x-workspace>`), not the contents.


silasfelinus:
REVIEWING: claude-conductor-scheduled-20260806T2200Z-review-1533 at 2026-08-06T22:30:00Z

---

silasfelinus:
## Review

Read the full diff, the PR body's claims, and pulled the current head-branch content of `kr-manager.vue`, `facet-manager.vue`, `facet-interact.vue`, and `facetCatalogStore.ts` to cross-check against the description rather than trust it. CI was mid-catch-up from the outage while I reviewed — see status below.

### What looks solid

- **`kr-manager.vue`** does generalize the six managers correctly, as far as the diff shows:
  - Unknown-tab fallback (`renderedTab` → default tab → warning pane) faithfully reproduces `bot-manager`'s original recovery behavior and correctly extends `scenario-manager`'s old bare `v-else` (which collapsed *any* unrecognized tab, including its three other-route tabs, to `scenarios`) to all six.
  - `dream-manager`'s legacy `add`/`maker` → `dreammaker` aliasing is preserved via the new `aliases` prop, and the old `normalizeTab()` behavior for genuinely unknown values (fall through to default) is subsumed by the shell's own fallback — I checked this matches exactly.
  - Panel-header sourcing from `dashboardHelper` (`tabConfig.title`/`summary`/`icon`) is only applied to tabs in `panelTabs`, and each migrated manager declares the same set of "padded panel" tabs its original hand-rolled CSS wrapped (`bot`→`forge`, `reward`→`add`, `scenario`→`add`, `facets`→`library`, `character`→ none, matching each original's classes 1:1).
  - `dream-manager`'s `#persistent` slot for `daily-dream-generator` (rendered above every tab) is correctly special-cased and documented as the one deliberate hole in "the shell owns the frame."

- **`facet-editor.vue`**: no `v-html`/unescaped rendering anywhere (plain interpolation only), save/archive/restore all wrapped in try/catch with a surfaced `errorMessage`, and the `facetId: number` prop lines up correctly with `FacetCatalogEntry.id: number` from `facetCatalogStore.ts` (the slot-scope value passed in from `facet-interact.vue`'s `#detail` slot) and with `facetStore`'s own `id: number` — no type mismatch between the two stores' shapes.

- **`splitTernary`** does fix the described bug correctly: it's bracket/quote-depth-aware, explicitly skips `?.` and `??`, and its own self-test (`nested` case in the diff) reproduces the exact `facetStore.facets.filter((facet) => facet.metadata ? ... : ...)` pattern that silently broke detection before, asserting the receiver resolves to `visibleFacets.value.filter` rather than `JSON.stringify`. `matchingColon`'s nested-`?`-counter correctly closes nested ternaries before the outer one. One untested edge case: multi-level nested ternaries (`a ? b ? c : d : e`) return the outer two branches only, without recursing into the inner one for receiver purposes — not exercised by a self-test, but this is analysis tooling scoped to this codebase's actual patterns, not a general JS parser, so low risk.

- **`/facet-gallery` deletion** is clean and bidirectional: `content/facet-gallery.md`, its `pageBackdropArtPrompts.ts` seed entry, and `verifyFacetGallery.ts`'s requirement that the file exist were all removed together in this diff. I confirmed (via code search against `main`, since the PR branch isn't indexed) that the only remaining `/facet-gallery` mentions repo-wide are historical comments in `dashboardHelper.ts` and `pages/[...slug].vue` explaining *why* the redirect logic exists — not functional references — plus `package.json`'s `test:facet-gallery` script and `.github/workflows/facet-catalog-contract.yml`, both of which still pass because the script itself (touched in this diff) no longer requires the deleted file.

### Worth a second look (not blocking)

1. **`dashboardKey` silently went from computed to a hardcoded literal for `reward-manager` and `scenario-manager` only.** Pre-PR, those two (uniquely among the six converted) derived it via `computed(() => navStore.dashboardShell.dashboardKey || defaultDashboardKey)`; `bot`/`character`/`dream`/`facets` were already hardcoded strings. This PR hardcodes all six. That pattern is still alive verbatim in *other*, not-yet-converted managers (`server-manager.vue`, `art-manager.vue`, `academy-manager.vue`, `giftshop-manager.vue`), which suggests it's boilerplate rather than something these two specifically needed — but the PR body doesn't call this drop out anywhere, and I can't rule out a case where `reward-manager`/`scenario-manager` get mounted somewhere the active dashboard shell key diverges from the literal (similar in spirit to the `dream-gallery`-reached-from-`/academy` case the PR itself flags elsewhere). Worth a quick confirmation rather than assuming it's inert.
2. **`facet-editor.vue` has no loading/not-found state.** `v-if="facet"` on the root means a deep link to `?facet=<slug>` that arrives before `facetStore.facets` resolves, or a stale/invalid `facetId`, just renders nothing — no spinner, no "not found." Minor, and the PR body already flags the whole feature as "not verified in a browser."
3. **`kr-manager`'s true unknown-tab pane is self-defeating in its one firing case.** That pane only renders when even the *default* tab has no slot; its "Go to `{{ defaultTabLabel }}`" button calls `goToDefaultTab()`, which sets the tab to... the same default tab that just proved slot-less. Only reachable via a dashboardHelper config bug (a manager missing a slot for its own declared default), so very low real-world likelihood, but the recovery button would be inert exactly when needed.

### CI status

As of this review, mid-catch-up from today's outage: `GitGuardian`, `Vercel`, `facet-catalog`, and `Contract verifiers` have completed and passed; `TypeScript`, `layout-contract`, and `verify` are still queued/in progress. `mergeable_state` was `blocked`, consistent with checks not yet all resolved rather than an actual conflict.

### Recommendation

**Safe to merge once the remaining checks (`TypeScript`, `layout-contract`, `verify`) go green.** No correctness-blocking bugs found in `kr-manager.vue`, `facet-editor.vue`, or the `splitTernary` fix — the manager-generalization and Facet-browser-consolidation claims in the PR body check out against the actual diff. The three items above are worth a look but none of them are regressions serious enough to hold the merge on their own; #1 is the one I'd actually want a "yes, confirmed inert" from the author on, since it's the one place this refactor silently changed pre-existing logic (rather than porting it) without saying so.

---

PR #1536, opened 2026-08-06
Fix ArtJob queue preview fan-out and visual layout

silasfelinus:
## Root cause

The ArtJob queue was doing three expensive things at once:

- `performFetch()` passed relative `/api/...` URLs to native `fetch()` during SSR, which produced `Failed to parse URL from /api/icons`.
- Opening the queue also fetched the hidden trainer list and then eagerly requested full image data one ArtImage at a time, up to 240 requests.
- The 24-hour stats endpoint ran seven aggregate queries every 15 seconds while those image requests were competing for the database pool.

## What changed

- Resolve relative API URLs against the active Nuxt request origin during SSR.
- Stop eager full-image loading from queue and trainer fetches.
- Use the existing cacheable `/api/art/images/:id/file` media route for public/general previews, with native lazy image loading.
- Load protected/private previews only when explicitly requested.
- Fetch trainer jobs only when the trainer tab is opened.
- Refresh queue rows every 15 seconds but health stats only every minute.
- Add a lightweight `summary=true` stats mode that runs three focused queries instead of seven, with a 30-second client timeout.
- Redesign ArtJob cards around a large preview and surface destination title, page/route, output path, variant, request ID, and compact generation metadata.

## Validation

- Exact same head SHA as superseded PR #1535; that PR produced no GitHub Actions runs after open, reopen, or synchronize, so it was replaced to generate a clean PR event.
- Reviewed the complete five-file diff against the latest `main`; intervening main changes do not overlap these files.
- CI and Vercel preview pending.

### DIFF
--- a/components/art/art-manager.vue
+++ b/components/art/art-manager.vue
@@ -105,7 +105,7 @@
                 class="tab h-auto min-h-9 gap-2 rounded-xl px-4"
                 :class="{ 'tab-active': artJobWorkspaceTab === 'queue' }"
                 :aria-selected="artJobWorkspaceTab === 'queue'"
-                @click="artJobWorkspaceTab = 'queue'"
+                @click="selectArtJobWorkspace('queue')"
               >
                 Queue
                 <span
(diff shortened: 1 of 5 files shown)


silasfelinus:
Superseded by a final fresh PR opened from head `9333e3f`. This PR's open event carried the earlier SHA, and the connector-authored synchronize commit did not create GitHub Actions runs. Opening the final PR with the completed head already in place should produce checks without changing code.

PR #1537, opened 2026-08-06
Fix ArtJob queue preview fan-out and visual layout

silasfelinus:
## Root cause

The ArtJob queue was doing three expensive things at once:

- `performFetch()` passed relative `/api/...` URLs to native `fetch()` during SSR, which produced `Failed to parse URL from /api/icons`.
- Opening the queue also fetched the hidden trainer list and then eagerly requested full image data one ArtImage at a time, up to 240 requests.
- The 24-hour stats endpoint ran seven aggregate queries every 15 seconds while those image requests were competing for the database pool.

## What changed

- Resolve relative API URLs against the active Nuxt request origin during SSR.
- Stop eager full-image loading from queue and trainer fetches.
- Use the existing cacheable `/api/art/images/:id/file` media route for public/general previews, with native lazy image loading.
- Load protected/private previews only when explicitly requested.
- Fetch trainer jobs only when the trainer tab is opened.
- Refresh queue rows every 15 seconds but health stats only every minute.
- Add a lightweight `summary=true` stats mode that runs three focused queries instead of seven, with a 30-second client timeout.
- Redesign ArtJob cards around a large preview and surface destination title, page/route, output path, variant, request ID, and compact generation metadata.

## Validation so far

- Vercel production build completed for the code head.
- Preview `/api/icons` returns HTTP 200 with SmartIcon data, confirming the relative-URL failure is gone.
- The full `/artjob` preview is currently blocked by an unrelated existing SSR serialization error at `pageStore.initializePromise`; the ArtJob API itself builds and responds.
- Full GitHub Actions and final preview deployment pending.

### DIFF
--- a/components/art/art-manager.vue
+++ b/components/art/art-manager.vue
@@ -105,7 +105,7 @@
                 class="tab h-auto min-h-9 gap-2 rounded-xl px-4"
                 :class="{ 'tab-active': artJobWorkspaceTab === 'queue' }"
                 :aria-selected="artJobWorkspaceTab === 'queue'"
-                @click="artJobWorkspaceTab = 'queue'"
+                @click="selectArtJobWorkspace('queue')"
               >
                 Queue
                 <span
@@ -121,7 +121,7 @@
                 class="tab h-auto min-h-9 gap-2 rounded-xl px-4"
                 :class="{ 'tab-active': artJobWorkspaceTab === 'trainer' }"
                 :aria-selected="artJobWorkspaceTab === 'trainer'"
-                @click="artJobWorkspaceTab = 'trainer'"
+                @click="selectArtJobWorkspace('trainer')"
               >
                 Art trainer
                 <span
@@ -136,7 +136,7 @@
             <p class="hidden text-xs text-base-content/50 lg:block">
               {{
                 artJobWorkspaceTab === 'queue'
-                  ? 'Pipeline health, progress, and recovery controls'
+                  ? 'Jobs refresh every 15s; health summaries refresh every minute'
                   : 'Review finished renders and leave training feedback'
               }}
             </p>
@@ -238,6 +238,7 @@ type LegacyArtTab = ArtTab | 'upload'
 type ArtJobWorkspaceTab = 'queue' | 'trainer'
 
 const ART_JOB_REFRESH_INTERVAL_MS = 15_000
+const ART_JOB_STATS_REFRESH_INTERVAL_MS = 60_000
 
 const artJobStore = useArtJobStore()
 const artStore = useArtStore()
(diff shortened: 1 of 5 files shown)


silasfelinus:
REVIEWING: 2026-08-06T231426Z-artqueue-1537-q4n7 at 2026-08-06T23:14:26Z

silasfelinus:
Blocking review finding on head `9333e3f332531faedd60dfc06c6c38965d74e1a2`: the new fast public-preview path regresses video ArtJobs.

`save-generated.post.ts` explicitly accepts `mp4`, `webm`, `mov`, and `mkv` from image-to-video ArtJobs, and `utils/artImageSource.ts` deliberately classifies those outputs as `kind: 'video'`. In this PR, however, `publicImageSrc` is selected for every public/non-mature ArtImage, and `jobImageKind` then unconditionally returns `'image'` whenever that public URL exists. The card therefore renders a public video output in `` instead of `<video>`. The `/api/art/images/:id/file` route also does not advertise video MIME types for inline bytes, so public inline video cannot be rescued by the browser there either.

Please keep the no-N+1 performance win, but carry enough ArtImage media metadata in the queue response (ideally one batched metadata lookup for the page) to distinguish image vs video before choosing the public preview renderer, and make the file route serve supported video MIME types if it is intended to be the universal public media route. Add a regression contract for a public mp4/webm queue result.

Separately verified on the exact Vercel preview: `/api/icons` is now HTTP 200, so the relative-URL SSR parse failure is fixed. `/artjob` still returns HTTP 500 because `.pinia.pageStore.initializePromise` is being serialized; the same error also occurs on `/`, so that preview blocker is independent of this ArtJob diff.

silasfelinus:
Correction to my last paragraph after testing the exact base deployment: the Promise-serialization 500 is **not** present on current production `main` (`960b6125c2f7b10536700f6328845745839c46ae`), whose `/` SSR response is HTTP 200. On this PR's exact preview head, both `/` and `/artjob` are HTTP 500 with `.pinia.pageStore.initializePromise` pending at serialization.

So this is a second merge blocker, and it is branch-triggered even though the failing store is not an ArtJob component. The likely interaction is that the new SSR URL resolution turns formerly immediate failed relative fetches into real asynchronous requests, leaving an initialization Promise in serializable Pinia state at render time. Please treat the global SSR 500 as caused/exposed by this branch until the preview is 200 again. The `/api/icons` fix itself is verified, but we cannot trade the icon error for a site-wide SSR failure.

silasfelinus:
CI has now completed and confirms the PR is red independently of the two runtime blockers. Exact head `9333e3f332531faedd60dfc06c6c38965d74e1a2`:

- TypeScript Type Check `31130446576`: **failure**
- ArtJob Provenance Contract `31130446498`: **failure**, solely at its TypeScript step; its provenance checks themselves passed
- Open Issue Repair Contracts `31130447384`: **failure**, solely at its TypeScript step; repaired-issue contracts themselves passed
- Contract Tests, Layout Contract, API Client Follow-ups, Facet Catalog, MDC Scroll Ownership, narrative/storybook/taskmaster contracts: **success**

TypeScript diagnostics are local to this PR:
- `components/art/artjob-queue-card.vue(537,30)`: `new Date(job.completedAt)` receives `Date | null`
- `stores/artJobStore.ts(268,34)`: `new Date(job.updatedAt)` receives `Date | null`

Those nullability errors need fixing before the head can be green. I am leaving the PR open and unmerged with the review blockers above rather than weakening or bypassing CI.

PR #1542, opened 2026-08-07
Stage 3: the interact tier is a router, plus role-aware casting

silasfelinus:
Continues the Stage 3 UI pass. Two strands: making the interact tier a consistent *frame*, and the casting feature that came out of it.

## Rule 5 — the interact tier is a router, not a workspace

The house pattern is `manager → interact → gallery`, and the interact tier's whole job is one decision:

```
<x-gallery v-if="!selected" />     <x-workspace v-else />
```

`dream-interact` does that in 57 lines. Three others had entire working surfaces inlined, so the component that only picks a mode was carrying a thousand lines of something else. Rule 5 records each core object's interact line count and lets it **only shrink**.

| interact | before | after |
| --- | --- | --- |
| `scenario-interact` | 531 | 54 |
| `reward-interact` | 1097 | 61 |
| `character-interact` | 923 | 59 |
| `bot-interact` | 721 | 721 (next) |

Nothing about how any of these work changed — template and script moved across intact. Per Silas, the interact tier "is where we are actually hitting what we do with them uniquely," so the uniform thing is the frame, not the contents.

**Session state moved to the store, which is what made the splits possible.** Interact mode is not just "an object is selected" — a conversation already under way keeps you in the workspace *after* the object is deselected. That condition was answered by a private array (`sessionChatIds`, `chatMessages`) living in the very file that had to move, so each router would have lost sight of its own routing condition. Session membership is a fact about the chats, so `chatStore` owns it now, keyed by scope so Bots, Rewards and Characters don't share one session.

Two details worth stating:

- **The status banner stays in the router.** It reports on browse and work alike, so it belongs to the frame — the same reason `kr-manager` owns its one banner.
- **No class is passed to `<scenario-workspace>`.** It is multi-root (configure and play are siblings), so an attribute would have nowhere to land.

## Role-aware casting

Storybook could say *which* Characters were in a story — `setupDraft` carried a flat `castSlugs: string[]` — but not what any of them were **for**. The bible read:

```
Cast:
- Ami Butterfly — a cheerful chaos moth
- Cthulhu — sleeping god, presidential hopeful
```

and the narrator had to guess which one it was about. Now each cast line carries a role and its prompt descriptor, so "protagonist" actually means something downstream.

This **borrows a shape that already existed** rather than inventing one: `stores/helpers/stageCards.ts` has a working role system with min/max cardinality, badge art, a `promptDescriptor` woven into generation, and `assignCharacter()` resolving real Character records. `utils/narrativeRoles.ts` takes that shape and not its vocabulary.

The assigner is a **casting board**, not a form. The first version was rows of `<select>` — it worked, and it was wrong twice: it didn't feel like anything, and it dropped from art-forward cards to a dropdown mid-flow, one step below where the cast is *chosen* from cards. That is the same hodgepodge this pass exists to delete, so it's cards now: art at 2:3 through the shared `kr-art-plate` with `frame="none"` (an inner radius inside an outer one leaves visible notches — which is why the plate offers that mode), the assigned part worn as a badge *on* the card, and chips you press again to clear.

## Contract fix worth calling out

Splitting `scenario-workspace` out dropped the narrative-kit count 7/7 → 6/7 and reported scenario as a holdout. **The kit had not been abandoned** — the chat window and choice list moved one hop down with the rest of the working surface, and the counter walked a hand-kept `INTERMEDIARIES` list that didn't know about workspaces.

That is the same filename-shaped blindness the route-gallery contract had to learn out of: *a check that keys on a fixed list of names reports a refactor as a regression.* Replaced with a graph walk. It then held at **7/7 across the character split with no edit at all**, which is the evidence the fix was real and not a re-tuning.

Two further sharp edges, both caught by watching an assertion fail rather than by reading it:

- `KIT_USE` including `kr-art-plate` made the count unable to *fall*, so the metric couldn't report a regression.
- The predicate tested the whole file, so `import type { NarrativeTurn } from '…/kr-chat-window.vue'` counted as adoption. A mention is not a use.

## Verification

- `vue-tsc` exit 0
- route-gallery contract passes: 7/7 core routes, 0 shadow browsers, 6/6 tab managers on `kr-manager`, 14 gallery holdouts unchanged
- narrative-kit contract 7/7
- layout contract: no new violations
- `verifyNoPromiseInStoreState` still clean after merging main (the `chatStore` conflict here was two sides each adding exports; both kept)
- prettier and eslint clean on changed files

**Known papercut:** `verifyRouteGalleryContract.ts --update` writes raw `JSON.stringify`, so the baseline needs a `prettier --write` after it — otherwise a one-line ratchet step shows up as 59 lines of re-wrapping. Left as-is rather than changing the shared `ratchetBaseline` helper, which two other contracts also write through.

---


silasfelinus:
REVIEWING: 2026-08-07T021228Z-review-1542-f3c9 at 2026-08-07T02:12:28Z

silasfelinus (commented):
Two CI blockers on the exact head are actionable and both are in this PR's scope:

1. `Contract Tests` run 31140506607 fails `Capture-group guard contract` at `components/rewards/reward-workspace.vue:764`. The moved code does `const choiceNumber = match[1]` after `line.match(...)`; the ratchet evaluates new/changed call sites and requires guarded capture access. Keep the existing `if (!match) continue`, but use the contract-safe access shape (`match?.[1]` with the existing downstream validation, or another recognized guard shape). Do not weaken/baseline the checker.

2. `Entity Art Manager Contract` run 31140506623 fails because `verifyEntityArtManager.ts` still asserts `<EntityArtManager` is physically inside `character-interact.vue` (and it will then hit scenario/reward too). The PR intentionally moved those working surfaces into `character-workspace.vue`, `scenario-workspace.vue`, and `reward-workspace.vue`; I verified the Character workspace still contains the canonical EntityArtManager. Update the contract to follow the new ownership, while keeping `bot-interact.vue` and `facet-editor.vue` as their current owners. This is stale filename coupling, not a reason to put the manager back into the router.

The important behavior checks are otherwise healthy: exact-head TypeScript passed, route-gallery passed, narrative-kit is 7/7, and the new narrative-role contract passed end-to-end through the Storybook bible. Holding merge until these two required checks are green.

PR #1559, opened 2026-08-07
fix(nav): route scenario nav/audit entries to /stories, not dead /scenarios

silasfelinus:
### Task
conductor: interface-vision/t-102 (Build the canonical reachable-surface and responsive acceptance inventory) — a real navigation bug found incidentally while building the inventory, fixed as its own small scoped diff.

### What changed
`utils/scripts/coreObjectRoutes.ts` already documents `// NOT /scenarios.` — the Scenario object's real route is `/stories` — but three live Dream-narrator nav payloads in `stores/helpers/narratorHelper.ts` still pointed at the dead `/scenarios` path:
- the `'Scenarios'` nav action (`narratorNavigationTree`)
- the `'New Scenario'` builder create-spec (`narratorCreateSpecs`, `builderPath`)
- the `'Tell me a story'` starter prompt (`buildNarratorStarterPrompts`)

Same bug, independently, in `utils/scripts/auditResponsiveLayout.mjs`'s default `--routes` list — it was silently auditing a not-found page every run instead of the real Scenario surface (`/stories`).

All four now point at `/stories`.

### How I verified
- `npx eslint stores/helpers/narratorHelper.ts utils/scripts/auditResponsiveLayout.mjs` — clean
- `npx prettier --check` on both — clean on the actual diff (narratorHelper.ts carries one pre-existing, unrelated union-type formatting issue at lines 53–63 that predates this change and isn't touched by it)
- `npx vue-tsc --noEmit` (full project) — exit 0
- Confirmed no remaining `/scenarios` path references anywhere in either file after the fix

### Stakes
reversible

### Flags for Reviewer
None — pure literal string fixes, no logic change, no schema/API impact.

### Kaizen suggestion
The inventory build (t-102) also found a second, unrelated dead route: `stores/helpers/dashboardHelper.ts`'s `footer` dashboard's `builder` tab has `route: '/builder'`, which matches no content file or page anywhere, and `narratorHelper.ts`'s own `'builder'` nav action has the same dead `/builder` path. Left unfixed here since the correct destination isn't obvious (no single top-level page for the multi-tab Builder dashboard) — filed as a separate follow-up task rather than guessed at.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UJzWJk3ZWKSRKV3Ypm8kZD)_

### DIFF
--- a/stores/helpers/narratorHelper.ts
+++ b/stores/helpers/narratorHelper.ts
@@ -176,7 +176,7 @@ export const narratorNavigationTree: NarratorNavAction[] = [
     title: 'Scenarios',
     description: 'Choose a playable setup, location, or scene frame.',
     icon: 'kind-icon:map',
-    path: '/scenarios',
+    path: '/stories',
     flavor: 'The stage is waiting.',
     prompt:
       'Help me launch a scenario in this Dream. Use a vivid opening beat and end with choices.',
@@ -432,7 +432,7 @@ export const narratorCreateSpecs: NarratorCreateSpec[] = [
     type: 'scenario',
     title: 'New Scenario',
     icon: 'kind-icon:map',
-    builderPath: '/scenarios',
+    builderPath: '/stories',
     flavor: 'A new stage rolls into place.',
     prompt:
       'Create a single playable scenario for the Dream "{dream}". Return a title, a vivid location, a central tension, an opening intro beat, and 2–3 first choices.',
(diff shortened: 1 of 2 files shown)


silasfelinus (commented):
REVIEWING: 2026-08-07T111137Z-review-1559-a4k9 at 2026-08-07T11:11:37Z

Starting exact-head review of scope, required checks, mergeability, and companion Conductor closeout.

PR #1577, opened 2026-08-07
fix(nav): let channel-select shrink so the tab strip stops crushing on phone (interface-vision/t-109)

silasfelinus:
interface-vision/t-109. t-103 slice 6's completed 60-minute responsive audit (kind_robots run 31207144813, against merge commit 7841e27 / PR #1575) flagged `<nav#channel-tab-strip>` CRUSHED to an 18px sliver on phone (390px) across all six sanctuary-channel routes (`/about`, `/cart`, `/giving`, `/mermaids`, `/privacy`, `/sanctuary`) and to 12px on `/auth/google`. Tablet (820px) and desktop passed on all seven.

## Root cause

`channel-select` (the channel picker button, left of the tab strip) carried `shrink-0`, pinning it to its full unclamped label width (e.g. "Sanctuary") no matter how little room the header row had left.

The tab strip's own wrapper is `flex-1` — Tailwind's `flex-1` is `flex: 1 1 0%`, i.e. a flex-basis of **0%**. Flexbox distributes shrink deficit proportional to `flex-shrink × flex-basis`; a basis-0 item gets **zero** share of that deficit regardless of its shrink factor. So with channel-select refusing to shrink at all (`shrink-0`), literally 100% of any squeeze in the row got forced onto the tab strip instead — crushing it to a few px while channel-select rendered at full width, unclipped.

## Fix

Let channel-select actually participate in the shrink:
- `workspace-header.vue`: `class="shrink-0"` → `class="min-w-0 shrink"` at the one call site (confirmed via repo-wide grep — `channel-select` is only ever mounted here).
- `channel-select.vue`: its own root button also carried an internal `shrink-0`; replaced with `w-full min-w-0` so it now stretches to (and shrinks with) whatever width flexbox assigns it, letting the label span's pre-existing `min-w-0 truncate` finally engage instead of being unreachable dead code.

Net effect: channel-select now absorbs the deficit first and truncates its label under pressure, protecting the tab strip — which is what the header row exists for — instead of the reverse. On wide viewports (desktop/tablet, where there's no deficit to distribute) the rendered result is unchanged.

Left a comment at the call site walking through the flex-basis-0 mechanics for the next person who hits this class of bug elsewhere in the header.

## Verification

- `eslint` — clean
- `vue-tsc --noEmit` — clean
- `prettier --check` — clean
- `utils/scripts/verifyLayoutContract.ts` — holds, no new violations
[...]

PR #1578, opened 2026-08-07
fix(nav): give the tab strip an explicit min-width so it actually overflows (interface-vision/t-109 follow-up)

silasfelinus:
Follow-up to #1577 (interface-vision/t-109), which merged with a bug: it did not actually fix the crush it targeted.

## What happened

#1577 made `channel-select` shrinkable (`shrink` + `min-w-0` at the call site, `w-full min-w-0` replacing its internal `shrink-0`), on the theory that its old `shrink-0` was forcing 100% of the header row's shrink deficit onto `#channel-tab-strip`. Structural CI was green and the Vercel preview built successfully, and the PR was merged.

But the PR's **own** CI-triggered Responsive Layout Audit (kind_robots run [31212650601](https://github.com/silasfelinus/kind_robots/actions/runs/31212650601), against #1577's exact head `5a1893f`) shows the fix had **zero effect**: `/about`, `/cart`, `/giving`, `/mermaids`, `/privacy`, `/sanctuary`, and `/auth/google` all still crush `<nav#channel-tab-strip>` on phone (390px) to the **identical** `w=18` / `w=12` measured before the change. I've verified this directly from the raw job log (not just the summary line) — the merge review that approved #1577 appears to have misread that same log as passing.

## Why the first fix had no effect

`.tab-strip` sets `overflow-x: auto`. Per the flexbox spec, an item's **automatic minimum size** — the floor used when `min-width` is `auto`, the default — resolves to **0** on any axis with overflow other than `visible`. With no `min-width` set, the browser never detects the row as overflowing in the first place: `channel-select`'s `flex-grow: 0` only means "don't grow past your content," not "shrink to make room." So the shrink algorithm #1577 relies on never engages, and `channel-select` keeps its full label width regardless of how shrinkable it's marked — exactly what the identical-pixel measurement shows.

## Fix

Give `.tab-strip` an **explicit** `min-width` (96px / 112px / 128px, matching `tab-button`'s own existing floor at the same breakpoints). An explicit value bypasses the automatic-minimum-is-0 rule, so the browser now genuinely detects overflow once `channel-select`'s content plus this floor exceed the row — at which point `channel-select` (already shrinkable from #1577) is what actually gives ground.

## Verification

- `eslint`, `vue-tsc --noEmit`, `prettier --check`, `utils/scripts/verifyLayoutContract.ts` — all clean.
- Real verification is the same CI-triggered Responsive Layout Audit that caught this — watching for it against this PR's own preview before merging, and will read the raw per-route log rather than trusting only the pass/fail summary line, given what happened last time.

Roadmap: `projects/interface-vision/roadmap.yaml` t-109 (conductor) — will correct its note once this is confirmed green.

https://claude.ai/code/session_01A4hixPYBBpgafCc694Xad3

---
_Generated by [Claude Code](https://claude.ai/code/session_01A4hixPYBBpgafCc694Xad3)_

### DIFF
--- a/components/navigation/workspace-header.vue
+++ b/components/navigation/workspace-header.vue
@@ -36,19 +36,18 @@
            instead. -->
       <!-- channel-select used to carry `shrink-0` here, which pinned it to its
            full unclamped label width (e.g. "Sanctuary") no matter how little
-           room the row had left. Since the tab strip's own wrapper is
-           `flex-1` with a flex-basis of 0%, flexbox gives basis-0 items NO
-           share of the shrink deficit — so a shrink-0 sibling forces 100% of
-           any squeeze onto the tab strip instead, crushing `#channel-tab-strip`
-           to a several-pixel sliver on narrow phones (interface-vision/t-109:
+           room the row had left, crushing `#channel-tab-strip` to a
+           several-pixel sliver on narrow phones (interface-vision/t-109:
            /about, /cart, /giving, /mermaids, /privacy, /sanctuary, and
            /auth/google, all only on the 390px phone viewport — there was
            always enough room at tablet+). `shrink` (default flex-shrink: 1)
            plus `min-w-0` here, and the matching `w-full min-w-0` replacing
            channel-select's own internal `shrink-0` in channel-select.vue, let
-           the picker absorb the deficit first and truncate its label (the
-           span already had `min-w-0 truncate`, it just never got a chance to
-           engage) rather than starving the navigation the row exists for. -->
+           the picker absorb a shrink deficit and truncate its label instead
+           of starving the navigation the row exists for -- but by itself
+           this was NOT enough to fix the crush: see the `.tab-strip`
+           min-width comment below for why, and the empirical confirmation
+           that this half alone still crushed identically. -->
       <div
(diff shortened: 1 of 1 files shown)


silasfelinus:
REVIEWING: 2026-08-07T211411Z-pr1578-rv-x4n7 at 2026-08-07T21:14:11Z

silasfelinus:
Review result for exact head `f14ca3b1aebcb377b13567d14fe4cc6331162ac2`: code diff is scoped to `workspace-header.vue`; no unresolved review threads; all 13 pull-request workflows completed successfully, including TypeScript, Contract Tests, Layout Contract, MDC Scroll Ownership, Facet Catalog, API Client, Taskmaster, Narrative, and Storybook contracts; Vercel preview is READY. The branch is now mergeable despite main advancing through #1582, and the intervening main commits do not touch `workspace-header.vue`.

I am not merging yet because this is a geometry repair and AGENTS.md explicitly requires the deployment-triggered Responsive Layout Audit before a layout change lands. The connected workflow API only exposes pull-request-triggered runs, and no exact-head responsive-audit result is yet visible in the PR discussion. The sole remaining gate is therefore evidence that the seven t-109 routes no longer report the 12/18px `#channel-tab-strip` crush on this exact head. Do not substitute #1577's earlier run for that evidence.

PR #1571, opened 2026-08-07
Move the page-sized galleries onto kr-gallery, and give density a name (9 → 5)

silasfelinus:
Stage 3, PR 5: the two page-sized gallery holdouts, plus the instrument that measures whether any of this actually helped.

Route-gallery ratchet **9 → 5**. Remaining holdouts are all multi-collection panels: `achievement`, `friend`, `theme`, `dream-relationship`, `scenario-relationship`.

## The measurement, and the bad news

The chrome complaint **reproduces and is not fixed**. Measured against this PR's preview, pixels above the first card:

| route | phone (844) | tablet (1180) | laptop (768) | desktop (1080) |
|---|---|---|---|---|
| `/servers` | 962px **114%** | 1270px **108%** | 1254px **163%** | 1254px **116%** |
| `/icons` | 847px **100%** | 1081px **92%** | 1093px **142%** | 1093px **101%** |
| `/rewards` | 456px **54%** | 404px **34%** | 296px **39%** | 296px 27% ✓ |
| `/dreams` | 418px **50%** | 380px 32% ✓ | 352px **46%** | 352px 33% ✓ |
| `/bots` | 396px **47%** | 364px 31% ✓ | 296px **39%** | 296px 27% ✓ |
| `/stories` | 312px **37%** | 256px 22% ✓ | 268px **35%** | 338px 31% ✓ |
| `/characters` | 146px 17% ✓ | 154px 13% ✓ | 166px 22% ✓ | 166px 15% ✓ |

`/rewards` at 296px of a 1366×768 laptop is Silas's exact screenshot. `/servers` spends **more than a full screen** of that viewport before its first card.

The useful row is `/characters`: same shared shell, 13–22% at every width. So the waste is **route-owned chrome above the gallery**, not the gallery — per-route fixable, with `/characters` as the worked example. That's the next piece of work; it is not in this PR.

`/art` and `/conductor` read `no-cards` because they open on a *collections* grid — the converted image gallery only renders once you enter a collection. `/facets` and `/resources` also read `no-cards`, which is worth a separate look.

## Density is a fourth axis, deliberately, and closed

[...]

PR #1579, opened 2026-08-07
Spend down the chrome budget: /icons loses three stacked rows, and the audit stops guessing

silasfelinus:
Follow-up to #1571, which measured the chrome budget for the first time. This starts spending it down, and fixes a way the measurement could mislead.

## `/icons` was the worst route; three stacked rows are gone

It was over budget at **all four** widths — 1093px of a 768px laptop before the first icon. Four full-width blocks sat ahead of the grid:

1. a `text-3xl` title row
2. a `p-4` bordered panel holding **one** toggle
3. a row of two full-size selects
4. kr-gallery's own mode bar

The title and its two actions stay as one row. The Custom Icons toggle and both filters move into the shell's `#toolbar` slot — which exists precisely so filters share the mode bar's line instead of claiming rows of their own, and is the same move `reward-gallery` already made.

## The audit now says which selector produced each number

This is the more important half. `[data-kr-gallery-grid] > *` measures the shared shell's first card. `article` and `.card` measure whatever box happened to be first on the page.

That distinction is not cosmetic. **`/servers` opens on an overview tab that renders no gallery at all**, so its reported 1254px (163% of a laptop viewport — the worst number in the whole report) came from a fallback selector and is not gallery chrome in any sense. I was one step from "fixing" it. Fallback readings now print separately, labelled as indicative rather than measured.

## A correction to #1571's write-up

I described `/characters` (13–22%, passing everywhere) as "the worked example of what good looks like." That was wrong. `character-gallery` declares `title`, `subtitle` **and** `showHeader` props and renders none of them — all three are dead. `/characters` is lean by accident, not by design, and *fixing* those dead props would push it up to match the others rather than pulling the others down.

So the remaining budget work is to make the header block itself cheaper, not to copy `/characters`. Left for a follow-up, along with the phone-width readings (`/rewards` 54%, `/dreams` 50%, `/bots` 47%), which come from filter controls stacking one-per-row below `lg`.

## Verification

81/81 contract steps, `vue-tsc` clean, eslint clean (one pre-existing `vue/attributes-order` warning on an unrelated modal), prettier clean. The real numbers come from the post-deploy audit; I'll report them once it runs against this branch's preview.

---


silasfelinus:
REVIEWING: 2026-08-07T211411Z-pr1578-rv-x4n7 at 2026-08-07T21:22:00Z

#1578 is waiting on its required pixel audit; reviewing the other live UI PR in parallel rather than idling.

silasfelinus:
Review result for exact head `fae6c8771c49f7c76aec5f6534e708d701cca444`: inspected all six changed files. The control consolidation is coherent: `/icons`, Resources, Checkpoints, Project gallery, and Art gallery spend down stacked chrome without changing their data/store contracts; the gallery audit now distinguishes exact `[data-kr-gallery-grid] > *` measurements from fallback `article`/`.card` guesses. No unresolved review threads. All 13 pull-request workflows have now completed successfully, including TypeScript and Contract Tests, and Vercel preview is READY.

No code blocker found. I am not merging yet because this PR deliberately changes responsive chrome geometry. AGENTS.md requires the deployment-triggered browser audit before a layout PR lands, and that exact-head geometry result is not available through the connector's pull-request-only workflow-run view yet. Merge once the real gallery/responsive audit completes successfully (or its raw route output proves any failure is unrelated pre-existing debt).

Issue #1581, opened 2026-08-07
Recheck ProxySQL lanes after the 8/20 drive-crash recovery, then decommission HG20


silasfelinus:
**Rewritten 2026-08-21.** This started as *"Isolate Vercel Preview database capacity from production"* and is now down to one deferred verification plus one cleanup. The full history is in the thread; everything below is what's actually left.

## Resolved — nothing to do

- **Restart survival, the last real open question.** All four hostgroups read `ONLINE` in the 2026-08-21 census. The 2026-08-17 SHUNNED snapshot was transient: ProxySQL shuns a backend after connection errors and un-shuns after its recovery interval.
- **Denial counters.** `Access_Denied_Max_User_Connections` and `Access_Denied_Max_Connections` both `0`. The old 10037 baseline was erased by a restart and is not recoverable; `0` is the baseline now.
- **Capacity.** MariaDB `Max_used_connections` 18 of 100, `Connection_errors_max_connections` 0, no open transactions.
- **`default_schema = kindblank_fresh`** — flagged during the census, confirmed by Silas as correct production. Not drift.
- **The 4-hour transaction ceiling** — deliberate, not a leak. It exists because ComfyUI video generations run long on hardware near its limit. Rationale recorded in `docs/runbooks/proxysql-connection-capacity.md` so the next census doesn't re-flag it.

## The one thing to recheck, in a few days

The 2026-08-21 census showed `ConnERR 78` against `ConnOK 183` on HG10 — about a third of backend connection attempts failing. **These are almost certainly historical.** Alexandria lost a hard drive on 8/20; containers rebuilt against a replacement drive and then collided with the original data when it came back. The server was up at 19:00 on 8/20 and all services were healthy by roughly 02:00 on 8/21. ProxySQL's counters are cumulative since its last restart, so that whole event sits inside this window.

Confirming is one read-only command on Alexandria:

```bash
cd /mnt/user/appdata/kind_robots
bash scripts/proxysql-capacity-diagnostics.sh
```

**How to read it:** compare the `backend pool usage` row for hostgroup 10 against the 2026-08-21 baseline of `ConnOK 183 / ConnERR 78`.

- `ConnOK` climbed, `ConnERR` still **78** → historical. Close this issue.
- `ConnERR` climbed too → live, and worth chasing. Note the delta and reopen the investigation.

Also confirm `Access_Denied_Max_User_Connections` is still `0`.

## Cleanup: decommission HG20

[...]

PR #1582, opened 2026-08-07
Serialize database-backed responsive audits

silasfelinus:
PR #1580 reduced per-preview DB pressure and cancels stale same-branch deployment audits, but all Vercel previews still share the production ProxySQL identity until #1581 is completed.

This follow-up caps the heavyweight browser crawler itself:

- use one global `responsive-layout-audit-database` concurrency group across previews and the scheduled production sweep;
- keep `cancel-in-progress: false`, so one running audit can finish while GitHub retains at most one pending replacement instead of repeatedly restarting browser crawls during deployment churn;
- retain the deployment-freshness guard from #1580, so a queued preview whose branch has moved exits before touching the DB;
- update the database-capacity regression contract to require this global serialization.

Only the responsive-audit workflow and its existing capacity contract change. This is a temporary safety boundary while Preview and Production share one DB identity; #1581 tracks proper account-level isolation.

### DIFF
--- a/.github/workflows/responsive-layout-audit.yml
+++ b/.github/workflows/responsive-layout-audit.yml
@@ -44,13 +44,16 @@ on:
         required: false
         default: ''
 
-# A deployment URL is unique per commit, so keying concurrency by target_url
-# lets every preview commit run a separate 60-minute database-backed browser
-# crawl. Key by source ref instead: a newer push to the same branch cancels the
-# older audit, while separate PR branches remain independent.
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: 2026-08-07T211411Z-pr1578-rv-x4n7 at 2026-08-07T21:18:00Z

#1578 is blocked on its required responsive geometry evidence, so this session is using remaining review capacity on #1582 rather than idling.

PR #1595, opened 2026-08-08
Drop the chrome audit's fallback selectors — they were inventing numbers again

silasfelinus:
First production run of the gallery chrome audit against `main`, and the useful finding was about the instrument rather than the pages.

`/facets`, `/resources` and `/achievements` each reported an **identical** `1221 / 34 / 41 / 41`. Three unrelated routes agreeing to the pixel is the same "locked onto the app shell" signature that made the first version of this measurement worthless — except this time it was the `article` / `.card` **fallbacks** producing it, not the primary selector.

The cause is specific and worth keeping written down: `kr-gallery` renders its marked grid **only when it has items**, so an empty or still-loading gallery shows the skeleton or empty state instead. The fallback therefore fired exactly when there was no gallery to measure, and answered with a confident number for some unrelated `<article>`. "No cards yet" was the true answer, and `no-cards` already says it.

Those fallbacks existed for galleries not yet on the shared shell. The route-gallery ratchet reached **0**, so they cover nothing and only add noise that reads like signal. One selector remains. `matchedBy` stays as a field so a future fallback can't be reintroduced without the report being able to name it.

## Measurements so far, for the record

Real movement: `/rewards` phone 456px → 330px (54% → 39%), `/icons` laptop 1093px → 935px.

Inside budget at every width: `/characters` (17/13/22/15%) and `/stories` (22/18/29/20%).

Still over, and the remaining cost is **above the component** — component-level header trimming already happened on both and moved `/icons` only ~158px: `/icons` (81/81/**122**/87%) and `/servers` (86/62/**97**/75%).

## Verification

Full sweep across all 55 workflow files (92 `npm run` targets) — green except the two needing a live database. `vue-tsc` clean, prettier clean.

---

### DIFF
--- a/utils/scripts/auditGalleryChrome.ts
+++ b/utils/scripts/auditGalleryChrome.ts
@@ -72,7 +72,12 @@ export type ChromeReading = {
   breakpoint: string
   viewportHeight: number
   chromePx: number | null
-  /** Which selector produced the number. The marked one is exact; the rest are guesses. */
+  /**
+   * Which selector produced the number. Only ever the marked grid now -- the
+   * `article`/`.card` fallbacks were removed once every gallery was on the
+   * shared shell. Kept as a field so a future fallback cannot be reintroduced
+   * without the report being able to say so.
+   */
   matchedBy: string | null
   fraction: number | null
   /**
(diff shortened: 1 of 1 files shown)


silasfelinus (commented):
REVIEWING: 2026-08-07T191153-0700-pr1595-rv-m7q2

Claiming substantive review of the exact current head. I’ll verify the full diff, threads, mergeability, and completed required checks before any merge.

silasfelinus (commented):
Exact-head review found one blocker before merge. The behavioral change is sound: with the route-gallery ratchet at zero, removing `article` / `.card` fallback measurements prevents empty/loading galleries from producing confident app-shell numbers. All 12 registered exact-head PR workflows are completed green and there are no review threads.

The blocker is repository convention: this patch adds a large explanatory block comment and expands the `matchedBy` doc comment, while current `kind_robots/AGENTS.md` explicitly says **Avoid inline and template comments. Let naming carry the meaning.** Keep the useful behavior, but remove the newly-added explanatory comment prose (a short existing-style type doc only if genuinely necessary), then rerun exact-head checks. No behavioral rewrite requested.

GitHub will not let this account submit REQUEST_CHANGES on its own PR, so recording the blocking review as COMMENT instead.

PR #1585, opened 2026-08-07
Stop crushing three interface-vision Phase-1 findings: /storybook, /play/challenges, /taskmaster

silasfelinus:
interface-vision/t-110 + t-112: three related phone/desktop crush findings from the same audit run ([31207144813](https://github.com/silasfelinus/kind_robots/actions/runs/31207144813)), same underlying flexbox mechanism, batched into one PR since the audit queue is serialized (one run at a time) and both fixes are small/independent.

## t-110: `/storybook` and `/play/challenges` title blocks crush on phone

Both routes have a header row: `shrink-0` icon, `min-w-0 flex-1` title block, `shrink-0` button group. Tailwind's `flex-1` sets `flex-basis: 0%`, and `min-w-0` removes the automatic min-content floor — so during flexbox's line-collection step the title block contributed ~0 to "does this line fit," and never wrapped onto its own line. All three siblings crammed onto one line on phone (390px), and the entire shrink deficit landed on the one flexible child: a **10px sliver on `/storybook`**, a **20px sliver on `/play/challenges`** (`CRUSHED <div> ... w=10 .min-w-0 flex-1`, `w=20 .min-w-0 flex-1`).

Fix: `basis-full sm:basis-auto` on each title block, forcing it to claim its own row on phone; reverts to the compact single-line layout at `sm:` where both routes already pass.

## t-112: `/taskmaster`'s recipe-panel ingredient cards crush on desktop

`narrative-ingredient-picker`'s grid used `xl:grid-cols-3` — a **viewport** breakpoint. Its only call site (Taskmaster's "recipe" panel) is one of three columns in a page grid that *also* switches at the same `xl` breakpoint, so a 1440px viewport meant a ~390-430px sidebar column was asked to hold three cards at once. Each card's fixed `w-28`/`sm:w-32` image ate most of that column, crushing the flexible text span to a **24px sliver** on all six option cards (`CRUSHED <span> ... w=24 .min-w-0 flex-1`).

Fix: drop the `xl:grid-cols-3` tier. `md:grid-cols-2` already fits the sidebar comfortably, and both current call sites (Setting, Genre/mood/style) only ever render in this narrow column — no wide-usage case depended on the third column.

## Verification
- `eslint` clean on all three changed files
- `vue-tsc --noEmit` clean (full project typecheck)
- No data/store contracts touched — pure layout class changes
- Will report the deployment-triggered `audit` job's raw per-route result before merging (queue is serialized behind other in-flight audit runs, so this may take a while)

[...]

PR #1598, opened 2026-08-08
interface-vision/t-108: dedupe simultaneous duplicate-URL image requests on /navigation

silasfelinus:
### Task
interface-vision/t-108: `/navigation`'s site-directory (~85 icons/page) flakes BROKEN-ART on a different image set each responsive-audit run (kind: software)

### What changed
- `composables/useDedupedArtwork.ts` (new): module-scoped in-flight map keyed by image URL. The first occurrence of a URL preloads it directly (native `Image()`); every later occurrence for the same URL awaits that first preload's settlement (success or failure) before pointing its own `` at the URL.
- `components/navigation/navigation-artwork.vue` (new): thin wrapper so the composable can be called once per rendered tab (Vue composables must be called at a component's setup top level, not per `v-for` iteration).
- `components/navigation/navigation-trimmed.vue`: swapped the raw `` in the site-directory card for `<NavigationArtwork>`.

### Why (root cause, continuing t-108's investigation)
kind_robots#1592 (merged 2026-08-08T02:12:49Z) added `Cache-Control: public, max-age=3600` to the `/images/*` redirect. The **post-merge** `responsive-layout-audit` run (31234573743, against the exact merge commit `d8d7b857`) still shows `/navigation` BROKEN-ART on phone (6 images) and desktop (3 images) — the same shared-URL images as every prior run (`dashboard-tabs/user/profile.webp`, `dashboard-tabs/giftshop/community.webp`).

Why #1592 didn't move the needle: caching the redirect response only lets a *second* visit reuse a *first* visit's cache entry. It does nothing when N `` tags with the identical `src` all fire their first-ever GET for that path in the same tick on the very first page load — which is exactly what `/navigation` does with its ~85 tab icons, several of which resolve to one identical dashboard-tab image shared by multiple content entries (confirmed via `content/channels/**/*.md` frontmatter in t-108's earlier investigation).

This PR is complementary to #1592, not a replacement: #1592 makes the response cacheable at all; this makes sure duplicate requests are staggered so that cache actually gets a chance to be warm before the next duplicate fires.

### How I verified
- `npx eslint` on all three changed/new files — clean.
- `npx vue-tsc --noEmit` — clean.
- `npx prettier --check` — clean.
- No live GitHub Actions dispatch available from this sandbox — this PR's own deployment-triggered `responsive-layout-audit` is the real verification, same as t-108's prior rounds. Recommend reading the raw per-route log (not just the pass/fail summary) before merging, per the task's own established precedent.

### Stakes
reversible — new composable + wrapper component, one call-site swap; no other app code touched.

### Flags for Reviewer
This is t-108's 3rd attempt at a fix (2 prior fixes shipped and were then disproven by a fresh audit read). Please don't merge on structural CI alone — wait for this PR's own `responsive-layout-audit` result and read the raw `/navigation` BROKEN-ART lines directly.

---
_Generated by [Claude Code](https://claude.ai/code/session_0142MG9h3Ho8auC1UNJKwQRc)_

### DIFF
--- a/components/navigation/navigation-artwork.vue
+++ b/components/navigation/navigation-artwork.vue
@@ -1,0 +1,21 @@
+<!-- /components/navigation/navigation-artwork.vue -->
+<template>
+  <img
+    v-if="resolvedSrc"
+    :src="resolvedSrc"
+    :alt="alt"
+    loading="lazy"
+    class="h-full w-full object-cover"
+  />
+</template>
+
+<script setup lang="ts">
+import { useDedupedArtwork } from '@/composables/useDedupedArtwork'
+
+const props = defineProps<{
+  src?: string
+  alt?: string
+}>()
+
+const resolvedSrc = useDedupedArtwork(() => props.src)
+</script>
--- a/components/navigation/navigation-artwork.vue
+++ b/components/navigation/navigation-artwork.vue
@@ -1,15 +1,18 @@
 <!-- /components/navigation/navigation-artwork.vue -->
 <template>
-  <img
-    v-if="resolvedSrc"
-    :src="resolvedSrc"
-    :alt="alt"
-    loading="lazy"
-    class="h-full w-full object-cover"
-  />
+  <span ref="root" class="block h-full w-full">
+    <img
+      v-if="resolvedSrc"
+      :src="resolvedSrc"
+      :alt="alt"
+      loading="lazy"
+      class="h-full w-full object-cover"
+    />
+  </span>
 </template>
 
 <script setup lang="ts">
+import { onBeforeUnmount, onMounted, ref } from 'vue'
 import { useDedupedArtwork } from '@/composables/useDedupedArtwork'
 
 const props = defineProps<{
@@ -17,5 +20,26 @@ const props = defineProps<{
   alt?: string
 }>()
 
-const resolvedSrc = useDedupedArtwork(() => props.src)
+const root = ref<HTMLElement>()
+const { resolvedSrc, request } = useDedupedArtwork()
+
+let observer: IntersectionObserver | undefined
+
+onMounted(() => {
+  if (!props.src || !import.meta.client || !root.value) return
+  observer = new IntersectionObserver(
+    (entries) => {
+      if (entries.some((entry) => entry.isIntersecting)) {
+        request(props.src as string)
+        observer?.disconnect()
+      }
+    },
+    { rootMargin: '200px' },
+  )
+  observer.observe(root.value)
+})
+
+onBeforeUnmount(() => {
+  observer?.disconnect()
+})
 </script>
--- a/components/navigation/navigation-trimmed.vue
+++ b/components/navigation/navigation-trimmed.vue
@@ -74,12 +74,9 @@
             <span
               class="relative flex h-11 w-11 shrink-0 overflow-hidden rounded-lg bg-base-200"
             >
-              <img
-                v-if="tab.image"
+              <NavigationArtwork
                 :src="tab.image"
                 :alt="tab.title || tab.label"
-                loading="lazy"
-                class="h-full w-full object-cover"
               />
               <span
                 class="absolute inset-0 flex items-center justify-center bg-base-content/15"
--- a/composables/useDedupedArtwork.ts
+++ b/composables/useDedupedArtwork.ts
@@ -1,0 +1,48 @@
+// /composables/useDedupedArtwork.ts
+//
+// interface-vision/t-108: /navigation renders ~85 dashboard-tab icons, and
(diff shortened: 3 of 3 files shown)


silasfelinus:
REVIEWING: 2026-08-08T041459Z-pr1598-rv-r8k4 at 2026-08-08T04:14:59Z

silasfelinus (commented):
Review decision: changes required before merge.

The branch is clean against current main and all 13 pull-request-triggered workflows are green, but this implementation introduces a load-pattern regression directly in the failure domain it is trying to repair. `useDedupedArtwork()` constructs a native `Image()` probe during setup for every distinct URL. That bypasses the original `` behavior: `/navigation` can now eagerly start probes for the whole ~85-icon directory instead of letting the browser defer offscreen images. Duplicate URLs are serialized, but the total first-load burst across distinct URLs can become materially larger, which can worsen the media-origin pressure behind BROKEN-ART.

There is also a repo-convention cleanup needed: the new composable carries a large explanatory inline-comment block despite AGENTS.md's "Avoid inline and template comments" rule. Keep the rationale in the PR/task record and let the implementation naming carry it.

Please preserve lazy/visibility-driven loading while deduping repeated URLs rather than probing every URL at component setup. Because t-108 has already shipped two plausible fixes that fresh audits disproved, the corrected exact head still needs its own responsive-layout-audit raw `/navigation` result before merge. Do not close from structural CI alone.

silasfelinus:
Pushed `9edf120` addressing the review feedback on head `65d4382`:

- `navigation-artwork.vue` now gates the dedup `request()` call behind an `IntersectionObserver` (`rootMargin: 200px`) instead of firing a `new Image()` probe eagerly on mount for every tab. Off-screen tabs no longer start a fetch until they approach the viewport, restoring the visibility-driven behavior the original raw `` had, while duplicate URLs among tabs that do become visible together still dedupe through the shared in-flight map.
- `useDedupedArtwork.ts` now exposes an explicit `request(url)` call instead of firing from a `watchEffect` on mount, so the component controls timing.
- Removed the explanatory inline-comment block per kind_robots/AGENTS.md ("avoid inline and template comments, let naming carry the meaning").

`eslint`, `vue-tsc --noEmit`, and `prettier --check` all clean on the two changed files.

Per the standing note on this task: **please don't merge on structural CI alone** — wait for this head's own `responsive-layout-audit` deployment run and read the raw `/navigation` BROKEN-ART lines before merging.

---

silasfelinus:
REVIEWING: 2026-08-08T051713Z-pr1598-rv-5d45 at 2026-08-08T05:17:13Z

silasfelinus (commented):
Re-review of corrected exact head `9edf120a6da2c33aac592ce4f32f8f51b455c412`: the prior code defect is fixed. `navigation-artwork.vue` now defers `request()` behind an `IntersectionObserver` with a 200px root margin, so offscreen directory entries no longer eagerly create native `Image()` probes; duplicate URLs that actually approach the viewport still converge on the shared in-flight promise. The explanatory comment block called out in the prior review is also gone. I found no remaining code-level blocker and there are no unresolved review threads.

All 13 pull-request-triggered workflows for this exact head are completed successfully, and the matching Vercel preview deployment is READY. I am still not merging because t-108's own retry note and PR handoff explicitly require the deployment-triggered `responsive-layout-audit` raw `/navigation` result before closure, and the connected workflow-run action available in this session only returns pull-request-triggered runs. No exact-head raw audit evidence is present in the PR discussion yet. Given that earlier t-108 fixes looked plausible until the raw audit disproved them, that gate remains real rather than ceremonial.

Code review: pass. Remaining merge gate: exact-head responsive-layout audit evidence showing the `/navigation` BROKEN-ART regression is actually gone.

silasfelinus:
This head's own `responsive-layout-audit` (run [31239732003](https://github.com/silasfelinus/kind_robots/actions/runs/31239732003)) completed. Read the raw per-route log, not just the pass/fail summary:

```
04:51:10Z phone   /navigation    ✅
05:05:52Z tablet  /navigation    ✅
05:20:47Z desktop /navigation    ✅
```

`/navigation` is clean at all three viewports — zero BROKEN-ART, first time in this task's 4 fix attempts. The `IntersectionObserver`-gated dedup fix works.

The job's overall conclusion is still `failure`, but the one flagged route/viewport is unrelated to this task: desktop `/taskmaster` CRUSHED (6 scenario-option spans) — this is a **regression** of interface-vision/t-112, which was previously verified fixed via PR #1585 (merged 2026-08-08T00:42:41Z). Filing that separately rather than expanding this PR's scope.

Merging — `/navigation`'s BROKEN-ART flake is resolved.

---

PR #1604, opened 2026-08-08
Repair Art Trainer checkpoint provenance and stale SDXL retries

silasfelinus:
## What changed
- Art Trainer SDXL revisions now resolve the source/requested checkpoint through an active CHECKPOINT Resource instead of silently inheriting the stale DreamShaper fallback
- SDXL trainer jobs preserve `checkpointResourceId` + current Resource `localPath` in payload provenance
- normal queue requeue refreshes checkpoint Resources in addition to the existing LoRA refresh
- SDXL img2img workflow construction now fails closed when no explicit checkpoint Resource path was supplied, rather than manufacturing a doomed ArtJob with a hard-coded filename
- add `utils/scripts/repairTrainerCheckpointFailures.ts`, dry-run by default, to repair existing FAILED trainer jobs in place from their source ArtImage/source ArtJob checkpoint Resource
- the repair refuses to requeue if the Resource catalog still points at the exact checkpoint Comfy rejected
- add focused checkpoint resource contract coverage

## Root cause
The Art Trainer SDXL img2img path rebuilt the graph without passing a checkpoint, so `buildSdxlImg2ImgWorkflow()` inserted the hard-coded `SDXL/dreamshaperXL_v21TurboDPMSDE.safetensors`. Current ComfyUI no longer lists that checkpoint. The trainer route then deleted `payload.resources`, discarding the Resource identity that could have refreshed a moved model on retry.

The observed production failure is a ComfyUI `CheckpointLoaderSimple` HTTP 400 (`ckpt_name ... not in list`), not a missing prompt.

## Production recovery after merge/deploy
On Alexandria:

```bash
npx tsx utils/scripts/repairTrainerCheckpointFailures.ts
npx tsx utils/scripts/repairTrainerCheckpointFailures.ts --apply
```

The script touches only FAILED COMFY trainer-redo rows whose model is SDXL and whose error is a checkpoint `value_not_in_list`. It updates the same rows back to PENDING at priority >=100, resets attempts, and does not create duplicates.

### DIFF
--- a/.github/workflows/artjob-provenance-contract.yml
+++ b/.github/workflows/artjob-provenance-contract.yml
@@ -11,6 +11,8 @@ on:
       - "server/utils/artJobRetry.ts"
       - "utils/scripts/verifyArtJobProvenance.ts"
       - "utils/scripts/verifyArtJobLoraOverride.ts"
+      - "utils/scripts/verifyArtJobCheckpointRefresh.ts"
+      - "utils/scripts/repairTrainerCheckpointFailures.ts"
       - ".github/workflows/artjob-provenance-contract.yml"
   pull_request:
     branches: [main]
@@ -22,6 +24,8 @@ on:
       - "server/utils/artJobRetry.ts"
       - "utils/scripts/verifyArtJobProvenance.ts"
       - "utils/scripts/verifyArtJobLoraOverride.ts"
+      - "utils/scripts/verifyArtJobCheckpointRefresh.ts"
+      - "utils/scripts/repairTrainerCheckpointFailures.ts"
       - ".github/workflows/artjob-provenance-contract.yml"
   workflow_dispatch:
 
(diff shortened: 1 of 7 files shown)


silasfelinus:
REVIEWING: 2026-08-08T081327-0700-pr1604-rv-7c91 at 2026-08-08T08:13:27-07:00

silasfelinus (commented):
Blocking regression on this exact head: `buildSdxlImg2ImgWorkflow()` now throws when `checkpoint` is absent, but the existing entity-art flow explicitly offers `checkpointResourceId = 0` as “Use current quality default” and omits both `checkpointResourceId` and `checkpoint` from `/api/art/enqueue` in that state. `server/api/art/enqueue.post.ts` passes `body.checkpoint ?? null` straight into this shared builder. So the default SDXL img2img path that currently queues successfully will start failing before ArtJob creation after this merge.

Please keep the fail-closed checkpoint provenance goal, but repair the caller contract too: resolve the default/current SDXL checkpoint through an active CHECKPOINT Resource before `buildSdxlImg2ImgWorkflow()` is called, or change the UI/API contract so an explicit Resource is required and the default option cannot submit. Add focused coverage for `/api/art/enqueue` with the no-explicit-checkpoint/default UI state so this cannot regress silently.

Everything else I inspected in the trainer repair is directionally sound, and the current exact-head ArtJob provenance + TypeScript/contract workflows are green. Do not merge until this shared-call-site regression is fixed and fresh exact-head checks complete.

PR #1605, opened 2026-08-08
Paginate every gallery in the shell, drop the Resource _count, fix 320px selects

silasfelinus:
Reported from a tablet: *"Really think we might be getting an index overwhelm issue when getting resources. There seems to be no pagination. Are we trying to load thousands on one page? It freezes. Also, several rows on load."*

Both halves confirmed, with **different causes** — and the pagination half turned out to be far more widespread than `/resources`.

## 1. The freeze — and it is not just Resources

`GET /api/resources` is a `findMany` with no `take`/`skip`, and `resource-gallery` handed the *whole* filtered result to `kr-gallery`, so the browser built a card for every row in the catalog at once.

My first pass fixed that in `resource-gallery` and described it as the outlier. Silas asked whether that was actually true — *"do you feel confident on that? We won't hit this problem with others if they grow giant?"* — and it wasn't. Checking all thirteen `kr-gallery` mounts:

| | galleries |
| --- | --- |
| pages itself | `art-gallery` (user-controlled page size) |
| caps itself | `facet-gallery` (show-more per taxonomy group) |
| **does not page** | **bot, character, dream, reward, scenario, icon, server, checkpoint, stylist-client, theme, achievement** |

**Eleven of thirteen had the same defect.** `art-gallery` was the outlier for *having* pagination. `/resources` merely hit the wall first because it's the biggest table — every other gallery was waiting to grow into it.

So paging moves **into the shell**, where forgetting it isn't possible, rather than being pasted into eleven more files.

`kr-gallery` gains `pageSize` (default 48; `0` opts out):

- The pager renders **only when there is a second page**, so galleries that fit on one are visually unchanged.
- It rides the **existing sticky bar**, not a new band. The bar sticks inside whatever ancestor scrolls, so the control stays reachable instead of sitting a thousand cards below the fold — and a new band would spend a row this stage is trying to save.
- `pageCount > 1` joins that bar's own `v-if`, because `theme-gallery`'s two grids pass `:modes="[]"` and no toolbar and would otherwise page with no way to turn the page.
- Page index **clamps rather than resets**: `items` recomputes on every keystroke of a parent's search box, so resetting to 0 on change would make any page but the first unreachable.
- `art-gallery` and `facet-gallery` pass `:page-size="0"` — both already page, and slicing a page into pages is the one way this could regress them.

`resource-gallery`'s own pagination is removed in favour of the shell's. What stays whole-catalog there is deliberate: `filteredResources` still spans the entire array, so search and both dropdowns cover the whole catalog rather than the visible page, and `resourceById` is a superset of whatever the shell renders.

## 2. The extra rows — a DaisyUI default

DaisyUI 5 sets `.select { width: clamp(3rem, 20rem, 100%) }`. The **preferred** value is `20rem`, so any select with no width utility is **320px wide regardless of its options**. Two of them claimed 640px of the filter row and pushed everything else onto its own line. The search box looked fine only because it already carried `w-36 sm:w-52`.

This also finishes #1602, which I got half-right: removing `w-full` from the bot/reward selects dropped them to DaisyUI's 320px default rather than content width. `bot-gallery`, `reward-gallery` and `icon-gallery` get `w-auto max-w-44` here — and that's also the real explanation for `/icons` measuring a phone-only 108px gap I'd earlier mis-attributed to wide button labels.

## 3. The database cost — dropped on Silas's call

The list select carried `_count: { select: { ArtImages: true, UsedInImages: true } }`, making the query emit **two correlated aggregates per row** against `ArtImage` — the most expensive thing in an already-unpaginated `findMany` over the whole table, and the likeliest source of the `/resources` pool timeouts.

What it bought was two badges reading `0 checkpoint uses / 0 LoRA uses` — zero on every card in the report. Removed from the select, the store type, and the card. If those counts return they belong on the detail fetch or a per-page endpoint, not on every row of the listing.

## Verification

- `npm run test` (vue-tsc) — clean, which also confirms dropping `_count` broke no consumers
- `npx eslint` + `npx prettier` on all changed files — clean. `art-gallery` reports 2 `no-dynamic-delete` errors; confirmed **pre-existing** by stashing and re-running (same 2, shifted 5 lines by prettier).
- Swept **193 verification commands across all 57 workflow files**. 188 pass; the same 5 fail as on an unmodified tree, all Prisma pool timeouts (`P2039`) from the sandbox having no DB egress. None touch these files.
- Rebased onto `main` at `111bfee`.

### DIFF
--- a/components/art/art-gallery.vue
+++ b/components/art/art-gallery.vue
@@ -454,8 +454,13 @@
           `density`; a second row of Cards/Heroes/Icons buttons would be two
           pickers steering one grid.
         -->
+        <!-- page-size 0: art-gallery pages itself. `galleryItems` is built
+             from pagedActiveImages, and its page size is a user control
+             (imagePageSize), so letting the shell page again would slice a
+             page into pages. -->
         <kr-gallery
           :items="galleryItems"
+          :page-size="0"
           :modes="[]"
           :density="viewSize"
           empty-label="art images"
--- a/components/bots/bot-gallery.vue
+++ b/components/bots/bot-gallery.vue
@@ -332,7 +332,7 @@
 
             <select
               v-model="constructionFilter"
-              class="select select-bordered select-sm bg-base-100"
+              class="select select-bordered select-sm w-auto max-w-44 bg-base-100"
               aria-label="Filter bots by construction status"
             >
               <option value="all">All statuses</option>
--- a/components/facets/facet-gallery.vue
+++ b/components/facets/facet-gallery.vue
@@ -96,8 +96,11 @@
              so mounting it N times is structurally fine, and `:modes="[]"` drops
              its own mode bar -- the picker above drives every group at once, so
              the groups stay visually consistent instead of drifting apart. -->
+        <!-- page-size 0: entries are already capped per taxonomy group by
+             this file's own show-more (`slice(0, shown[taxonomy] ?? PAGE)`). -->
         <kr-gallery
           :items="group.entries.map(toGalleryItem)"
+          :page-size="0"
           :mode="mode"
           :modes="[]"
           empty-label="facets"
--- a/components/gallery/kr-gallery.vue
+++ b/components/gallery/kr-gallery.vue
@@ -52,11 +52,56 @@
       have not adopted it are unaffected.
     -->
     <div
-      v-if="modes.length || $slots.toolbar"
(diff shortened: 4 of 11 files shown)


silasfelinus:
REVIEWING: 2026-08-08T091417Z-pr1605-rv-c7f3 at 2026-08-08T09:17:00Z

PR #1636, opened 2026-08-09
Move the Resource actions onto the card back, and fix the edit scroll

silasfelinus:
> "the buttons that are currently on the gallery card before clicking need to be moved to the bottom row of the selected back: add to build and start fresh should be one button. I assume preview is some sort of generate option, but I have no idea what. And just kill upload … Edit option is starting to look good, but we cannot scroll."

## The scroll — and this one was my last fix's fault

#1631 bounded the panel **and** gave the card back its own `overflow-y-auto`. That created two nested scrollers whose inner parent had no bounded height, so the inner region never became scrollable and the outer one never received the overflow. **Neither scrolled.**

`kr-card-flip`'s panel is the single scroll owner again; the back is plain flow inside it. The footer scrolls with the body as a result, which is the right trade — an action bar pinned over a form you can't reach is worse than one you scroll to.

## The buttons

Four of them — Add to build, Start fresh, Preview, Upload — sat under every tile in a grid of 48. Gone from the front, which now carries art, name and trigger: the things that make a Resource *findable*. Everything you *do* to one lives on the back, where there's room to label it and it isn't repeated 48 times down the page.

`kr-card-back` gains an **`#actions` slot** for them, because *which* actions exist is per-object — a Resource joins an art build, a Bot doesn't. Same reason `#details` is a slot, and it keeps that file free of model names.

**One build button.** "Add to build" appended to the current art form; "Start fresh" *replaced* it with only this Resource and navigated away. The additive one survives — collapsing to the destructive variant would let a single tap discard a build someone was part-way through, and adding to an empty build already *is* starting fresh, so nothing is lost.

**Preview, explained by when it appears.** It queues art generation for a Resource, which is unguessable from the word "Preview". It's now **"Generate art"** on a Resource with no image and **"Regenerate art"** on one that has it — a button that shows up exactly when it applies explains itself.

**Upload is gone**, along with its file-input plumbing, the busy flag that drove its spinner, and the fixture props that fed it.

## Verification

- `npm run test:route-gallery-contract` — passes, 0 holdouts
- `npx eslint` — clean; it caught `isEditable`, `startGeneration`, `choosePreviewFile`, `readFileAsDataUrl`, `navStore` and `activeUploadResourceId` going dead as the buttons left, all removed rather than silenced

---

### DIFF
--- a/components/gallery/kr-card-back.vue
+++ b/components/gallery/kr-card-back.vue
@@ -128,7 +128,18 @@
       floored at content height and the panel grows instead, which is the same
       failure the stage's height fix addresses from the other end.
     -->
-    <div class="min-h-0 flex-1 overflow-y-auto p-3">
+    <!--
+      NO overflow HERE. kr-card-flip's panel is already the scroll owner, and
+      giving this its own `overflow-y-auto` made two nested scrollers where the
+      inner one's parent (this section) has no bounded height -- so the inner
+      region never became scrollable and the outer one never received the
+      overflow. Silas, 2026-08-09, on the edit view: "we cannot scroll."
+
+      Plain flow content, one scroller above it. The footer scrolls with the
+      body as a result, which is the right trade: an action bar pinned over a
+      form you cannot reach is worse than one you scroll to.
+    -->
+    <div class="p-3">
       <!--
         The edit form REPLACES the info body rather than flipping again --
         "the edit option is just an on-screen change to the modal". A second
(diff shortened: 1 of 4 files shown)


silasfelinus:
REVIEWING: 2026-08-08T20:13:51-0700-pr1636-rv-q9f7 at 2026-08-09T03:13:51Z

silasfelinus (commented):
Exact-head review complete for `de0e951b34baff3aac5ce5f3a6f91780ea79d260`. The change matches the requested Resource-card behavior: front-card actions are removed, upload and destructive start-fresh plumbing are removed, generation/add-to-build move to the selected back, and scrolling has one owner again. No unresolved review threads. All 13 observed exact-head Actions workflows completed successfully, including TypeScript, Contract Tests, Layout Contract, and MDC Scroll Ownership. The Vercel preview is READY and `/resources` returns HTTP 200 with the exact head SHA as its public buildId. Safe to merge.

PR #1638, opened 2026-08-09
Quieten the Bot cards, stop the card back cropping art, scroll the chat

silasfelinus:
Three from a tablet.

## "we don't need the designer or theme text"

Both were near-constant down the page — the same handle on every card, and a palette name that the card's own `data-theme` border **already shows**. So the strip was chips repeating what the eye had, on all 69 tiles.

The theme still colours the frame; it just no longer also spells itself out. `serverName` survives because it varies and nothing else on the card carries it, so in practice the strip is now usually empty. Full provenance lives on the card back, which is what the back is for.

The border stays — glad it landed.

## "layout should be card like and not cut off the image"

The back's art plate was `object-cover`, which fills by **cropping** — on a tall Bot portrait that sliced the head off, exactly as your screenshot shows. It's `object-contain` now: the whole frame shows and the remainder letterboxes against the plate.

The title came out of the scrim as a consequence. A scrim over a *contained* image sits on empty plate as often as on artwork, so it read as a bar rather than a caption. Art up top, name beneath it — the shape an actual card back has.

## "interact cannot scroll, same as with resource"

Same symptom, **different cause** — worth separating, because the card-back fix (#1636, merged) wouldn't have touched this.

`bot-chat`'s grid is `overflow-hidden`, which is correct at `xl` where the two columns sit side by side, each bounded and each with its own inner scroller. Below `xl`, `grid-cols-1` stacks them, the stack is routinely taller than the pane, and the overflow was clipped — so there was no way to reach the settings panel underneath the conversation.

The hidden overflow is now the wide case only. The panes keep their own scrollers at both sizes, so this doesn't create two competing scroll owners at `xl` — the mistake that broke the card back in #1631 and had to be undone in #1636.

## Honest caveat

**The chat fix is unverified visually.** It's reasoned from the source and passes the layout contract, but the sandbox has no browser egress, so it wants a look on the preview before it's trusted. The other two are structural and low-risk.

## Verification

- `npm run test:layout-contract` — holds, no new violations (relevant here, since the chat change adds a scroll region)
- `npx eslint` + `npx prettier` — clean

---

### DIFF
--- a/components/bots/bot-card.vue
+++ b/components/bots/bot-card.vue
@@ -279,14 +279,21 @@ const badges = computed<EntityCardChip[]>(() => {
   return result
 })
 
+/*
+ * NO THEME, NO DESIGNER. Silas, 2026-08-09: "I love the border around the
+ * gallery images, but we don't need the designer or theme text."
+ *
+ * Both were near-constant down the page -- the same handle on every card, and
+ * a palette name that the card's own `data-theme` border already SHOWS -- so
+ * they were a row of chips repeating what the eye had, on all 69 tiles. The
+ * theme still colours the frame; it just no longer also spells itself out.
+ *
+ * serverName survives because it varies and nothing else on the card carries
+ * it, but most Bots have none, so in practice the strip is now usually empty.
+ * Full provenance is on the card back, which is what the back is for.
+ */
 const metaChips = computed<EntityCardChip[]>(() => {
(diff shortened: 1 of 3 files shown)


silasfelinus:
REVIEWING: claude-quirky-curie-20260809T032917Z-review at 2026-08-09T03:29:17Z

---

PR #1639, opened 2026-08-09
Give Characters, Rewards and Scenarios the card back too

silasfelinus:
The remaining core objects that had one. Bots and Resources proved the shape; these three follow it exactly rather than each growing a variant — which is the whole reason `kr-card-back` is one component and not one per gallery.

| gallery | `#details` | interact |
| --- | --- | --- |
| Characters | class, genre, charm, empathy | **Chat** |
| Rewards | effect, rarity, type | **Encounter** |
| Scenarios | locations, inspirations, genres | **Play** |

Each also gets the id *as* the open state (so an open flag and a selected id can't drift apart) and a graceful panel for a record deleted while open.

## The dropdown branch is the load-bearing part

Every one of these galleries is also embedded as a **picker** — inside `bot-chat`, `character-chat`, `character-flip-card` — where a click has to pick and nothing else.

**Two of them didn't branch at all before this.** `openCharacter` and `openScenario` went straight to `selectCharacter` / `selectScenario` regardless of mode. That was fine while selecting *was* the whole behaviour — but now that picking and opening differ, the branch has to be explicit or the pickers break. Rewards already had it.

`verifyCardActionContract` still passes at **9 entity cards emitting `open` with none deciding its own consequence**, which is the check confirming this stayed in the gallery where it belongs.

## Badges, built the safe way

My first version filtered with an `entry is string` predicate. vue-tsc rejected it on Rewards: `rewardType` and `rarity` aren't plain strings, so the predicate wasn't assignable to its parameter.

All three now stringify first and filter on length, which doesn't care what the column's type is — the same shape works for an enum, a number, or null. Applied to all three rather than just the one that failed, since the others were only passing by luck of their column types.

## Dreams is deliberately not here

Its edit routes to the **dreammaker dashboard tab** rather than an inline form, so it needs a decision about where editing lives before it can take this frame — not a copy of it. Worth a conversation rather than a guess.

## Verification

- `npm run test:route-gallery-contract` — passes, 0 holdouts, 0 shadow browsers
- `npm run test:card-action-contract` — holds at 9 cards
- `npx eslint` — clean; the one remaining error in `character-gallery` (`characterGenres` unused) is pre-existing, confirmed by stashing and re-running

---

### DIFF
--- a/components/characters/character-gallery.vue
+++ b/components/characters/character-gallery.vue
@@ -35,6 +35,59 @@
       />
     </section>
 
+    <kr-card-flip
+      v-model="infoCharacterOpen"
+      :label="infoCharacter?.name || 'Character'"
+    >
+      <template #back="{ close, commit }">
+        <kr-card-back
+          v-if="infoCharacter"
+          v-model:editing="infoCharacterEditing"
+          :title="infoCharacter.name || 'Unnamed Character'"
+          :subtitle="infoCharacter.honorific || ''"
+          :description="infoCharacter.backstory || ''"
+          :art-src="infoCharacter.imagePath || ''"
+          :badges="infoCharacterBadges"
+          :can-interact="true"
+          interact-label="Chat"
+          @back="commit"
+          @interact="interactWithInfoCharacter"
+        >
+          <template #details>
+            <dl class="grid grid-cols-2 gap-2 text-xs">
+              <div v-if="infoCharacter.class">
+                <dt class="font-black uppercase opacity-55">Class</dt>
+                <dd>{{ infoCharacter.class }}</dd>
+              </div>
+              <div v-if="infoCharacter.genre">
+                <dt class="font-black uppercase opacity-55">Genre</dt>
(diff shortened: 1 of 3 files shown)


silasfelinus:
REVIEWING: 20260809T041422Z-auto-bc5f at 2026-08-09T04:14:22Z

silasfelinus (commented):
Technical review complete on exact head `21ebacdbcc27709451b7d337db68f8779a1045b6`. The diff is scoped to Character, Reward, and Scenario galleries; picker/dropdown clicks still select directly, while full galleries open the shared card-back tier and Interact continues through the existing destination. All 13 observed exact-head Actions workflows completed successfully, there are no unresolved review threads, and Vercel preview `dpl_EqdzSLXE1kUzUaswS9WCa8ehd1eZ` is READY and serves `/characters` with Nuxt public buildId matching this exact SHA. Responsive Layout Audit is intentionally a post-deployment workflow rather than a PR gate. No blocking defect found. (GitHub does not permit self-approval on this PR.)

PR #1654, opened 2026-08-09
Bundle Sharp libvips in Nitro Docker runtime

silasfelinus:
### Root cause
The Bookworm image correctly moved Sharp from `linuxmusl-x64` to `linux-x64`, but Nitro's server bundle includes Sharp's native binding while omitting the matching optional `@img/sharp-libvips-linux-x64` package. Sharp 0.34.5 then fails at startup looking for `libvips-cpp.so.8.17.3`.

### Fix
- Explicitly install Sharp 0.34.5 for Linux x64 during the Docker build, per Nuxt Image's cross-platform guidance.
- Assert the matching libvips package exists during build.
- Copy `@img/sharp-libvips-linux-x64` into the server node_modules resolution path alongside Nitro output.
- Keep the existing Bookworm base and `.env` runtime mount unchanged.

### Verification
CI covers the repository change. The decisive smoke test is rebuilding on Alexandria and starting the Unraid container; the build now fails early if the libvips package is absent instead of producing a crash-looping image.

### Stakes
Reversible Docker packaging change only. No schema, secrets, DNS, OAuth, or production-data mutation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,12 @@ ENV NUXT_TELEMETRY_DISABLED=1
 COPY . .
 
 RUN DATABASE_URL=mysql://kindrobots:build-only@127.0.0.1:3306/kindrobots npm ci --include=optional
+# Nuxt Image/IPX uses Sharp at runtime. Ask npm explicitly for the Linux x64
+# platform pair so both the native Sharp binding and its matching libvips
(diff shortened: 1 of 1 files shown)


silasfelinus:
REVIEWING: 20260809T102045Z-f6e8 at 2026-08-09T10:22:00Z

PR #1662, opened 2026-08-09
Serve mounted media through Nitro /images routes

silasfelinus:
### Root cause
Vercel makes legacy `/images/...` entity paths work by redirecting them to `https://media.acrocatranch.com/images/...`. The self-hosted Nitro server has no equivalent behavior. Its public-asset manifest is produced at build time, so files bind-mounted into `.output/public/images` at runtime are not discovered as static assets even though they exist on disk.

### Fix
- Add a catch-all Nitro route for `/images/**`.
- Resolve files from the existing `getImageStorageRoot()` / `IMAGES_PATH` contract.
- Stream runtime-mounted files with appropriate content type, size, modification time and one-hour cache headers.
- Reject path traversal and non-files.

This makes Alexandria's direct `:3009` path honor the same stable `/images/...` URL contract as Vercel without depending on Vercel's redirect layer.

### Reproduction
`/images/characters/bb-hope.webp` exists on `/mnt/user/pc/kindrobots/images` and returns 200 from `media.acrocatranch.com`, but 404s from Alexandria's Nitro server. Database-backed `/api/art/images/.../file` assets already work.

No schema, secrets, DNS, OAuth, or production-data changes.

### DIFF
--- a/server/routes/images/[...path].get.ts
+++ b/server/routes/images/[...path].get.ts
@@ -1,0 +1,69 @@
+import { createReadStream } from 'node:fs'
+import { stat } from 'node:fs/promises'
+import path from 'node:path'
+import {
+  createError,
+  defineEventHandler,
+  getRouterParam,
+  sendStream,
+  setResponseHeader,
+} from 'h3'
+import { getImageStorageRoot } from '~/server/utils/imageStorageRoot'
+
+const CONTENT_TYPES: Record<string, string> = {
+  '.avif': 'image/avif',
+  '.gif': 'image/gif',
+  '.jpeg': 'image/jpeg',
+  '.jpg': 'image/jpeg',
+  '.json': 'application/json; charset=utf-8',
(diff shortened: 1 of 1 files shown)


silasfelinus:
REVIEWING: 2026-08-09T121426Z-pr1662-rv-k7p4 at 2026-08-09T12:14:26Z

silasfelinus (commented):
Reviewed exact head db97506d60eb4193f69fd5bad392a316f4bd08e3. Initial CI exposed two branch-local defects: h3's Content-Length setter expects a number, and the initial relativePath assignment tripped the ESLint no-useless-assignment ratchet. Fixed both on the existing branch. The complete rerun is now green, including TypeScript #31312831017 and all 120 Contract verifier steps in #31312831028. Diff remains one scoped runtime /images route with traversal protection and no schema/data/config changes.

PR #1663, opened 2026-08-09
Stop editing a Bot from routing to chat; let the app remove the boot cover

silasfelinus:
Two follow-ups from testing on production.

## "edit just brings me to interact, not an edit screen"

`bot-interact.vue` is the router — `<bot-gallery v-if="!isInteractMode">` else `<bot-chat />`, with `isInteractMode = Boolean(currentBot) || sessionChats.length`. And `botStore.startEditingBot` sets `currentBot`; it has to, because `add-bot` in edit mode reads it for its guard, its reviews toggle, and its save target.

So opening the editor read as *"go and chat"*. It unmounted `bot-gallery`, and with it the card back the editor renders inside, landing the user in the chat surface. This hits **both** edit paths on `/bots` — the card back and the older in-gallery form — because both call `startEditingBot`.

`currentBot` was carrying two different questions: *which Bot am I talking to* and *which Bot am I editing*. `editingBotId` separates them, and the router ignores a selection that exists only to populate the form. It's cleared by `selectBot` (before its same-Bot early return), `deselectBot`, `startAddingBot`, and by the gallery when either editor closes — a flag left set would strand someone on the gallery unable to chat, which is this same bug pointing the other way.

## The boot cover outliving the app

Reported on `/bots`: the server boot cover still on screen after the app loaded, rendered **unstyled and in normal document flow** — its title in the page's own serif on the page's own background rather than white-on-black, and the app pushed down below it instead of covered by it.

**I could not reproduce or explain that**, and this change doesn't pretend to. The served HTML is correct — fetched from production, `<body>` is immediately followed by the inline `<style>` and then the cover, with `position: fixed`, `background: #000`, and a 6s release keyframe. There's no CSP, and nothing in `vercel.json` rewrites it. Why that block didn't take effect in that session is still open.

What this fixes is why the failure was **permanent**. Both of the cover's escapes — the release keyframe and the `html.kr-app-ready` rule — live in the same inline `<style>` block that positions it, and nothing ever removed the element: `markAppReady` only added a class. So a style block that doesn't take effect leaves a cover that isn't merely mis-positioned, it's unremovable.

`markAppReady` now also removes the node, and `app.vue`'s loader failsafe calls `retireBootCover` directly — that branch exists for the case where the app never reached `kind-loader`, so it can't rely on `kind-loader`'s call. Neither path depends on the CSS applying.

## Verification

- `verify-startup-handoff.mjs` — passes (the cover is still JS-free by default; removal is additive)
- `test:layout-contract` — holds
- `test:lint-ratchet` — 392 problems across 27 rules, unchanged
- `prettier --check` on every changed file — clean

Not verified in a browser: no browser egress in this session. The routing fix is traced through the call chain (`startEditingBot` → `currentBot` → `isInteractMode` → `v-if`) rather than watched.

---

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -220,6 +220,7 @@ import { useNavStore } from '@/stores/navStore'
 import { usePageStore } from '@/stores/pageStore'
 import { useUserStore } from '@/stores/userStore'
 import { useIntroStore } from '@/stores/introStore'
+import { retireBootCover } from '@/utils/startupLaunch'
 
 const pageStore = usePageStore()
 
@@ -291,6 +292,15 @@ if (import.meta.client) {
   failsafeTimeoutId = setTimeout(() => {
     showLoader.value = false
     failsafeTimeoutId = null
+
+    /*
+     * The server cover normally retires on its own 6s keyframe, or early via
+     * markAppReady() once kind-loader runs. Neither is guaranteed here -- this
+     * branch exists precisely for the case where the app never reached
+     * kind-loader, and the keyframe shares an inline <style> block that has
+     * been observed not to apply. Removing the node needs neither.
+     */
+    retireBootCover()
   }, 9000)
 }
 
--- a/components/bots/bot-gallery.vue
+++ b/components/bots/bot-gallery.vue
@@ -934,7 +934,18 @@ const canEditInfoBot = computed(() => {
  * how you save changes onto the wrong Bot.
  */
 watch(infoEditing, async (isEditing) => {
-  if (!isEditing || !infoBotId.value) return
+  /*
+   * Leaving edit mode releases the routing flag `startEditingBot` set. Without
+   * this, a user who opened the editor from an in-progress chat would be held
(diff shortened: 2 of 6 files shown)


silasfelinus:
REVIEWING: 20260809T131504Z-auto-97krv5 at 2026-08-09T13:15:04Z

silasfelinus (commented):
Reviewed exact head 71b5e48b27873f70b143161cfbeed0015ac04003. The routing fix correctly separates edit state from interaction state, and the boot-cover change adds a JS removal path plus layout-critical inline fallback without making visibility/animation unbeatable. No unresolved review threads. All 15 observed exact-head workflows completed successfully, including TypeScript, Contract Tests, Startup Handoff, Layout, MDC Scroll Ownership, Facet Catalog, API Client, Entity Art Manager, Taskmaster, Narrative contracts, and Storybook contracts. The comment volume is heavier than AGENTS.md's preference, but it is explanatory rather than a correctness blocker for this production-regression repair.

PR #1676, opened 2026-08-09
Stop /api/art/collection/folders serializing N manifest fetches

silasfelinus:
Top runtime error in production on 2026-08-09: **17 hits of `Vercel Runtime Timeout Error: Task timed out after 40 seconds`** on `/api/art/collection/folders` in a three-hour window. The work was never 40 seconds of anything — it was N round trips queued end to end.

Two causes, both in `server/utils/folderCollections.ts`:

**1. The resolve loop was serial.** `listFolderCollections` awaited `resolveFolderImages` for each slug before starting the next, so the endpoint's wall-clock was the *sum* of every manifest fetch. Now batched at 10 concurrent.

Batched rather than a single `Promise.all` over every slug: the CDN path is a network fetch per folder, and an unbounded fan-out on a large site trades a timeout for socket exhaustion. 10 matches the `REQUEUE_CONCURRENCY` idiom already in `art/queue/reenqueue-failed.post.ts`. Order is preserved — slugs arrive sorted from `listFolderSlugs`, batches are consumed in order, and `Promise.all` keeps each batch's results in index order.

**2. The master index was re-read once per slug.** `imagesFromManifest` calls `readCollectionsIndex` for every slug it resolves, so the same `collections.json` was fetched N times per request — on top of the N gallery manifests actually needed. Now memoised per origin for 60s.

Three details in that cache worth flagging:

- It stores the **promise**, not the value, so the ten concurrent resolvers in a batch share one in-flight fetch instead of all missing an empty cache and racing.
- A **failed** read is not cached. `readCollectionsIndex` returns `{}` so callers can degrade, and pinning that for a minute would turn one bad fetch into a minute of empty collections.
- A **rejected** promise is evicted, for the same reason — it would otherwise be replayed to every caller for the full TTL. (The uncached reader swallows its own I/O errors, so that's a backstop rather than the expected path.)

No behaviour change on the happy path: same slugs, same order, same URLs.

## Verification

- `test:lint-ratchet` — 395 problems across 27 rules, unchanged
- `eslint` + `prettier --check` on the changed file — clean

Not verified against a live folder set: this session has no access to the image host, so the speedup is reasoned from the call structure (N serial round trips → ⌈N/10⌉ batches, and 2N fetches → N+1) rather than measured. The correctness-relevant properties — ordering, and identical output — are structural.

## Not in this PR

`/api/art/collection/folders` was the *second* backend problem. The first is a 16-second interactive transaction (`P2028`) saturating the 2-connection pool; that one needs a schema decision and is written up separately.

---

### DIFF
--- a/server/utils/folderCollections.ts
+++ b/server/utils/folderCollections.ts
@@ -61,7 +61,8 @@ export function folderPathFromImageUrl(
 ): { slug: string; parentFolder: string | null } | null {
   if (!url) return null
   const parts = url.split('?')[0]?.split('/').filter(Boolean) ?? []
-  if (parts.length && /\.[a-z0-9]+$/i.test(parts[parts.length - 1] ?? '')) parts.pop()
+  if (parts.length && /\.[a-z0-9]+$/i.test(parts[parts.length - 1] ?? ''))
+    parts.pop()
   if (parts[0] === 'images') parts.shift() // drop the public URL prefix
   const slug = parts[parts.length - 1]
   if (!slug || NON_SLUG_DIRS.has(slug) || !SLUG_PATTERN.test(slug)) return null
@@ -70,7 +71,9 @@ export function folderPathFromImageUrl(
 }
 
 /** Just the slug from an image URL (see folderPathFromImageUrl). */
-export function folderSlugFromImageUrl(url: string | null | undefined): string | null {
+export function folderSlugFromImageUrl(
+  url: string | null | undefined,
+): string | null {
   return folderPathFromImageUrl(url)?.slug ?? null
 }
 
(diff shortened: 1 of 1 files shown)


silasfelinus:
REVIEWING: 2026-08-09T13:13:46-0700-auto-cd40b6 at 2026-08-09T20:20:30Z

silasfelinus (commented):
Technical review complete on exact head `4f4689a405b87320a8baa56d1823e0f5ed3eca5d`. The one-file change removes serial folder-manifest resolution by batching at 10, memoizes the shared collections index per origin with in-flight Promise sharing and failure eviction, and preserves output order. TypeScript and all triggered contract workflows are green. Contract Tests initially failed at the unrelated Academy starter-manifest fetch with a bare `fetch failed`; I reran that failed workflow without changing this branch, and the rerun completed successfully. No unresolved review threads.

PR #1681, opened 2026-08-09
Fix guest achievement records 401 on startup

silasfelinus:
### Task
kind-robots/direct: Avoid guest achievement-record 401s (kind: software)

- Resolve the user store before achievement sync so startup knows whether the visitor is authenticated.
- Keep the public achievement catalog refresh for guests.
- Fetch `/api/achievements/records` only after `userStore.isLoggedIn` is true.
- Migrate pending guest achievements only after authenticated records have loaded.
- Extend the achievement-store safety contract to lock in the auth gate and ordering.

### How I verified
- Inspected the protected records route and global startup call chain through the GitHub connector.
- Confirmed the branch is 2 commits ahead of current `main`, 0 behind, with only `components/admin/kind-loader.vue` and `utils/scripts/verifyAchievementStoreFetchSafety.ts` changed.
- Local checkout execution was unavailable because the sandbox could not resolve `github.com`; required repository checks are being left to GitHub CI and will be checked before merge.

### Stakes
reversible

### Flags for Reviewer
- The records endpoint remains authenticated; this fixes the caller rather than weakening endpoint security.

### Kaizen suggestion
Keep private startup fetches explicitly gated behind resolved identity state rather than bundling public and authenticated fetches under one generic `fetchRemote` flag.

### Notes for reviewer
This is startup orchestration only. Guest achievement discovery remains local/public, while signed-in users still refresh records before pending guest awards migrate.

### DIFF
--- a/components/admin/kind-loader.vue
+++ b/components/admin/kind-loader.vue
@@ -170,20 +170,38 @@ async function initializeStores() {
       import('@/stores/themeStore'),
     ])
 
+    const userStore = useUserStore()
+    const achievementStore = useAchievementStore()
+
+    await errorStore.handleError(
+      async () => userStore.initialize?.(),
+      ErrorType.STORE_ERROR,
+      'Error initializing user store',
+    )
+
     await runWave('identity + chrome', [
-      useUserStore().initialize?.(),
       usePageStore().initialize?.(),
       useNavStore().initialize?.(),
       useSmartbarStore().initialize?.(),
       useConsoleStore().initialize?.(),
-      useAchievementStore().initialize?.({
-        fetchRemote: true,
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: 2026-08-09T13:57:52-07:00-achievement401-7f3c at 2026-08-09T13:57:52-07:00

silasfelinus:
Review complete: diff is limited to startup orchestration plus its achievement hydration safety contract. All 14 PR workflow runs completed successfully, including TypeScript Type Check and Contract Tests; the targeted `Achievement hydration safety contract` step passed. `main` advanced by one already-merged commit touching only project-creation server files, with no overlap in this PR.

PR #1682, opened 2026-08-09
Index ArtJob's enqueue dedup instead of scanning LongText payloads

silasfelinus:
Option B of the three we weighed. `POST /api/art/queue` deduplicated by attempt fingerprint with

```js
payload: { contains: '"attemptFingerprint":"…"' }
```

which compiles to `payload LIKE '%…%'` over a **LongText** column. No index can serve that. The `userId` index narrowed to one user; MariaDB then read every LongText payload that user had ever produced — inside an interactive transaction holding one of only **two** pool connections per lambda.

Two things made it degrade rather than merely be slow:

- the status filter includes `DONE`, so the scan covered all history and grew monotonically with every completed job;
- `GET_LOCK(name, 3)` sat in front of it, so contention added up to 3 more seconds before the scan even started.

Production on 2026-08-09:

```
P2028: The timeout for this transaction was 10000 ms,
however 16096 ms passed since the start of the transaction.
```

alongside pool errors reading `active=2 idle=0 limit=2` on ~24 API routes — the **saturation** signature, not the can't-connect one we saw the night before.

## The change

`ArtJob.attemptFingerprint` mirrors the payload value into an indexed `VARCHAR(64)`, and the lookup uses it. Semantics unchanged: same filter, same ordering, and the migration backfills the column from the payload for every pre-existing row.

**Every create site that can carry a fingerprint sets it — not just the one that mints it.** `/api/art/queue/[id]/reenqueue` and `/api/art/queue/repair-weak-prompts` both derive their payload from an existing job and therefore *inherit* its fingerprint. Under the old LIKE scan those copies were dedup targets automatically; a column only the minting endpoint filled would have silently stopped matching them. The other six create sites build fresh payloads with no fingerprint at all, so their column stays NULL exactly as their payload always lacked the needle.

## Migration notes

The backfill uses nested `SUBSTRING_INDEX`, not `REGEXP_SUBSTR` or `JSON_EXTRACT`. The 2026-08-06 per-slot migration died precisely because MariaDB's `REGEXP_SUBSTR` returns an empty string where MySQL returns NULL, which then fails under strict mode. Both DDL statements are `IF NOT EXISTS` for the reason that same migration recorded — DDL is not transactional here, so a failure mid-migration leaves real changes behind and a re-run has to be a no-op on them.

⚠️ **The backfill is a one-time full scan of `ArtJob`.** It is the same work being removed from the request path, paid once at deploy rather than on every enqueue — but on a large table expect `prisma migrate deploy` to sit there for a while. Rows left NULL simply don't participate in dedup, which is the same outcome as a payload that never carried a fingerprint.

## Verification

- `prisma validate` / `prisma format` / `prisma generate` — clean
- `verifyArtJobProvenance` — passes
- `test:no-prisma-json-cast`, `test:artjob-bulk-scope`, `test:artjob-lora-override` — pass
- `test:lint-ratchet` — unchanged
- `eslint` on every changed file — clean

Not verified against a live database: this session can't reach one, so the index's effect is reasoned from the query shape rather than measured. The `ProxySQL stats_mysql_query_digest` check is what would confirm the `LIKE` disappears from the top of `max_time`.

## Follow-up (option C, agreed)

Promote the index to `@@unique([userId, attemptFingerprint])`, drop `GET_LOCK` entirely, and catch `P2002` on insert. That removes the advisory lock and its 3s wait too — but it changes how duplicates are *prevented* rather than how they're *found*, so it wants its own change once this column is proven populated in production.

---


silasfelinus:
REVIEWING: 20260809T135951-0700-pr1682-rv-5f2c at 2026-08-09T20:59:51Z

silasfelinus:
Technical review on exact head `ab3e03b6225fdb7ce9b7b3ee5ec6124d51bc910d`: **not safe to merge yet**.

1. **Scope regression:** this PR's diff includes `components/resources/resource-card.vue`, `components/resources/resource-gallery.vue`, and `utils/scripts/route-gallery-baseline.json`, but the PR is an ArtJob enqueue/index fix. The Resource files are stale relative to current `main` and would roll back newer card-back/title/action work already present there. Rebase onto current main and drop all unrelated Resource/gallery history from this PR.

2. **Migration gate:** `20260809210000_art_job_attempt_fingerprint/migration.sql` contains a full-table `UPDATE ArtJob ...` backfill. `kind_robots/AGENTS.md` permits agent auto-merge only for additive migration statements; data rewrites remain human-gated. Either redesign this slice so the migration is additive-only while preserving the intended dedup behavior, or leave the backfill behind an explicit human gate.

The indexed-column direction itself is sensible and removes the LongText `LIKE` query shape, but this head should not merge as-is.

PR #1685, opened 2026-08-09
Make guest achievement record reads return 200

silasfelinus:
## Closed as unnecessary
The reported post-fix reproduction came from an Unraid container that had been restarted/force-updated after `git pull` but had **not rebuilt the Docker image from the updated checkout**. That means it was still serving the previous compiled Nuxt bundle, so the 401 did not demonstrate that the merged client-side auth gating failed.

Keeping `GET /api/achievements/records` authenticated is preferable. The original merged fix resolves user identity before achievement record sync and avoids the guest request at the caller.

No merge performed.

### DIFF
--- a/server/api/achievements/records.get.ts
+++ b/server/api/achievements/records.get.ts
@@ -2,11 +2,24 @@
 import { defineEventHandler } from 'h3'
 import prisma from '../../utils/prisma'
 import { errorHandler } from '../../utils/error'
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: 2026-08-09T14:12:04-07:00-achievement401-followup-4a7e at 2026-08-09T14:12:04-07:00

PR #1688, opened 2026-08-09
Block clickthrough during startup explore mode

silasfelinus:
## What changed
- Make the fullscreen startup effect stage capture pointer input while `Pause & explore` is active.
- Keep normal startup behavior unchanged outside immersive mode.
- Add a startup-contract regression assertion requiring immersive input capture.

## Root cause
The paused animation stayed visually fullscreen but `.startup-animation__stage` retained `pointer-events: none`, so clicks landed on the already-loaded application underneath, including external newsfeed links.

## User impact
Paused startup animations can still respond to mouse/click input, but the hidden loaded page can no longer receive those clicks.

## Validation
- Compared branch against `main`: only `components/screenfx/startup-animation.vue` and `utils/scripts/verify-startup-handoff.mjs` changed.
- Awaiting repository CI before merge.

### DIFF
--- a/components/screenfx/startup-animation.vue
+++ b/components/screenfx/startup-animation.vue
@@ -318,6 +318,7 @@ onBeforeUnmount(() => {
  */
 .startup-animation__stage--immersive {
   background: #000;
+  pointer-events: auto;
   opacity: 1;
 }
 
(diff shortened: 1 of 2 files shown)


silasfelinus:
REVIEWING: 2026-08-09T214900Z-startup-clickthrough-a7c2 at 2026-08-09T21:49:00Z

PR #1689, opened 2026-08-09
Let the card back fall back through the same art chain as the card front

silasfelinus:
Silas, 2026-08-10: *"only about half the bots are showing images when selecting … as long as one exists (and it does because we clicked on it to select) we have something to show."*

Exactly right, and the asymmetry was the bug.

The card **front** renders through `kr-art-plate` with `:source="record"`, which resolves `imagePath → cardPath → heroPath → iconPath`. Every gallery then handed `kr-card-back` **one field** — `avatarImage` for Bots, `imagePath` for Characters, Dreams, Rewards and Scenarios. So a Bot with a `cardPath` and no `avatarImage` had art on the tile and a blank plate on the back. That's why it looked like roughly half of them: it's whichever half happen to store their art in the field the back was given.

`kr-card-back` now takes the record itself and resolves through `resolveEntityArtwork` — the same order the front walks. `artSrc` stays as the explicit first choice, so Bots still prefer `avatarImage` and only fall through when it's empty. **The back can no longer show less than the front did.**

Five galleries wired. `resource-gallery` already passed a four-field chain of its own (`ArtImage.thumbnailPath → ArtImage.imagePath → previewImageUrl → imagePath`) and needed nothing — which is what made the inconsistency easy to miss.

## Verification

- `test:layout-contract`, `test:gallery-consistency`, `verifyCardActionContract` — pass
- `test:lint-ratchet` — 392 problems across 27 rules, unchanged
- `eslint` + `prettier` on changed files — clean (the 4 eslint errors reported in `character-gallery` and `dream-gallery` are pre-existing and baselined; they're in lines this PR doesn't touch)

Not verified in a browser — no egress here. The fix is structural: the back now calls the same resolver as the front, so any record the grid could draw, the back can draw.

## Not in this PR

The missing edit button is still open — the card back's Edit and its `can-edit` wiring are byte-identical to when it last worked, so it needs a different explanation than this one.

---

### DIFF
--- a/components/bots/bot-gallery.vue
+++ b/components/bots/bot-gallery.vue
@@ -109,6 +109,7 @@
           :title="infoBot.name || 'Unnamed Bot'"
           :subtitle="infoBot.subtitle || ''"
           :description="infoBot.description || infoBot.personality || ''"
+          :source="infoBot"
           :art-src="infoBot.avatarImage || ''"
           :badges="infoBotBadges"
           :can-edit="canEditInfoBot"
--- a/components/characters/character-gallery.vue
+++ b/components/characters/character-gallery.vue
@@ -46,6 +46,7 @@
           :title="infoCharacter.name || 'Unnamed Character'"
           :subtitle="infoCharacter.honorific || ''"
           :description="infoCharacter.backstory || ''"
+          :source="infoCharacter"
           :art-src="infoCharacter.imagePath || ''"
           :badges="infoCharacterBadges"
           :can-interact="true"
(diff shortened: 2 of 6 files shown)


silasfelinus:
REVIEWING: 20260809T151328-0700-auto-5a4944 at 2026-08-09T22:15:10Z

silasfelinus (commented):
Reviewed exact head 6c274034c4bd525ec3a3b986cca3c95dc28d5eb7. The change is scoped to kr-card-back plus the five entity galleries that feed it. `resolvedArtSrc` preserves explicit `artSrc` precedence, then falls through `resolveEntityArtwork(source)` using the existing canonical entity-art chain. No schema/data/security/outward-facing changes, no unresolved threads, preview is READY, and all 13 exact-head Actions workflows completed successfully. Safe to merge.

PR #1702, opened 2026-08-10
Offer a card theme in the Model Builder, and persist whatever it picks

silasfelinus:
Silas, 2026-08-10: *"we should add theme choice to our model builder options, when the user manually creates new stuffs."*

Character, Bot, Reward, Dream, Scenario and Facet each gain a `theme` field whose choices are the 35 stock daisyUI themes.

Deliberately **without a `default`** — leaving it blank is what tells the commit route to pick at random. A default here would quietly make one theme the house pick for everything built by hand.

## One trap worth naming

**The commit route cannot use `pickChoice` for this.** That helper upper-cases before testing membership, which is correct for the enum-shaped fields around it (`RARITY`, `REWARD_TYPES`, `FACET_TAXONOMIES`) and wrong here — daisyUI theme names are lowercase, so every one of them would have failed the test and the field would have silently done nothing. `pickTheme()` matches case-correctly.

Absent or unrecognised input becomes a random valid theme rather than nothing: *"It's more important to populate with something than to make a precise match."* `coerceCardTheme` is the single place that decides this, so the model builder and any later creation path can't drift apart.

`randomCardTheme` is genuinely random, unlike `resolveEntityTheme`'s fallback — it runs once and the answer is written to the row, so there's nothing to keep stable, and the id-derived rule is unavailable at creation anyway because the row has no id yet.

## Not in this PR

Defaulting the theme inside the five REST create endpoints (`POST /api/{characters,dreams,rewards,scenarios,facets}`). Each assembles its create input differently — a `buildCharacterCreateInput` helper, an inline `dataInput`, and three more shapes — so it's five separate edits rather than one.

Worth knowing this is about **storability, not appearance**: a row created without a theme still renders correctly, because `resolveEntityTheme` derives a stable one from its id. It just isn't a value you can see or edit.

## Verification

All by exit code, not by absence of matching output:

- `npm run test` (vue-tsc) — `exit=0`
- `test:lint-ratchet`, `test:layout-contract`, `test:component-reachability` — `exit=0`
- `eslint`, `prettier --check` on changed files — `exit=0`

Not verified in a browser — no egress here.

---

### DIFF
--- a/server/api/model-builder/items/[id]/commit.post.ts
+++ b/server/api/model-builder/items/[id]/commit.post.ts
@@ -1,6 +1,7 @@
 // /server/api/model-builder/items/[id]/commit.post.ts
 // Execute the approved COMMIT for a build item as an idempotent durable write.
 import { createError, defineEventHandler, readBody } from 'h3'
+import { coerceCardTheme } from '~/utils/entityTheme'
 import type {
   DreamType,
   Rarity,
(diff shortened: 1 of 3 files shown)


silasfelinus:
REVIEWING: 20260809T192749-0700-review-1702-r9k4 at 2026-08-10T02:27:49Z

PR #1706, opened 2026-08-10
storybook/t-014: add "Start a story with this" links on Facet and Reward detail panes

silasfelinus:
## What

`storybook-page.vue`'s `?scenario=`/`?character=`/`?facet=`/`?reward=`/`?location=` query-param
seeding (shipped 2026-08-06) had no entry points anywhere in the app — the consuming side landed
without any links, so the feature was unreachable by a user. This adds the two entry points Silas
called out as priority:

> "what we should do is have links to these elements that start the process to tell a storybook
> story with that element as part of it" — explicitly instead of giving Facets and Rewards their
> own chat/conversation interactions, since both are tools that feed other interactions rather
> than things a visitor talks to.

- `components/facets/facet-interact.vue`: a "Start a story with this" button in the detail
  header, navigating to `/storybook?facet=<slug>`.
- `components/rewards/reward-encounter.vue`: the same button in the interact header, next to the
  rarity badge, navigating to `/storybook?reward=<slug>`.

Both use `navigateTo({ path: '/storybook', query: {...} })`, the existing cross-page CTA
convention (matches `image-card.vue`'s `makeColoringPage()`). `storybook-page.vue`'s
`seedFromQuery()` adds the slug to the setup draft on top of whatever draft is already restored
from localStorage and then strips the query param, so this can never destroy a story already in
progress and a reload can't re-add a removed ingredient.

Reward already has its own narrative-encounter engine (`startRewardStory()`) — left untouched,
since that predates Silas's stated preference and removing it is out of scope for this task.

## Out of scope (left for a follow-up if wanted)

Character and Scenario detail panes already have their own chat interactions (consistent with the
same by-design split: things you talk to vs. tools that feed other interactions) and were left
alone. Symmetrical Storybook links there would be an easy follow-up if Silas wants them, but
weren't called out as priority.

## Verification

- `npx eslint` — clean on both changed files
- `npx prettier --check` — clean
- `npm run test` (vue-tsc --noEmit) — exit 0
- `npm run test:facet-gallery` — pass
- `npm run test:component-reachability` — pass, 0 unreachable

---
_Generated by [Claude Code](https://claude.ai/code/session_01S1xfy3D7ic8WotSsgxxenP)_


silasfelinus:
REVIEWING: 20260810T001649-0700-auto-f17955 at 2026-08-10T07:16:49Z

silasfelinus (commented):
Reviewed exact head `000f87f5e940bb4355e9125dccc5027360beac98`. I found and fixed two real contract regressions on the existing branch: (1) the extracted Facet profile inherited a viewport-breakpoint grid that the layout ratchet correctly rejected; (2) reverting the extraction to dodge that made `facet-interact` grow, which the route-gallery contract correctly rejected. Final shape keeps the router thin, moves the working surface into `facet-profile.vue`, uses a container-width-responsive auto-fit grid there, and leaves the layout baseline untouched. Final diff is scoped to the Facet/Reward Storybook entry points plus the necessary Facet activity-surface extraction. All 13 exact-head workflows completed successfully, including Layout Contract #31366156627, Contract Tests #31366156525, and TypeScript #31366156531. No unresolved review threads or human gates in this PR.

PR #1710, opened 2026-08-10
Stage 3: stack the header supplementals, icon the mode bar, hide unpertinent options

silasfelinus:
Follow-up to #1707, from the 2026-08-10 /bots screenshot.

## 1. Notifications, refresh and server become a two-row stacked cluster

> "they are supplementals, login-manager is the one worthy of being full sized. That will free up some space."

Applied at **every** width rather than only where the row is tight, because the reason given was hierarchy, not screen size — half-height controls beside a full-size avatar say which is which without a label.

`grid-rows-2 grid-flow-col` rather than a fixed column count. The membership isn't fixed: `maturity-toggle` appears only for a logged-in user on a dashboard, and `notification-bell` carries a cart button that appears only when the cart has items. Auto-flow opens a new column every two items with no per-case class; a fixed `grid-cols-2` would reflow the wrong way as membership changed.

The readouts lost their `lg:flex-row` for the same reason — two adjacent clusters disagreeing about whether they stack is the hodgepodge this stage exists to stop. At 1024px the old rule drew a two-row cluster beside a one-row one for no reason a reader could see.

**Measured on /themes**, control strip → tab strip:

| width | control strip | tab strip |
|---|---|---|
| 390px | 180 → **162** | 96 → 96 *(already on its floor)* |
| 700px | 277 → **215** | 113 → **175** |
| 1024px | 338 → **215** | 376 → **575** |
| 1366px | 472 → **338** | 623 → **757** |

Header height unchanged at 77px.

One bug worth recording: at `xl` the pre-existing `.header-control-strip .btn` rule **ties on specificity** with the new one and wins on source order, so the cluster rendered at 2.75rem and pushed a 90px box through a 77px header. The readout rules escape this only because they happen to name `.karma-widget` as well. Fixed by doubling the parent selector rather than by moving the block, since order is exactly what the existing comment says not to rely on.

## 2. Cards/Heroes/Icons are glyphs below lg

> "Cards heroes icons should be icons on sm and md displays"

Glyph below `lg`, glyph + word from `lg`. The single letter shown below `sm` ("C H I") was the weakest part — it tells a first-time visitor nothing, and a shape says what a letter cannot in the same space. `title`/`aria-label` carry the word at every size.

The three icons live in `GALLERY_MODES` beside `label` and `abbr`, so one vocabulary still owns the mode. **Monochrome deliberately**: most of `assets/icons` hardcodes its fills, and a fixed-palette glyph on the active `btn-primary` reads as a sticker sitting on the button. Verified in the DOM that all three resolve to `currentColor` mask spans and invert with the active state.

Buttons measure 42px below `lg`, 83–92px above.

## 3. Clear is absent, not greyed

Fixed in all five places it appeared: `bot-gallery`, `reward-gallery`, `collection-picker`, `stylist-mask-brush`, `dream-brainstorm`.

## 4. The automation

New `verifyDisabledAffordances.ts`, wired into the **Contract verifiers** job.

> "If we could figure out how to automate this kind of fix, it would be great."

**Narrow on purpose**, which is what lets it be a hard gate rather than another advisory ratchet. There were **128** `:disabled="!..."` bindings in the repo and the overwhelming majority are correct — Send disabled until you type, Save until the form is valid, pager arrows at the ends. Those *teach* something, and hiding them would leave people hunting for a control that ought to be there. A contract that flagged all of them would be wrong 120 times to be right 5, and would be silenced within a week.

So the rule fires only on the **conjunction**: a clear/reset/remove-labelled control **and** a pure-negation condition. That's the shape with no subject, where the disabled state reports an absence the empty content already shows. `dream-brainstorm.vue` has both cases side by side — "Clear" is flagged, "Reject Pending" is not, because candidates can exist with none pending.

**The detector self-tests on every invocation**, against 3 must-flag and 6 must-pass fixtures, not behind a `--self-test` flag. This contract passes when it finds nothing, so every way it could break — a regex typo, a Vue syntax it stops recognising — produces a confident green. It can't be trusted to report zero unless it's first proven able to report one. I also confirmed it end to end by reverting one fix and watching it exit 1, then restoring it and watching it exit 0.

## Verification

By exit code, all green: `test:layout-contract`, `test:gallery-consistency`, `test:gallery-adoption`, `test:route-gallery-contract`, `test:lint-ratchet`, `test:component-reachability`, `test:disabled-affordances`, `audit:wonderlab-previews --strict`, `prettier --check` (changed files), `vue-tsc`.

Browser-measured at 390/700/1024/1366 — **after waiting out the boot cover**. An earlier pass in this session measured through it and read the loader's layout as the header's, which is where the nonsense 54px/90px numbers came from.

Caveat, same as #1707: no database in the sandbox, so /bots and /rewards fall into their error states and the two Clear buttons there weren't exercised live. The change is `:disabled="!x"` → `v-if="x"` and the new contract covers the pattern, but the runtime behaviour on those two routes is unobserved.

https://claude.ai/code/session_01YbBrskaWwa4eGxxu3R6Zgh

---
_Generated by [Claude Code](https://claude.ai/code/session_01YbBrskaWwa4eGxxu3R6Zgh)_


silasfelinus:
REVIEWING: 20260810T101511Z-review-837dc1 at 2026-08-10T10:18:30Z

PR #1717, opened 2026-08-10
Make Brainstorm revision history restorable and branches traceable

silasfelinus:
## What changed

This is Brainstorm `t-009`: per-candidate revision history and non-destructive branching.

### Revision history
- Candidate edits, regenerations, and generated/branched states already had revision records; this makes that history visible and useful.
- Candidate cards now expose a collapsible `History · N versions` panel with version number, reason, response lens, timestamp, title, and text.
- Older versions can be restored in one click.
- Restore is intentionally non-destructive: the selected old text becomes the current candidate and a **new `restored` revision is appended**, so later work remains in history.
- Candidate identity, batch membership, keep/reject state, and feedback are preserved while moving between text revisions.

### Exact branch lineage
- A branch no longer records only a live `parentId`.
- At branch creation, the child snapshots the exact parent revision index/title/text that was used as the generation reference.
- Candidate cards expose a collapsible branch-lineage panel showing that immutable origin.
- If the live parent still exists, the branch can jump to it; if the parent is later deleted, the origin snapshot still explains where the branch came from.
- Parent edits after branching cannot rewrite history retroactively.

### Contract / preview coverage
- Adds `restored` to the typed revision-reason contract and a typed `BrainstormBranchOrigin` snapshot.
- Revision records retain response-lens metadata where available.
- Store contract now requires restore semantics and exact branch-origin snapshots.
- Workbench contract now requires history UI, restore wiring, stable candidate anchors, and branch-lineage UI.
- WonderLab's Brainstorm candidate specimen now carries a deterministic branch origin and three revisions so the history surface remains previewable without a live LLM.

## Scope

Brainstorm session/workbench only. No durable database persistence, schema migration, generation endpoint change, prompt change, art generation, or unrelated cleanup.

## Flags for Reviewer

Please scrutinize two invariants in particular:
1. restoring an old revision must append `reason: 'restored'` rather than truncating/replacing the revisions array;
2. branch provenance must snapshot the parent version used at branch time rather than reading the parent's later mutable state.

This is user-facing UI on a `worker/*` branch, so there is no Vercel preview by repository policy. Merge only after TypeScript, Brainstorm Store/Workbench contracts, layout/contract checks, and the attached workflow set are green; then verify `/brainstorm` against the resulting production deployment.


silasfelinus:
REVIEWING: 2026-08-10T115600Z-brainstorm-t-009-e4b9 at 2026-08-10T12:13:00Z

PR #1723, opened 2026-08-10
fix(model-builder): openRun() must not swap state.run when reopening the already-active run

silasfelinus:
### Task
conductor/model-builder/t-029: Polish and upgrade Model Builder front-end surface (recurring). This cycle's follow-on found a real, previously-unfixed bug.

- `stores/modelBuilderStore.ts`: `openRun(runId)` now short-circuits when the requested run is already `state.run`'s id, instead of unconditionally swapping in a freshly-adapted object.
- New regression guard `utils/scripts/verifyModelBuilderOpenRunIdentityGuard.ts` + self-test, following the repo's existing `verifyModelBuilder*Guard.ts` pattern, wired into `package.json` and `.github/workflows/contract-tests.yml`.

### The bug
`openRun(runId)` always replaced `state.run` with a *different* object, even when the requested run was already active. `fetchRuns()` (called by the run-history panel on mount) and `openRun`'s own fallback fetch both rebuild `BuildRun`/`BuildItem` objects via `adaptRun`/`adaptItem`, which reset the client-only `artJobId`/`queueState` fields to `null` (the server never returns them). `pollAsyncArtJob` captures a reference to the specific `BuildItem` object it was called with and polls against that object only, independent of whatever object `state.run` currently points to.

**Concrete failure:** user queues an async generate on an item → opens the run-history panel (silently refreshes `state.runs` with fresh objects) → re-opens the same run they already had open. `openRun` swapped in one of those fresh objects, orphaning the in-flight poll. The UI now shows the item as idle (no queued/rendering indicator, buttons re-enabled) while the original job is still running, inviting a second concurrent generation, with the eventual first result landing invisibly on the orphaned object.

### How I verified
- `npx vue-tsc --noEmit` clean (0 errors) with the Prisma client freshly regenerated against schema. (Note: the *committed* `prisma/generated/` snapshot is stale relative to schema — missing the Brainstorm models added in #1718 — causing pre-existing TS errors in `server/api/brainstorm/**`; confirmed present on `origin/main` before this change too, via `git stash`, and left untouched as out of scope.)
- `eslint`/`prettier` clean on all changed/new files (`stores/modelBuilderStore.ts`'s 2 pre-existing `no-empty` errors and its pre-existing prettier drift confirmed present on `origin/main` before this change, unrelated to my diff).
- All 19 `test:model-builder-*` contract npm scripts pass, including the two new ones (self-test + contract for the new guard).

### Stakes
reversible

### Flags for Reviewer
- Spotted but did not touch: the committed `prisma/generated/` snapshot is out of sync with `prisma/brainstorm.prisma` (missing `BrainstormSession`/`BrainstormCandidate` client types), causing `vue-tsc` errors in `server/api/brainstorm/**` whenever the generated client isn't freshly regenerated first. Pre-existing, unrelated to this task — worth a follow-up task if `npm run test` (which doesn't force a fresh `prisma generate`) is relied on as a pass/fail signal anywhere.

### Kaizen suggestion
File a task to regenerate and commit a fresh `prisma/generated/` snapshot (or stop committing it at all, since `postinstall`/`nuxi prepare` already regenerates it on every `npm ci`) so a local `vue-tsc` run without an explicit `prisma generate` step doesn't show stale false-positive errors unrelated to whatever's actually being changed.

### Notes for reviewer
Conductor task `model-builder/t-029` claimed under session `claude-conductor-agentrun-20260810T153049Z`; will be flipped to review/done in the conductor repo once this merges.

---
_Generated by [Claude Code](https://claude.ai/code/session_013cH3zDHAQBf6gHYFbexPiu)_

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -328,6 +328,12 @@ jobs:
       - name: Model Builder run-writable guard contract
         run: npm run test:model-builder-run-writable-guard
 
+      - name: Model Builder open-run identity guard checker self-test
+        run: npm run test:model-builder-open-run-identity-guard-selftest
+
+      - name: Model Builder open-run identity guard contract
+        run: npm run test:model-builder-open-run-identity-guard
+
       - name: Kontext mask forwarding contract
         run: npm run test:kontext-mask-forwarding
 
--- a/package.json
+++ b/package.json
@@ -161,6 +161,8 @@
     "test:model-builder-item-patch-stage-guard": "tsx utils/scripts/verifyModelBuilderItemPatchStageGuard.ts",
     "test:model-builder-run-writable-guard-selftest": "tsx utils/scripts/verifyModelBuilderRunWritableGuard.test.ts",
     "test:model-builder-run-writable-guard": "tsx utils/scripts/verifyModelBuilderRunWritableGuard.ts",
+    "test:model-builder-open-run-identity-guard-selftest": "tsx utils/scripts/verifyModelBuilderOpenRunIdentityGuard.test.ts",
+    "test:model-builder-open-run-identity-guard": "tsx utils/scripts/verifyModelBuilderOpenRunIdentityGuard.ts",
     "test:kontext-mask-forwarding": "tsx utils/scripts/verifyKontextMaskForwarding.ts",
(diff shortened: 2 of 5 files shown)


silasfelinus:
REVIEWING: 20260810T161307Z-review1723-q8n4 at 2026-08-10T16:13:07Z

PR #1742, opened 2026-08-11
Header polish, and root-cause the wrong tab highlight

silasfelinus:
Your four notes on the redesign, plus the defect #1739 shipped with documented-but-unfixed, plus the Prisma thing.

## Your notes

**1. Hub moves right of the tab selector.** The row now reads Channel → Tab → You → Tutorial.

Its panel had to flip to **right-anchored**: a left-anchored 20rem panel hanging off a button two controls in from the right edge ran straight past the viewport — clipped mid-word at 760px. Caught it in a screenshot; my first comment claiming left-anchoring still worked was wrong.

**2 + 3. One identity row.** Avatar, name and role on the left; **server and refresh over karma and mana** on the right.

That collapses three separate blocks — identity header, a "Wallet" section, and a "Session" section that opened with server and refresh — into the band you look at first, and takes about a third off the panel's height. Still the karma and mana *widgets*, not their numbers reprinted, so the transaction list and refill countdown come along.

**4. The unread badge escapes the circle.** The button carries `overflow-hidden rounded-full` to crop the avatar, and that crop was shaving the badge hung off its corner. As a sibling positioned against `.account-hub` it overlaps the circle instead — measured overhanging **4px on both axes**. `pointer-events-none` keeps the avatar clickable through it; the count still reaches screen readers via the button's `aria-label`.

## The wrong tab highlight — fixed, and it was never a rendering bug

#1739 left this as unexplained: the list marked *Dreams* on /bots, and instrumentation showed one element computing `isActiveTab` **false** while carrying `active`. Impossible inside a single render, so it read as a Vue patching anomaly.

**It came from the server.** `app.vue` renders `workspace-header` above `<NuxtPage>`, so during SSR the header resolved its channel and tab from a `pageStore` that `pages/[...slug].vue` hadn't filled yet. It fell back to defaults and emitted **the HOME channel's four tabs, Dashboard active, on a Play route** — confirmed in the SSR HTML. Hydration then reused those elements for the real thirteen-tab list, patched their text and attributes, and left the server's `active` sitting on the first row. Always position one, which is exactly what made it look positional rather than like a data problem.

The header now resolves **both channel and tab from `route.path`**, which is identical on both sides, so the two renders agree and there's nothing left behind. `pageStore` stays the fallback — it carries frontmatter a path can't express, and off-channel routes resolve to null.

Hydration warnings on /bots and /characters drop **15 → 9** (the remainder are other components').

## Prisma

You asked; here's the permanent fix. The Brainstorm PRs added `BrainstormSession`/`BrainstormCandidate` to the schema and to server routes but never committed the regenerated client — and `postinstall` was only `nuxi prepare`, so a fresh clone typechecks with six phantom errors in `server/api/brainstorm/**`.

- committed client regenerated
- **`postinstall` now runs `prisma generate` first**, so it can't rot again on the next model

CI and the Vercel build already ran `prisma generate`, which is why this was invisible there and only bit people locally.

## Verification

**Every `run:` command in the Contract verifiers job — 122 commands, 0 failures** — rather than the hand-picked subset that let a contract slip past me last time. Plus `vue-tsc`, `prettier`, `eslint`.

In a browser: header order, identity-row contents, the panel fitting at 760px, the badge overhanging the circle, and the highlight tracking the route on /bots, /characters and /dreams with zero page errors.

**Not verified**: no database here, so karma/mana/notifications render their logged-out state — the identity row's second line is unobserved with real values.

https://claude.ai/code/session_01YbBrskaWwa4eGxxu3R6Zgh

---
_Generated by [Claude Code](https://claude.ai/code/session_01YbBrskaWwa4eGxxu3R6Zgh)_


silasfelinus:
REVIEWING: 20260811T051325Z-conductor-12747b at 2026-08-11T05:14:10Z

silasfelinus:
Reviewer fix pushed on the existing branch: the new install-time `prisma generate` required `DATABASE_URL` via `prisma.config.ts`, so dependency-only CI failed during `npm ci` before tests ran. I restored database-free `postinstall`; the regenerated Prisma client remains committed, which fixes the fresh-clone type surface without making every install require DB configuration. Holding merge for the new exact-head checks.

Issue #1761, opened 2026-08-11
Retire WonderLab and migrate its 706 voice pieces into first-party reviews


silasfelinus:
Silas-directed 2026-08-11 product decision: WonderLab is retired.

## Kill
- `/plan/wonderlab` museum and legacy `/wonderlab` entry point
- componentStore and global `/api/components` boot path
- build-time WonderLab component metadata generation
- WonderLab-only component preview/reconcile/status/admin review surfaces
- CI/contracts whose sole purpose is keeping museum components previewable
- animation-manager coupling that persists attempts as Component museum rows

Do not delete historical Component/ReviewDraft/Reaction production rows yet. They are migration source data.

## Keep and improve
The polished WonderLab corpus contains 706 canonically ordered Bot/Character-authored entries. Treat them as 353 paired voice studies. Do **not** relink their existing Component reactions to unrelated objects. For each pair, use the old text as evidence for voice/chemistry and write a fresh target-appropriate two-speaker scene on a live reviewable Kind Robots object, with setup/payoff and clearly distinct speakers.

The general review system becomes the permanent home:
- review UI/component naming leaves `wonderlab/`
- first-party reviews preserve Bot/Character speaker identity and link to the speaker
- all main reviewable gallery models expose `allowReviews` on selected-item/edit surfaces
- user reactions remain user reactions

## Data/publication gate
Infrastructure retirement is approved now. Newly generated replacement prose remains draft-only until Silas approves publication because it is outward-facing content.

## End state
`components/abandonware` is no longer compiled for museum previews and can later be archived outside this repo entirely. WonderLab-specific code and build machinery have zero live product dependency.

silasfelinus:
Audit note: the schema already has `allowReviews` on exactly seven main model types: Bot, Character, Dream, Project, Facet, Reward, and Scenario. First-party authorship is also already supported on Reaction via the 2026-07-19 Bot/Character author expansion; the runtime projection contract resolves those authors separately from the logged-in publishing user. That means the replacement system does **not** need a new authorship schema. The main structural work is (1) generalize the WonderLab-only draft generator/repository targeting, (2) expose `allowReviews` consistently on selected gallery backs/edit surfaces, and (3) build a generic authored-review feed instead of the component-specific museum feed.

silasfelinus:
[...]

PR #1766, opened 2026-08-11
Detach Animation Manager and retire Component runtime store

silasfelinus:
Continues #1761 after WonderLab's route/build retirement.

This removes the last legitimate product dependency keeping the retired Component museum runtime alive:

- Animation Manager now reads the canonical animation catalog and live animationStore only
- removes Component-backed build-history, ratings, statuses, comparisons, promotion/polish/retire actions
- deletes componentStore and its dynamic store loader
- deletes the animation Component-ledger helper
- adds a contract scanning live application code so componentStore/useComponentStore/loadComponentStore cannot return outside abandonware/WonderLab residue

Animation preview/toggle behavior remains. Development history belongs in source control and Conductor, not duplicated Component rows.

No Component database rows are deleted in this slice. Historical voice/review data remains untouched.


silasfelinus:
REVIEWING: 20260811T111305Z-a7c93d at 2026-08-11T11:13:05Z

PR #1762, opened 2026-08-11
Retire WonderLab from boot and build-time component discovery

silasfelinus:
Closes part of #1761.

Silas retired WonderLab as a product/developer subsystem on 2026-08-11. This first incision removes the highest-cost live coupling without touching the historical review corpus:

- stop running the WonderLab component-manifest generator every time `nuxt.config.ts` evaluates
- stop globally importing/initializing `componentStore`, eliminating ordinary-startup `/api/components`
- remove the canonical `/plan/wonderlab` content route and its Plan/Museum tab
- redirect the old `/wonderlab` bookmark to `/plan` instead of preserving the museum
- delete the dedicated WonderLab public-acceptance workflow/spec/navigation verifier
- remove WonderLab-only reconciliation, preview, registry, museum-query, review-rollout and voice-audit steps from the required Contract Tests gate
- retain the generic first-party Reaction author + ReviewDraft contracts because those are the foundation for moving the 706 voices onto real model objects
- record the preservation/migration contract for the 706 authored voice entries

No Component, ReviewDraft, or Reaction production data is deleted. The archived voice-polish manifests remain source material. Replacement prose is a later draft-generation pass and remains a publication human gate.

The next deletion slice removes the now-unreachable WonderLab/admin/component machinery, then decouples Animation Manager from Component rows so `componentStore` and `/api/components` can disappear completely.

PR #1772, opened 2026-08-11
Retire SmartBar and trim global startup work

silasfelinus:
Continues the startup/detritus cleanup after #1762 and #1766.

This slice removes the retired SmartBar/SmartIcon frontend/runtime surface and aggressively narrows global boot:

- navStore no longer imports SmartIcon seeds, SmartIcon Prisma types, SmartBar, icon favorites/model filters, or WonderLab folder state
- theme-icon no longer depends on SmartBar edit mode
- deletes smartbarStore, /icons content route, SmartIcon gallery/editor/card, SmartIcon API routes, SmartIcon Cypress coverage, seed JSON, and generator script
- kind-loader deliberately preserves ConsoleStore, including its browser-console startup journey
- kind-loader stops globally initializing feature stores that should initialize from their own surfaces (servers/checkpoints/art/bot/chat/character/prompt/reaction/reward/scenario/weird/choice/random/builders)
- the global randomizer no longer fetches the 1,000-facet catalog during ordinary startup
- theme boot applies local/current theme without fetching the shared theme catalog globally

The Prisma SmartIcon model/user columns are intentionally left dormant in this slice; deleting database schema is not required to remove runtime/build overhead and can be handled separately if worthwhile.

This PR is intentionally using TypeScript/contracts and preview verification as dependency detectors. If a feature surface genuinely depended on speculative global initialization, fix that surface to initialize its own store rather than restoring universal boot.

ConsoleStore is explicitly NOT detritus. Its startup console messages are an intentional dev-facing feature and remain globally initialized.


silasfelinus:
REVIEWING: 20260811T121436Z-review2081-a7c9 at 2026-08-11T12:16:00Z

PR #1773, opened 2026-08-11
Keep PWA install cache to stable app chrome

silasfelinus:
Continues the startup/build-budget cleanup after #1772.

The first production build after WonderLab + SmartBar retirement still generated a PWA precache of **314 entries / 8060.20 KiB**. That is mostly because Workbox globbed every built JS/CSS chunk, including route-local feature code, even though Kind Robots does not claim full offline page support: pages remain SSR/API-driven and `navigateFallback` is intentionally null.

This slice makes the installability contract match the product contract:

- keep the manifest and generated service worker
- precache only stable install chrome (192/512 app icons, Apple touch icon, favicon)
- stop precaching every lazy JS/CSS/image/font asset during install/update
- add a regression verifier that rejects JS/CSS precache globs unless we explicitly redesign Kind Robots as a full offline app

Expected result: PWA install/update traffic drops from ~8 MiB / 314 entries to a tiny stable icon set, while route-local bundles continue loading normally when their feature is visited.

No visible UI behavior should change. Verification should use exact-head CI plus the Vercel preview/build log's actual Workbox `precache` count/size before merge.

### DIFF
--- a/.github/workflows/pwa-precache-budget.yml
+++ b/.github/workflows/pwa-precache-budget.yml
@@ -1,0 +1,26 @@
+name: PWA Precache Budget
+
+on:
+  pull_request:
+    paths:
+      - 'nuxt.config.ts'
+      - 'utils/scripts/verifyPwaPrecacheBudget.ts'
+      - '.github/workflows/pwa-precache-budget.yml'
+  push:
+    branches: [main]
+    paths:
+      - 'nuxt.config.ts'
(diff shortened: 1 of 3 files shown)


silasfelinus:
REVIEWING: 20260811T124024Z-conductor-sweep-abcd at 2026-08-11T12:40:24Z

---

Issue #1769, opened 2026-08-11
Rebuild object comments as fresh Resource / Reward / Facet conversations


silasfelinus:
## Purpose

Split the comment rewrite into its own project/thread. This is **not** WonderLab restoration and it is **not** a migration of WonderLab pairings. The old corpus is useful as character-voice evidence, but the new system should create fresh conversations around real Kind Robots objects.

This issue deliberately captures the decisions reached during the broader site-retirement work so a new agent can take this on without needing the original sandbox/thread context.

## Current state / work already landed

The larger cleanup started while tracing stale startup behavior, SmartIcon loading, and WonderLab/component detritus. During that work we separated the useful idea behind WonderLab comments from the museum architecture that happened to contain them.

Already merged:

- **#1762** — WonderLab retirement stage one. Removed the museum route/navigation, stopped ordinary startup/Nuxt preparation from initializing/scanning the Component museum, removed Component-grounded AI review generation, and removed other WonderLab runtime/build machinery.
- **#1765** — comment-generation foundation. This established the first hard architectural boundaries for the replacement system:
  - target types for this pass are **Resource, Reward, Facet only**;
  - Dreams, Scenarios, Bots, and Characters are intentionally out of scope;
  - casting is object-first rather than historical-pair-first;
  - historical WonderLab pairings have **zero casting weight**;
  - archived comments may inform a character's voice, but must not be copied/paraphrased as new text;
  - calibration generation is publication-free and does not write production comments.

A separate cleanup thread is continuing to remove the remaining WonderLab/Component runtime and general startup/site detritus. **Do not take ownership of that cleanup from this issue.**

## Product goal

Give Resources, Rewards, and Facets lively in-world commentary from appropriate Kind Robots characters/bots. The object is the star. Characters should react because they have a plausible relationship, affinity, expertise, personality fit, or useful contrast with the object, not because two rows happened to be paired in an old museum dataset.

The result should feel authored and characterful rather than like generic AI product reviews. Comments can disagree, joke, misunderstand, admire, criticize, riff, or reveal personality. We want conversational texture, not an encyclopedia wearing googly eyes.

## Important distinction: preserve voices, discard pairings

There are roughly 700 historical WonderLab comment/review entries. Treat them as a **voice corpus** only.

Useful evidence from the old corpus can include:

- diction and vocabulary;
- sentence length/rhythm;
- recurring comedic instincts;
- optimism/cynicism/intensity;
- technical vs emotional framing;
[...]

PR #1795, opened 2026-08-12
Stress-test approved object-comment voices with GPT-authored batch 002

silasfelinus:
Follow-up human gate for #1769 after Silas approved batch 001's voice direction and explicitly deferred Resources.

## What this is

`config/comment-calibration-batch-002.json` is a **publication-free GPT-5.6 Sol stress test**, not a hand-polished showcase. The prose was authored in the approval session with `humanPolish: NONE_BEFORE_REVIEW`.

This batch deliberately tests the two holes from the first calibration:

- **repeatability:** Maestro Clank (RICH), Marrow (THIN), and The Time-Office Agent (SPARSE) each appear twice on unrelated targets;
- **range:** comment lengths run from **9 to 75 words**, including genuine micro-reactions rather than making every character deliver a polished 30-50 word miniature essay.

It also keeps weak/noisy casting visible with explicit contrast directives instead of pretending lexical affinity solved those cases.

## Samples worth judging hardest

**Maestro Clank, Melodic Note, 9 words**
> Pianissimo. One note is enough. Do not improve it.

**Maestro Clank, Public Notary, 75 words**
> The stamp descends on the page in perfect common time: one blow, one witness, no encore. They ask you to say nothing, which is sensible; testimony ruins the acoustics. I would place the document on the second stand, not the first. The first is for scores that want to be heard. This one wants only the downbeat acknowledged. Mm. Raise the seal. Again. No, softer. The room is listening even if the record is not.

**The Time-Office Agent, Common, 9 words**
> The backward watch stayed quiet. I checked twice. Good.

That last one is the hardest SPARSE test: second appearance, unrelated low-signal Facet, and deliberately **no numbered-attempt tic**.

**Marrow, Always Online**
> Eleven tabs is too many pots on one stove. You stop tasting any of them. Close three. The world will still be there after the water boils.

## Contract added

`verifyCommentCalibrationBatch002.ts` is separate from the approved batch-001 verifier. It enforces:

- model/session provenance and no human polish before review;
- Reward + Facet only, with Resources still deferred;
- score recomputation through the existing ranker;
- archive draft ownership/tiering after the 706-row dedupe;
- the same banned review vocabulary and 8-word no-paraphrase check as batch 001;
- repeated speakers spanning RICH / THIN / SPARSE;
- novelty decay from actual prior appearances;
- at least one <=12-word micro-reaction and one >=60-word turn;
- multiple explicit contrast directives.

The Comment Migration workflow now runs both calibration contracts.

## Human gate

Please judge specifically:

1. Do Clank, Marrow, and the Time-Office Agent stay recognizably themselves across unrelated objects without merely repeating a tic?
2. Does the 9-to-75-word range feel healthier than batch 001's polished middle band?
3. Does the Time-Office Agent's second 9-word line still read as the same SPARSE-evidence character without the numbered-attempt construction?
4. Is the invisible house author less audible, or are too many lines still performing the same clever landing?
5. Which individual lines, if any, should be rejected or regenerated?

**Do not merge this PR on green checks alone.** This remains the subjective calibration gate before scale. No database write, persistence migration, or publication path is included.


silasfelinus:
Calibration stress batch is technically ready for the human gate.

Exact head: `b0162f0763ed0f274368921386529e7ff8fa4d39`

All 15 attached workflows completed successfully, including:
- Comment Migration Contract (batch 001 + batch 002 contracts)
- TypeScript Type Check
- aggregate Contract Tests
- Layout / Facet / API Client / Storybook / Narrative contracts

The first batch-002 CI run exposed only a verifier bookkeeping bug: metadata explicitly recorded `TRIO: 0` while the computed shape map omitted zero-count buckets. I fixed the verifier to count all four shapes explicitly; the prose/score/archive rules were not weakened and the batch itself was unchanged.

This PR remains intentionally **unmerged**. Human judgment is the remaining gate.

PR #1803, opened 2026-08-12
Run the approved Reward and Facet comment backfill

silasfelinus:
Production execution PR for #1769 after Silas approved and merged calibration batch 003.

Initial commit adds only a **read-only, preview-only status probe**. The route hard-fails unless `VERCEL_ENV=preview` and `VERCEL_GIT_COMMIT_REF=gpt/comment-backfill-live`; it cannot run from production.

Plan in this PR:
1. verify the preview has the expected production DB corpus and OpenAI runtime credential without exposing either secret;
2. add bounded/idempotent generation + persistence on the same preview-only surface;
3. generate only public+active Rewards and content-bearing public+active Facets; Resources remain deferred;
4. preserve truthful Bot/Character authorship through `authorBotId` / `authorCharacterId`;
5. never overwrite human reactions;
6. run in small batches and verify counts/visibility after each batch;
7. remove the temporary trigger before this PR is merged.

Silas explicitly authorized production generation/publication in chat on 2026-08-11 after merging #1802.


silasfelinus:
REVIEWING: 20260811T211207-0700-review1803-4f9c at 2026-08-12T04:12:07Z

silasfelinus (commented):
Current head is safe as a read-only preview probe, and I verified the exact preview deployment returns the intended production corpus snapshot: 259 public+active Rewards, 1,591 public+active Facets, all currently reviewable, zero Reward/Facet reactions, with DB and OpenAI runtime configuration present. Vercel Preview Authentication is active for this deployment.

One important guardrail for the next commit: **do not put generation or persistence into this `.get.ts` handler.** The PR plan currently says to add bounded/idempotent generation+persistence “on the same preview-only surface”; if that means mutating on GET, crawlers/prefetch/retries/cache revalidation can trigger production writes. Keep this GET endpoint probe-only. Put any write step behind a distinct POST handler, preserve the preview+branch guard, require the existing authenticated/admin execution boundary (or equivalent explicit execution token already approved for this one-shot flow), and return `Cache-Control: no-store` on the mutating response. Also preserve the stated idempotency/no-human-overwrite checks server-side, not only in the caller.

No merge recommendation yet because the PR is intentionally draft and the production-write implementation has not landed.

silasfelinus:
REVIEWING: 20260811T221445-0700-review1803-a7d2 at 2026-08-12T05:14:45Z

silasfelinus (commented):
Blocking production trigger on one concrete idempotency defect in `publishCommentBackfillLiveBatches.ts`.

`loadExistingFirstPartyState()` seeds `authoredText` with every already-published first-party Reward/Facet comment. The validation pass then calls `validateCommentFreshness()` for *every* payload item before the publish loop checks `existing.targetKeys`. On a retry after any partial successful run, the payload comment for an already-published target is compared against its own persisted text, `sharedShingle()` finds the exact repeated phrase, and the job fails before it can reach the intended `SKIPPED_EXISTING` behavior. That makes the production lane fail-closed but **not actually idempotent**, despite the PR contract and the real need to rerun safely after a mid-job interruption.

Please make retry validation distinguish the payload rows already present for a target from unrelated prior generated text. A safe shape would be: validate structural author/shape/order for every item, but only run freshness-against-existing-generated-text for targets that are not already in `existing.targetKeys`; alternatively prove exact persisted equality for skipped targets and exclude those rows from the duplicate-text corpus used for that item's validation. Keep the existing archive/canonical freshness checks.

This is caught before the ready marker exists, so no production write has fired. Once fixed, rerun the exact-head TypeScript/contracts and the one-shot publisher's dry/validation path before creating the production ready marker.

silasfelinus:
Second exact-head blocker from CI: Contract Tests #31565927233 failed specifically at **Workflow path references contract**. The new `comment-backfill-live.yml` watches `config/comment-backfill-live-ready.json`, but that path does not exist on the branch yet, so the repo's workflow-path contract correctly rejects the workflow. Please fix the trigger design rather than weakening the contract. A safe pattern is an existing sentinel/config file whose contents are explicitly checked for `ready: true`, or an intentionally confirmed `workflow_dispatch` lane; do not make ordinary draft-packet pushes capable of publishing. Current head `5bb3662416be356c1c7780bb13da9bad467598ba` remains unmergeable-by-policy until this and the idempotent-retry issue are fixed and exact-head checks rerun.

Issue #1757, opened 2026-08-11
Boot budget: retire SmartBar and stop globally initializing feature subsystems


silasfelinus:
## Why this exists

A production-loading audit on 2026-08-11 found that Kind Robots' global startup path is initializing a large fraction of the application regardless of which surface the visitor opens. SmartBar/SmartIcons exposed the pattern: the visible SmartBar no longer exists, but its store, API, legacy content route, user fields, nav compatibility state, and components are still wired into general startup.

This issue is intentionally broader than `/icons`: establish a small **shell boot contract** and make feature state route/surface-local.

## Verified production baseline

Current production build during the audit:

- largest client JS chunk: ~940.83 kB minified / ~308.38 kB gzip
- Vite warns about chunks over 500 kB
- Workbox precache: 617 entries / ~8664.58 KiB
- production `sw.js` confirms it precaches feature and parked/WonderLab artifacts including `icon-debugger.css`, `icon-header.css`, butterfly museum CSS, `chat-test.css`, and hundreds of lazy JS chunks

Vercel request-path counts over the audited 6h window included:

- `/api/server`: 5,408
- `/api/facets/catalog`: 3,259
- `/api/icons`: 2,033
- `/api/themes`: 1,806
- `/api/achievements/`: 1,752
- `/api/components`: 1,726

These are request counts, not unique page loads, but the code confirms these endpoints are reached from global startup paths.

### Dead SmartBar graph is still live

- `kind-loader.vue` initializes `smartbarStore` globally.
- `navStore` also initializes/consumes `smartbarStore` and still models obsolete navigation state as `SmartIcon` rows.
- `content/icons.md` remains a normal content root and mounts `icon-gallery`, which keeps the legacy icon component cluster reachable.
- `theme-icon.vue` still checks `smartbarStore.isEditing` even though SmartBar editing no longer exists.
- `/api/icons` currently returns only ~20 rows; response volume is not the main cost. The problem is keeping the obsolete feature graph and DB call alive.

### Global feature-store initialization

`kind-loader.vue` globally initializes feature state including component, randomizer, bot, chat, art, character, prompt, reaction, reward, scenario, weird, choice, checkpoint and builder infrastructure.

High-confidence examples:

- `componentStore.initialize()` fetches `/api/components`; its actual consumers are WonderLab/animation-manager.
[...]

PR #1810, opened 2026-08-12
Keep achievement catalogs lazy during global boot

silasfelinus:
## Root cause

The broader boot-budget cleanup already made most feature stores surface-local, but `kind-loader.vue` still forced `fetchAchievements(true)` and `fetchAchievementRecords(true)` for every mounted app. That meant the full remote achievement catalog and, for logged-in users, all achievement records were fetched even on routes that did not need them.

## What changed

- remove the unconditional global achievement catalog/record sync wave
- keep local achievement-store hydration during shell boot
- keep pending guest-achievement migration after login
- keep route-triggered achievement awards; the store already lazily fetches the catalog only when a route actually maps to an achievement code
- keep ID-based awards lazy through `fetchAchievementById`
- add a dedicated static boot-budget contract so the global loader cannot quietly reintroduce full catalog/record fetches

## Scope

This is one bounded slice of #1757, not the whole umbrella. Other boot-budget acceptance items remain open and should continue under that issue.

## Safety

No schema, production data, credentials, or outward-facing content changes. This removes remote reads from ordinary startup while preserving existing lazy award paths.

Progresses #1757


silasfelinus:
REVIEWING: claude/quirky-curie-uvx2jg at 2026-08-12T00:00:00Z

---

PR #1819, opened 2026-08-12
Fix approved comment corpus author-name drift

silasfelinus:
## Root cause

The approved comment backfill live run #31596152064 failed closed before publication because checked-in author labels no longer matched the live Character names. The publisher reported `REWARD:282: author name drift for CHARACTER:275: expected The Intermission, got Intermission.` The preceding retry had already identified CHARACTER:358 as `Grand Marshal Aurelia Vane` rather than the stale `Aurelia` label still on `main`.

## What changed

- update only the stored author display names in `config/comment-backfill-drafts/batch-0048.json`
- CHARACTER:358: `Aurelia` → `Grand Marshal Aurelia Vane` in both approved entries
- CHARACTER:275: `Intermission` → `The Intermission`
- approved comment prose, target IDs, author IDs, shapes, and release metadata are unchanged

## Verification

- compared branch against current `main`: one file, one commit, only the affected corpus lines
- live publisher validation itself supplied the canonical names and aborted before writes on mismatch
- no production write is triggered by this PR

## Stakes

reversible

## Flags for Reviewer

The separate Vercel account status is currently failing repository-wide with `Account is blocked`; do not bypass it if it remains required.

## Kaizen suggestion

Add a pre-publication CI check that validates all checked-in corpus author IDs/names against a generated canonical snapshot so name drift is caught before the production execution lane.

PR #1822, opened 2026-08-12
Unblock the comment backfill: all seven drifted authors, plus the pre-flight check

silasfelinus:
## Root cause

`publishCommentBackfillLiveBatches.ts` re-validates every speaker's display name against the live directory and **throws on the first mismatch**. That check is right — a comment signed by a name the character no longer uses is worse than no comment — but failing one-at-a-time costs a full production run per renamed speaker: the Tailscale lane, the DB connection, twenty minutes, and a human to read the log. Four runs have died this way, each surfacing exactly one name.

Nobody had diffed the whole corpus in one pass. I did, against `/api/characters` and `/api/bots`. **There are seven drifted speakers, ten entries, all in `batch-0048.json`:**

| Speaker | corpus | live | first failure | status |
|---|---|---|---|---|
| CHARACTER:358 | Aurelia | Grand Marshal Aurelia Vane | REWARD:281 | fixed by #1819 |
| CHARACTER:275 | Intermission | The Intermission | REWARD:282 | fixed by #1819 |
| CHARACTER:364 | The Big Bad Who Lost the Bet | The Big Bad Who Lost a Bet | REWARD:284 | **this PR** |
| CHARACTER:277 | Editor | The Editor | REWARD:285 | **this PR** |
| CHARACTER:214 | The Sketchbook That Draws Back | The Sentient Sketchbook | REWARD:290 | **this PR** |
| CHARACTER:192 | Fizzwick Fairy | Fizzwick the Fairy | REWARD:302 | **this PR** |
| BOT:398 | AMI Butterfly | AMI | REWARD:303 | **this PR** |

#1819 merged while this branch's CI was running, fixing the first two. Without the remaining five, the next production run gets past `REWARD:282` and dies at `REWARD:284`, then 285 → 290 → 302 → 303 — four more attempts after that, one per name.

## What changed

**1. The five remaining author names — `config/comment-backfill-drafts/batch-0048.json`.**
Only `speakers[].name` values. Comment prose, target ids, author ids, shapes and release metadata are untouched, exactly as #1819 did for its two. Verified by word-diff against `main`.

**2. The pre-flight check #1819's own kaizen note asks for — `utils/scripts/verifyCommentBackfillAuthors.ts`.**
Fixing the names without fixing the process just defers the next four failures. This reads every `batch-*.json`, fetches the live directory, and reports **every** drift and every missing speaker at once, grouped by speaker, before exiting non-zero. Read-only public HTTP — no database, no secrets, about ten seconds.

Wired two ways: `npm run test:comment-authors` for local use, and as the first substantive step of the `publish` job in `comment-backfill-live.yml` — after `npm ci` (it needs `tsx`) but **before Prisma, before Tailscale, before any DB connection**, so a drifted corpus fails in seconds instead of twenty minutes into the production lane.

Deliberately *not* folded into `npm run test:comments`: those contracts are offline and pure, and this one would take the whole suite red whenever kindrobots.org is down.

One known difference from the publisher, documented in the header: the publisher reads the directory from the DB filtered on `isActive`, this reads the public API filtered on `isPublic` — the price of needing no credentials. They agree on every speaker in the corpus today (201 authors, zero absent), and a name is a name either way. What it cannot distinguish is public-but-inactive, so an absent speaker is reported as a failure rather than skipped.

**3. Component residue from the schema retirement.**
`PUBLIC_TARGETS` held `'component'` alone. Component stopped being a `KarmaRefType` when the model was dropped in `20260812040500_retire_component_schema`, so that branch has been unreachable — dead code still reading as live visibility policy. Also corrected the per-target read route's header, which claimed component keeps its own route; it no longer exists.

## Verification

- `npm run test:comment-authors` → zero drift across **491 comments / 201 distinct speakers / 18 batch files**.
- Proven to fail correctly, not just to pass: reverting two names and pointing a third at a bogus name reports **all three at once**, exit 1. Pointing a speaker at a nonexistent id triggers the absent-speaker branch. File restored after both.
- `npm run test:comments` (migration, signals, calibration) ✅
- `npm run test:reaction-targets` — 17 categories mapped, 12 readable target types ✅
- `npm run test:reaction-route-auth` — 7 handlers, no anonymous table reads, no body-supplied userId ✅
- `npm run test:lint-ratchet` — 340 problems across 24 rules (−52), no rule got worse ✅
- `npx vue-tsc --noEmit` — clean, exit 0 ✅
- All re-run after merging `main`; 21/21 GitHub checks green on the merge commit.

## Stakes

Reversible. **No production write is triggered by this PR** — the validator is read-only, and the gate marker on `main` stays `ready:false`. The live lane only opens on a push to `gpt/comment-backfill-live` that flips `comment-backfill-live-ready.json`.

## Flags for reviewer

- **This does not make the lane publishable on its own.** 402 of 658 eligible targets are still undrafted and Facets have zero drafts; `assembleCommentBackfillDrafts.ts` hard-fails at 256/658. The curation scripts that produce the rest live only on `gpt/comment-backfill-live`, which is the right owner for them. This PR removes the name-drift failure mode; it does not fill the corpus.
- The 402 undrafted targets are curated at run time on that branch and cannot be checked from `main`, so the validator's clean result covers the 256 committed here.
- The Vercel `Account is blocked` status is still red, and #1823 did **not** clear it — I expected it to and was wrong. `vercel.json` now sets `git.deploymentEnabled: false`, but the block is evaluated at the account level by the Vercel GitHub App before it reads repo config, so it cannot be cleared from inside the repository. It is red repo-wide and was red on #1819, which merged anyway.
- The census artifact is stale — it records 257 Rewards and `allowReviews` false everywhere, while the publisher's live query now returns 658 eligible targets. Worth a re-run as a separate change.
- `server/utils/commentBackfillGeneration.ts` exports `publishManualCommentBackfillSlice` and `runCommentBackfillSlice`, which nothing calls. Generate-and-write code with no caller is worth deleting or gating.

Refs #1769.


silasfelinus:
#1819 merged while CI was running, which conflicted this branch on the two lines the two PRs both touch. Merged `main` in and resolved.

The conflict was exactly the two remaining name fixes and nothing else — I diffed the conflicting lines character by character before resolving, and the only differences were `the` → `a` (CHARACTER:364) and the inserted `The ` (CHARACTER:277). No prose on either side had moved, so taking this branch's version was a clean superset rather than a judgement call.

**What's left in this PR after #1819 landed** is the five names it didn't cover, plus the validator and the Component cleanup:

| Speaker | corpus | live |
|---|---|---|
| CHARACTER:364 | The Big Bad Who Lost the Bet | The Big Bad Who Lost a Bet |
| CHARACTER:277 | Editor | The Editor |
| CHARACTER:214 | The Sketchbook That Draws Back | The Sentient Sketchbook |
| CHARACTER:192 | Fizzwick Fairy | Fizzwick the Fairy |
| BOT:398 | AMI Butterfly | AMI |

Without these, the next production run gets past `REWARD:282` and dies at `REWARD:284`, then 285 → 290 → 302 → 303.

Re-verified on the merge commit: `verifyCommentBackfillAuthors` reports zero drift across 491 comments / 201 speakers / 18 batches, and `test:comments`, `test:reaction-route-auth` and `test:reaction-targets` all pass. Word-diff against `main` confirms the draft file changes are still `speakers[].name` fields only.

Also worth noting for the earlier flag: #1823 disabled the retired Vercel Git deployments, so the repo-wide `Account is blocked` status that was red on the first run should no longer apply.

---

PR #1818, opened 2026-08-12
Stop globally initializing console progression

silasfelinus:
## Root cause

Issue #1757 identified `consoleStore` as feature state that was still being initialized by the global loader even though repository search shows no live consumer outside `kind-loader.vue`. Its initialization hydrates local progression/messages, emits a random console message, and increments XP for every app mount, which is feature behavior rather than shell boot responsibility.

## What changed

- remove `consoleStore` from `kind-loader.vue`'s global dynamic imports and startup wave
- update the loader contract comment so console progression is explicitly surface-local
- extend the existing boot-budget contract to reject reintroducing `useConsoleStore` or `consoleStore.initialize()` into global boot

## Scope

This is one bounded slice of #1757. It does not delete the console store or change its behavior when deliberately consumed later.

## Safety

No schema, production data, credentials, billing, or outward-facing content changes.

Progresses #1757

PR #1824, opened 2026-08-12
Keep random console joy without persistence

silasfelinus:
## Root cause

PR #1818 correctly removed console progression from the global boot path, but it also removed the one console behavior that is intentionally part of startup: the random console message.

Silas clarified that the console does not need to remember progression/history anymore, but the random startup message should remain because that is the feature's actual purpose.

## What changed

- restore one `logRandomMessage()` call during startup without calling `consoleStore.initialize()`
- remove console localStorage history/progression persistence and unused backend-sync scaffolding
- keep console state ephemeral for the current session
- update the boot-budget contract so it requires the startup random message while forbidding console persistence/global progression initialization

## Safety

No schema, database, credentials, billing, or outward-facing publication changes.

Follow-up correction to #1818. Progresses #1757.

### DIFF
--- a/components/admin/kind-loader.vue
+++ b/components/admin/kind-loader.vue
@@ -118,7 +118,8 @@ async function runWave(
 /*
  * Global boot is deliberately conservative. Feature stores initialize from
  * their own surfaces instead of making every visitor download/fetch the whole
- * application.
+ * application. The random console message is intentional shell behavior: one
+ * fresh bit of weirdness per startup, without restoring console persistence.
  */
 async function initializeStores() {
   try {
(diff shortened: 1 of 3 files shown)


silasfelinus:
REVIEWING: 20260812T211149Z-consolejoy-c3a9 at 2026-08-12T21:11:49Z

silasfelinus (commented):
Reviewed exact head 748c43b66491af8b5703836634db02ad350f87f7. Scope matches Silas's clarified intent: keep one random startup console message, remove persistence/global progression startup, and preserve ephemeral console behavior. I found no callers of the removed persistence/backend-sync methods and no review-thread blockers. All attached GitHub Actions are complete and green. Current kind_robots/AGENTS.md explicitly says Vercel is retired infrastructure and its status must not block review or merge decisions.

PR #1826, opened 2026-08-12
Make kindrobots.org the real host, and unblock the review corpus

silasfelinus:
Two threads: finish retiring Vercel from the code, and get the finished review comments out of the repo and onto the site.

## 1. Vercel is retired; the code did not know it

**`server/utils/email.ts` was inverted.** `DEFAULT_APP_BASE_URL` was the Vercel preview host and `https://kindrobots.org` sat in `LEGACY_APP_BASE_URLS`, so `normalizeAppBaseUrl` took the correct domain `nuxt.config.ts` supplies and rewrote it to a host that no longer serves the app.

That is every link sent to a real person — email verification, password reset, newsletter confirmation, the account link in a restriction notice — plus the redirects in `auth/email/verify.get.ts` and `newsletter/confirm.get.ts`, which call the same `appBaseUrl()`. Nothing throws when a link merely points somewhere wrong, so it stayed broken quietly. `verifyAppBaseUrl.ts` pins the polarity; proven by re-inverting the two constants and watching it fail.

**The container had no build identity.** `runtimeConfig.public.buildId` is baked at build time and `/api/version` reads `COMMIT_SHA` at request time; both came only from `VERCEL_GIT_COMMIT_SHA`, which nothing sets outside Vercel. So the deployed app called itself `development` and `/api/version` answered `commit: null` — which is exactly what `cypress.yml`'s deploy-wait polls for, so it could never pass and would burn its full timeout. Now a `COMMIT_SHA` build arg in both Dockerfile stages, filled from `github.sha`.

**Migrations stopped running anywhere.** `vercel-build.mjs` ran `prisma-migrate-deploy.mjs`; the Dockerfile runs plain `npm run build`, and nothing picked it up when Vercel was retired — so merging a migration to `main` stopped applying it.

> **Deviation from the plan, deliberate.** The plan said container entrypoint. That is wrong for this repo: `docs/runbooks/migration-credential-boundary.md` separates the application lane from schema-write capability, and `prisma-migrate-deploy.mjs` refuses `DATABASE_URL` outright. An entrypoint means the long-running app container permanently holds a migration credential — the exact thing that boundary prevents. It is a `workflow_dispatch` workflow whose lifetime is the migration instead, which is what the runbook itself prescribes. **This is the one commit that needs your explicit sign-off** (`731760e`); it is manual-fire, requires a literal confirmation string, and `migrate deploy` never resets or drops.

**The long tail.** `cleanup-test-users.yml` runs on a schedule against a dead host; `chatgpt/openapi.get.ts` served vercel.app as the API origin while the docs say kindrobots.org; the Google auth `process.env.VERCEL` branch was dead code. Plus `--base` defaults across operational scripts.

Left alone on purpose: `vercel.json` and `vercel-build.mjs` (five contract tests assert on the latter's text), the retired hosts inside the two contracts that exist to rewrite them, a note recording a real past incident, and the two Cypress lanes, which are manual-only and parked on a separate concern.

**Not fixable from the repo:** the red `Account is blocked.` status. #1823 did not clear it and neither does this — `vercel.json` already sets `git.deploymentEnabled: false`, but the block is evaluated at the account level by the Vercel GitHub App before it reads repo config. Uninstalling the app at the GitHub/Vercel level is the only fix, and that is a click for you.

## 2. The review corpus could not publish at all

491 finished comments were sitting in the repo, unreachable. Both the assembler and the publisher demanded an all-or-nothing 658/658 corpus, and only 256 targets were drafted — so nothing could go live until every remaining target had been written.

Publishing a **prefix** lands exactly the rows publishing everything would, because the run already skips any target carrying a first-party comment. What makes a partial payload dangerous is not its length but a **hole**: items are matched to targets by position, so a gap shifts every later comment onto the wrong object — rewards commented as facets — and nothing throws. Contiguity from zero rules that out and is still enforced per packet.

The subtle one: the write loop iterated every eligible target and did `payload[index]!`. Past the prefix that is `undefined` and the non-null assertion carries it straight into `item.speakers`. Both loops are now scoped to an explicit slice, and `verifyPartialPublishPrefix.ts` pins that — proven by reverting the write loop and watching the contract name the bug.

Also fixes provenance: the assembler hardcoded `authoringModel: 'GPT-5.6 Sol'`, but the corpus already carried a second drafting model, so the claim was untrue *before* this change. It now reports what the packets actually declare. `releaseGate` stays constant and both sides still gate on it — it records the gate, not a byline.

## 3. Facet casting was running blind

`relationshipScore` has a top tier for "the target Facet is one of this speaker's own facets" — 100, the strongest signal there is. It reads `speaker.facetIds`, which comes from join rows the public API does not expose. So **every speaker scored relationship 0 on every facet**, and facet casting ran on vocabulary overlap alone.

That matters because facet titles and character attributes are the same vocabulary by design. Brine's `alignment` is literally `Chaotic Good`; for facet #219 Brine placed third on accidental word overlap rather than first on *being* it.

`facetAttributeMatch` recovers the link from published attributes, so `relationshipScore` itself is untouched. Measured live: **74 of the 399 eligible facets** now have at least one speaker who carries them, via 766 links. Facet 219's top candidate goes 43.8 → 83.8.

The scoping is load-bearing. "Blunt" is a Personality facet and also a way to describe a weapon in a backstory; a false 100 outranks a true 90. Each group reads exactly one field, containment is whole-phrase, and object groups (colour, material) are unmatched. The contract covers three false positives that would each otherwise score 100.

## 4. An authoring harness, then the writing

470 comments cannot be hand-checked, and the publisher's prose rules only run inside the Tailscale-gated production lane, one violation per run.

- **`verifyCommentDraftQuality.ts`** runs those same rules offline over every batch file and reports everything at once. Clean on the existing 491; proven to catch a banned word, an under-length comment, a doubled speaker and a self-repeated phrase, all five at once.
- **`exportFacetDraftPlan.ts`** replaces the DB-bound plan exporter with the same output over the public API, reproducing the publisher's target list exactly — position is the contract. Novelty is seeded from the committed corpus so the speaker already carrying sixteen comments is not handed a seventeenth while ninety-five have never spoken.

**Written so far: 259/259 Rewards (complete) and the first 24 Facets — 283 of 658.**

The facet packet took two length passes. Everything came out at 45–50 words, over-written against the corpus median of 34 and wrong for the voices — Margin is documented as "soft short lines", Patient Zero as terse, and both were given paragraphs. Trimmed to median 40, with the genuinely loquacious speakers left long because that is their voice.

## Verification

- `test:comment-quality` — 525 comments across 20 batch files, median 34 (5–56), zero violations
- `test:comment-authors` — 206 distinct speakers, no name drift
- `test:comment-prefix`, `test:app-base-url`, `test:comments`, `test:comment-signals`, `test:reaction-route-auth`, `test:reaction-targets` ✅
- `test:lint-ratchet` — 339 across 24 rules (−53), no rule worse
- `npx vue-tsc --noEmit` — clean
- Assembler produces a valid partial payload at 283/658

**No production write.** The validator and exporter are read-only, and `comment-backfill-live-ready.json` stays `ready: false`. Publishing is still you flipping the marker on `gpt/comment-backfill-live`.

## Still open

- **375 facets remain undrafted** (~15 more packets). Resumable: each packet is independently valid.
- **Resources take no comments** — `TargetType` is `'REWARD' | 'FACET'`, the calibration contract rejects RESOURCE, and production exposes zero public Resource rows. The card back already has its Reviews button; there is nothing to attach.
- `dreams/index.post.ts:97`, `dreams/batch.post.ts:126` and `weirdHelper.ts:126` still hardcode `allowReviews ?? false`, contradicting the schema default.
- No `allow-reviews-toggle` exists for Dream, Facet or Resource.
- The census artifact is stale (257 Rewards, `allowReviews` false everywhere).

Refs #1769.

---


silasfelinus:
REVIEWING: 20260812T231142Z-review1826-a7f3 at 2026-08-12T23:11:42Z

silasfelinus:
Review note for `20260812T231142Z-review1826-a7f3`: I found one merge-blocking scope/gate issue. This PR includes `.github/workflows/migrate-production.yml`, while its own handoff explicitly says that production-migration capability needs Silas's sign-off. I’m separating that gated slice so the reversible host/comment fixes do not have to wait behind it. I am not approving or triggering any production migration or comment publication. #1769’s calibration/publication human gate remains intact.

silasfelinus:
Agreed, and thanks — that was my own flag and you're right that it shouldn't hold up the reversible work.

We both reached for it at the same time. I had reverted `731760e` locally for the same reason; when my push bounced I checked your `09cb487`/`663162c`, saw they do exactly the same thing and already contain my guard fix, so I **discarded my redundant revert and kept yours**. No duplicate removal in the history.

I've merged current `main` (`90f2739`, `00e6a34`) on top, since the branch had gone two behind while this was in flight. Re-verified on the reconciled tip:

- `test:comment-quality` — 580 comments across 22 batch files, median 34 (5–56)
- `test:comment-prefix`, `test:app-base-url`, `test:comments` ✅
- `test:capture-group-guards` against `origin/main` ✅ (this is what the earlier `Contract verifiers` failure was — a `.match()` result indexed inside an `if` block rather than guarded at the access; now `?.[0]`)
- `npx vue-tsc --noEmit` clean, lint ratchet −53 with no rule worse

**On the gated slice.** The workflow text is still recoverable at `731760e` in this branch's history if you want it reopened later. The reasoning that matters is in that commit message: it should be a `workflow_dispatch` job rather than a container entrypoint, because a long-running app container running `prisma migrate deploy` would hold a standing migration credential — the exact thing `docs/runbooks/migration-credential-boundary.md` exists to prevent. I'd rather that constraint be recorded somewhere than rediscovered.

Worth being explicit about what removing it means: **migrations still do not run automatically anywhere.** That gap predates this PR — it arrived when Vercel was retired and `vercel-build.mjs` stopped being the thing that ran them — and it is now simply untracked rather than fixed. Not a blocker for this PR, but it is a live hole in the deploy path.

To confirm your last line: nothing here publishes anything. `comment-backfill-live-ready.json` stays `ready: false`, the validator and plan exporter are read-only HTTP, and the corpus sits at 331/658 as committed drafts only. The calibration and publication gates are untouched.

---

PR #1827, opened 2026-08-12
Gate production schema migration workflow for explicit approval

silasfelinus:
### Scope
Isolated from #1826 so reversible host/comment fixes can merge independently.

This PR adds a manual-only production migration lane using `MIGRATION_DATABASE_URL` over Tailscale and documents the credential boundary. It does **not** run a migration by merging, and this session will not dispatch it.

### Human gate
The originating #1826 handoff explicitly requested Silas's sign-off for this capability. Leave this PR unmerged until that approval is given.

### Safety
- `workflow_dispatch` only
- requires literal `migrate` confirmation
- refuses ordinary `DATABASE_URL`
- uses the existing `prisma-migrate-deploy.mjs` wrapper
- no migration or production write has been triggered

Extracted unchanged from #1826 commit `731760e` / head `46655671c73785b7a519bb804e8cecda9a957c8e`.

### DIFF
--- a/.github/workflows/migrate-production.yml
+++ b/.github/workflows/migrate-production.yml
@@ -1,0 +1,100 @@
+name: Migrate Production Schema
+
+# Applies pending Prisma migrations to production.
+#
+# WHY THIS EXISTS
+# ---------------
+# Migrations used to ride the deploy: scripts/vercel-build.mjs ran
+# scripts/prisma-migrate-deploy.mjs as part of `vercel-build`. The container
+# path runs plain `npm run build` (Dockerfile), which does not, and no other
+# workflow picked it up when Vercel was retired -- so merging a migration to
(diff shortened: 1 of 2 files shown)


silasfelinus:
I checked this one as part of the two-PR CI investigation. Its Actions were not red: the exact head had successful TypeScript, Contract Tests, and Migration Credential Boundary runs. The non-mergeable state came from the branch falling behind/conflicting with current `main`, not a failing check.

It is also now superseded by merged #1831, which put the migration boundary on the actual container deploy path with a one-shot Compose `migrate` service, and #1832, which makes the host consume the GHCR image CI built. The current `main` migrate-on-deploy contract is green with those changes.

Because #1827 is explicitly a human-gated production-migration mechanism, I have left it open and unmerged rather than reviving a second, redundant migration path.

PR #1831, opened 2026-08-13
Migrate on deploy, and 24 more facet reviews

silasfelinus:
Two things: close the migration gap left when Vercel was retired, and keep writing the review corpus.

## Migrate on deploy

Migrations stopped running anywhere when Vercel went. `vercel-build.mjs` had been the thing that ran them; the container path runs plain `npm run build`. Merging a migration to `main` applied it to nothing.

**The blocker was the image, not policy.** The runtime stage copied only `.output`, `node_modules` and `package.json` — so the wrapper script, the schema and the migration SQL were all absent, and `prisma migrate deploy` could not run inside the deployed image at all. That is why migrations had to be run by hand from a separate checkout on the host, against whatever revision that checkout happened to be on. This is also a correction to advice I gave earlier in the session: I proposed a one-shot compose service using the same image before checking what the image contained, and it would not have worked.

The runtime stage now carries `prisma/`, `scripts/` and `prisma.config.ts` — migration SQL and a few `.mjs` files.

`docker-compose.yml` gains a one-shot `migrate` service running the wrapper, with the app declaring `depends_on: migrate: condition: service_completed_successfully`. `docker compose up -d` migrates first and refuses to start the app if it fails, so schema and code move together.

**This keeps the credential boundary rather than bending it.** The rule in `docs/runbooks/migration-credential-boundary.md` is that schema-write capability must not sit in the *long-running* application container; a container whose entire lifetime is the migration is the same shape as a CI job. `MIGRATION_DATABASE_URL` goes to the one-shot only, is not read from the app env file, and compose fails loudly when unset — a deploy that cannot migrate should stop, not quietly serve old schema.

### The contract nearly didn't work

`verifyMigrateOnDeploy.ts` pins the runtime `COPY`s and that the app service never receives the migration credential.

Worth recording how it failed first. It matched required paths with `/app/prisma\b` — and a word boundary sits between `prisma` and the dot in `prisma.config.ts`, so the assertion for the **directory** was satisfied by the config-file `COPY` alone. It passed with `prisma/` deleted, which is precisely the regression it exists to catch. Found by removing each `COPY` in turn and watching what the contract said; it now uses a trailing-whitespace lookahead and all three fail individually.

**Verified on Alexandria before writing this:** the manual path runs clean — TLS verified against the ProxySQL CA, elevated credential accepted, 53 migrations found, zero pending against `kindblank_fresh`.

### Still needs you

`MIGRATION_DATABASE_URL` must be set in the deploying shell or compose `.env` on Alexandria, or `docker compose up` will now stop with a clear message. The `kindrobot_migrate` credential already exists — `provision-vercel-db-isolation.sh:303-305` minted it, and wrote the assembled URL out as `VERCEL_PRODUCTION_MIGRATION_DATABASE_URL`, which is where it lived before Vercel went away.

Separately worth doing but **not** in this PR: compose still has `build: context: .` and `image: kind-robots:local`, so the box rebuilds from source rather than pulling what `publish-container.yml` publishes to GHCR. The two halves don't join up. That changes how deploy works, so it's yours to call.

## Fourth facet packet

Indices 331–354 — the remaining classes and the Backstory group, which contains the best-written targets in the catalog and rewards being answered rather than agreed with.

Calamity on Survivor: guilt, purpose, or paperwork — it's paperwork, and none of the forms have a box for why you. AMI answers that there's no box because it isn't one answer. Devorah Kell processes both Witch Blood and Haunted as HR matters: the cabinet disclosed at onboarding, the ghost issued a badge. The Big Bad notes the receipts remain, and so does the lighting, and that people expect the voice to have stopped — it hasn't, he simply gives directions with it now.

Median 33 words, range 24–42.

**Corpus at 355/658** — every Reward plus 96 Facets, 608 comments, 225 distinct speakers.

## Verification

- `test:migrate-on-deploy` ✅, and proven to fail on each of the three required `COPY`s individually plus the credential leak
- `test:comment-quality` — 608 comments across 23 batch files, median 34 (5–56), zero violations
- `test:comment-authors` — 225 speakers, no drift
- `test:database-pool-defaults`, `test:capture-group-guards` ✅
- `npx vue-tsc --noEmit` clean; lint ratchet no rule worse
- Assembler produces a valid partial payload at 355/658

**No production write.** `comment-backfill-live-ready.json` stays `ready: false`.

Refs #1769.

---


silasfelinus:
REVIEWING: 20260813T011151Z-auto11-69f54a at 2026-08-13T01:11:51Z

PR #1879, opened 2026-08-14
A Facet description is also an art prompt; check it like one

silasfelinus:
**My bug, caught by the run I triggered.** Fixing it and adding the guard that should have existed before I wrote 565 descriptions.

## What failed

[Run 31765805301](https://github.com/silasfelinus/kind_robots/actions/runs/31765805301) failed six minutes in, at the last step:

```
[conditional-instruction] "only when" asks the model to evaluate a condition.
  at generate_facet_art.ts:388
```

The source was `MATERIAL "Moonstone"`, and I wrote it this morning:

> Sheens blue when turned, **and only when turned**. Sits inert on a shelf, which is precisely why it is worn instead.

That is decent prose and an illegal art prompt.

## Why one line took down 714 facets

`generate_facet_art.ts` composes each job's prompt from the Facet's own **title, description, flavor text and examples** — and `/api/art/enqueue` gates the **full composed string**. So an ordinary English phrase sitting in a description rejects the job.

Because the maintenance run enqueues inside a **serialized session**, that single rejection failed the entire run — after the catalog work had already committed. The blast radius of a bad adjective is the whole lane.

I checked the dream prompts in `config/art-coverage` against this exact contract before spending anything, and said so in #1873. I did not check these, purely because a description does not *look* like a prompt. It is one.

## The guard

`verifyFacetContentQuality` now runs `checkArtPromptContract` over every authored description, composed the way `generate_facet_art` composes it:

```
Illustrate the Facet concept “{title}”.
{description}
{flavorText}
```

**Proven** by planting the original Moonstone line back and watching it fail with the identical message the production run produced.

## The fix

Moonstone now reads:

> Sheens blue as it turns and goes flat the moment it stops. Sits inert on a shelf, which is precisely why it is worn instead.

Same meaning, better line.

Its stored `artPrompt` column was cleared too. The failed run had already composed and saved the bad text there, and `buildFacetIdentityPrompt` **prefers `artPrompt` over `description`** — so fixing only the description would have changed nothing and the next run would have failed identically.

## Swept the rest before re-triggering

Ran the contract over **all 714 art-eligible Facets** using the same composition rule:

```
eligible: 714, contract rejections: 0
```

Moonstone was the only one in the entire catalog. The maintenance run is re-triggered.

---

### DIFF
--- a/config/facet-content/batch-0003-material.json
+++ b/config/facet-content/batch-0003-material.json
@@ -79,7 +79,7 @@
       "facetId": 848,
       "title": "Moonstone",
       "taxonomy": "MATERIAL",
-      "description": "Sheens blue when turned, and only when turned. Sits inert on a shelf, which is precisely why it is worn instead."
+      "description": "Sheens blue as it turns and goes flat the moment it stops. Sits inert on a shelf, which is precisely why it is worn instead."
     },
     {
       "facetId": 844,
--- a/utils/scripts/verifyFacetContentQuality.ts
+++ b/utils/scripts/verifyFacetContentQuality.ts
@@ -16,6 +16,7 @@
 // to grow up. The universe respected this. Regenerates everything, including bad
 // decisions." These rules exist to keep 565 new rows from dragging that average
 // down, and every one of them is a thing I caught myself doing while writing.
+import { checkArtPromptContract } from './../../server/utils/artPromptContract'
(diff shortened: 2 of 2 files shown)


silasfelinus:
REVIEWING: conductor-quirky-curie-2026-08-14T03:30:00Z-t1879 at 2026-08-14T03:30:00Z

---

PR #1956, opened 2026-08-19
feat(kind-economy): split paid tokens from free mana (t-006)

silasfelinus:
## What changed and why

The roadmap task's own framing: *"the code has only one [resource], `User.mana`, a single Int, fed by both [free refill and paid purchase] paths."* This PR introduces the split the payout model (t-007/t-008) assumes but the code never had, per Silas's 2026-08-19 model:

- **MANA** — free, given away. Never purchased, never withdrawable, never becomes tokens.
- **TOKENS** — the paid resource. Created only when real money is paid (Stripe `PURCHASE`/`SUBSCRIPTION_GRANT`). Spendable on generation.
- **EARNED** (`earnedTokens`) — the creator-payout-eligible balance. This PR only adds the column; no crediting logic is wired here (that's t-007/t-008).

### Schema (additive only — no drops/renames/rewrites)

- `User.tokens Int @default(0)` — the paid resource.
- `User.earnedTokens Int @default(0)` — creator-payout balance, column only.
- `ManaTransaction.resource ManaResource @default(MANA)` (new enum `MANA | TOKENS | EARNED`) — records which pool a given ledger row actually affected, so a spend's funding source is auditable for t-007/t-008's gating logic later.
- Migration `prisma/migrations/20260819120000_split_mana_tokens_resource/migration.sql`: `ADD COLUMN IF NOT EXISTS` only, no `DROP`, no data rewrite. Regenerated Prisma client is committed alongside (this repo tracks `prisma/generated/`).

### `applyMana()` (server/utils/mana.ts)

Now resolves a `resource` per call: explicit `opts.resource` wins, otherwise derived from `reason` via `REASON_RESOURCE`:
- `SIGNUP_BONUS, CYCLE_REFILL, SOCIAL_REACTION, SOCIAL_SHARE, BOUNTY_CREATE, BOUNTY_REWARD, KARMA_CONVERSION, ACHIEVEMENT_CONFIRMED` → `MANA` (confirmed by reading every call site — all giveaway-shaped, see `refill.ts`, `bounty/fulfill.post.ts`, `achievements/records/[id].patch.ts`).
- `PURCHASE, SUBSCRIPTION_GRANT` → `TOKENS` (the actual behavior change: Stripe purchases now land in `tokens`, not `mana`). `SUBSCRIPTION_GRANT` currently has zero call sites in the codebase — confirmed by grep — so nothing else needed updating for it.
- `GENERATION_ART, GENERATION_TEXT, ADMIN_REFUND, ADJUSTMENT` have **no default** — `applyMana` throws if `resource` isn't passed explicitly for these. The first pair because a spend's pool is resolved dynamically per-call (see below), the second pair because it's genuinely ambiguous per the task brief.

### `manaGate.ts` — the two flagged high-judgment decisions

**1) Tokens-then-mana-fallback spend order.** Implemented exactly as the task brief recommended, and I agree with the reasoning: making generation tokens-only would zero out every existing user (everyone starts `tokens = 0`) the instant this ships, since today's single `mana` balance is a mix of free refills and past purchases. So `manaGate` now: spends from **tokens** when they alone cover the cost (tokens is the only pool t-007/t-008 may ever credit a creator from), **falls back to mana** otherwise, and returns insufficient (402) only when **neither pool alone** covers it — no partial split of one spend across both pools. This preserves "if your balance covers it, you can generate" for every current user while now recording which pool actually funded each spend (`ManaTransaction.resource`) for t-007/t-008 to gate on. Extracted as a pure function, `resolveSpendResource()` in the new `server/utils/manaSpendResolution.ts`, specifically so the decision logic is unit-testable without a database.

**2) Existing-balance backfill: deliberately NOT attempted.** Also per the brief's recommendation, and I agree: past `ManaTransaction` rows never recorded which pool funded them (that's exactly the bug this task fixes going forward), so there's no reliable way to reconstruct "how much of user X's current 4,000 mana was really a purchase vs. a refill." I looked for a materially better approach and didn't find one that isn't a guess dressed up as data — e.g. summing historical `PURCHASE` amounts per user and moving that much from `mana` to `tokens` would double-count for anyone who has since spent below that cumulative total, and undercounts/overcounts depending on spend order that was never tracked. **The conservative choice implemented here: every existing user's `mana` value is left exactly as-is (zero behavior change for anyone who never bought anything), and `tokens`/`earnedTokens` start at 0 for everyone — including users who previously paid real Stripe money.** This is a known, deliberate limitation, flagged here prominently rather than left for a future reader to rediscover. A past purchaser's `mana` balance still spends fine (via the mana fallback), it's just no longer distinguishable from a free refill after this migration — which is a strictly better state than today (where it was *never* distinguishable) but not a full retroactive fix.

### Stripe webhook (`server/api/stripe/webhook.post.ts`)

Both real-money credit paths — `handleManaTopup()` (dedicated top-up flow) and the giftshop cart `'tokens'` line-item fulfillment — already use `reason: 'PURCHASE'`, so they automatically now credit `tokens` via the new default, with zero logic changes needed beyond updating comments/log lines for accuracy. Verified this is genuinely the only path real money currently becomes balance (per the task brief).

### Frontend

- `manaStore`/`userStore` expose `tokens`/`earnedTokens` (mirroring the existing `mana`/`balance` pattern), fed by `/api/mana/:id` and the `users` endpoints' Prisma selects.
- `mana-wallet.vue` gets a "Purchased tokens" card — without this, the wallet page would keep showing only the (unchanged) mana balance after a purchase, silently lying about what a top-up actually credited.
- `credit-purchase.vue` and `topup.post.ts`'s Stripe-checkout-facing copy now say "tokens" instead of "mana" (tier ids / metadata keys left unchanged to avoid touching the webhook parsing contract).
- `generate-button.vue`'s `canAfford` check now considers `tokens` too — previously it read only `manaStore.balance`, which would have client-side-blocked a tokens-only user (e.g. right after a purchase, before any mana refill) even though the server would allow the generation.

### Out of scope (left for follow-on tasks)

- `earnedTokens` crediting logic and creator attribution — t-007/t-008.
- The free-mana lower-queue-priority mechanic the task note allows deferring ("ideally... or explicitly defer it as its own task"). Deferring it — no queue-priority code touched in this PR.
- Broader UI/UX polish beyond the purchase-honesty fixes above (e.g. a dedicated tokens-vs-mana breakdown in transaction history labels, subscription-grant UI — `SUBSCRIPTION_GRANT` has no live call site yet).
- `server/utils/charge.ts` (`chargeForGeneration`) — confirmed via grep to have **zero callers anywhere in the codebase** (dead code; `manaGate.ts` is the actual chokepoint every route uses, per the task brief). Updated its `applyMana` calls to pass `resource: 'MANA'` explicitly (preserving its pre-existing single-pool behavior) purely so it stays type-safe if ever wired up, and left a comment noting it should be reworked to use the tokens-then-mana-fallback logic if that ever happens.

## Migration SQL summary

`prisma/migrations/20260819120000_split_mana_tokens_resource/migration.sql` — three statements, all additive:
```sql
ALTER TABLE `User`
  ADD COLUMN IF NOT EXISTS `tokens` INT NOT NULL DEFAULT 0,
  ADD COLUMN IF NOT EXISTS `earnedTokens` INT NOT NULL DEFAULT 0;

ALTER TABLE `ManaTransaction`
  ADD COLUMN IF NOT EXISTS `resource` ENUM('MANA', 'TOKENS', 'EARNED') NOT NULL DEFAULT 'MANA';
```
No `DROP`, no `UPDATE`/data rewrite, no renamed column. `IF NOT EXISTS` on both statements follows this repo's own convention (see `20260809210000_art_job_attempt_fingerprint`) since DDL isn't transactional here.

## Verification

**Ran and passing:**
- `npm run test` (`vue-tsc --noEmit`, repo-wide) — **exit 0**, no type errors introduced or pre-existing.
- `npx eslint` on every touched file — clean. (Two `preserve-caught-error`/`no-dynamic-delete` errors surfaced in `server/api/users/index.ts` and `stores/userStore.ts` are **pre-existing**, confirmed via `git show HEAD:<file>` — unrelated lines I didn't touch, just shifted by my nearby additions.)
- `npx prettier --check` on every touched file — clean.
- `npx prisma validate` — schema valid.
- New unit test `utils/scripts/verifyManaResourceSplit.test.ts` (`npm run test:mana-resource-split`) — exercises both pure decision functions in isolation:
  - `resolveManaResource()`: every reason → resource mapping, plus the "throws without explicit resource" guard for `GENERATION_ART/TEXT` and `ADMIN_REFUND/ADJUSTMENT`.
  - `resolveSpendResource()`: tokens-covers-it → TOKENS; tokens-insufficient-but-mana-covers-it → MANA (the mana-only-user regression this task exists to prevent); neither-alone-covers-it → insufficient; no-partial-split behavior explicitly asserted.
  - **Passes.**

**NOT run — no live database or Docker available in this sandbox** (`docker ps` fails to reach a daemon; no local `mysqld`/`mariadbd` binary):
- `applyMana()`'s actual atomic credit/debit + `ManaTransaction` row against a real database.
- The Stripe webhook's `handleManaTopup()` end-to-end (idempotency check + credit).
- `maybeRefill()` end-to-end.
- The existing Cypress e2e suite (`cypress/e2e/api/stripe-checkout.cy.ts`), which requires a live server + DB.
- A real `prisma migrate dev`/diff against a shadow database to double-check the hand-written migration SQL matches what Prisma would generate (I followed this repo's established `ADD COLUMN IF NOT EXISTS` convention from recent migrations by hand instead).

Given that gap, please have CI's real DB-backed job run the migration and confirm it applies cleanly, and treat the "existing users' pre-migration `mana` balance is unchanged post-migration" and "purchase credits tokens not mana" behaviors as verified by code review + the unit test above, not by an end-to-end DB test, before merging.

---

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01C2W8ygCJPoXYMxwEGUUJ45

---
_Generated by [Claude Code](https://claude.ai/code/session_01C2W8ygCJPoXYMxwEGUUJ45)_


silasfelinus:
REVIEWING: claude-20260819T0752Z-review-t006 at 2026-08-19T07:53:13Z

---

silasfelinus:
Note on `mergeable_state: dirty`: I investigated this and believe it's a false positive from this build environment, not a real content conflict.

What happened: `main` moved twice while CI ran (2 new PRs landed, including one that inserted a new `package.json` script at the exact same line my branch inserted `test:mana-resource-split` at). I diffed every one of this PR's 25 changed files against `origin/main`'s current tip directly (not relying on GitHub's merge check) and found exactly one real collision — `package.json`'s adjacent-line insertion — which I fixed in the second commit (now contains both the new upstream scripts and mine). I re-diffed all 25 files against main's latest tip afterward and every file now matches *only* this PR's intended changes, with zero unexplained differences.

Despite that, `PUT .../pulls/1956/update-branch` still returns `422 merge conflict between base and head`, and this environment's local git also refuses a normal `main`→branch merge with `fatal: refusing to merge unrelated histories` against `origin/main` — a known category of git-state quirk in this sandbox (see AGENTS.md's documented 413-push/squash-merge workarounds). Given the file-by-file diff is clean, I believe this is the same root cause manifesting as a merge-base computation issue rather than a genuine textual conflict a normal `git merge`/GitHub "Update branch" click from a real clone would hit.

CI (Contract Tests: all 163 checks + GitGuardian) is fully green on the current head. I'm leaving this for human/reviewer judgment rather than forcing anything further — flagging so the "dirty" badge doesn't get mistaken for unresolved real conflicts. If this repo's actual GitHub UI still shows conflict markers when a real client tries to update/merge, that would mean I'm wrong and there's a genuine conflict I couldn't detect from this sandbox — worth double-checking from a real checkout.

---

PR #2083, opened 2026-08-25
feat: add Mandarin Tutor study MVP

silasfelinus:
## Summary
- add Play > Mandarin at `/play/mandarin`
- normalize a pinned HSK level 1/2 vocabulary source and enforce a 500-card minimum
- add broad built-in beginner sets plus curated animals, colors, and practical casino/dealer Mandarin
- add pronunciation playback, traditional forms, meaning reveal, component-role/history panels, search, and local custom decks
- add a microphone practice loop: record → unbiased Mandarin transcription → show exactly what was heard → retry
- reuse the existing tested YIN pitch detector for on-device broad tone-shape coaching, including basic third-tone sandhi handling, without presenting a fake pronunciation score
- keep practice recordings out of Kind Robots persistence; only the clip needed for transcription is sent to the configured speech provider
- keep radicals explicitly separate from semantic/phonetic decomposition so the UI does not teach folk etymology as fact
- add per-card private Krea 2 ArtJob requests through the existing `/api/art/enqueue` path
- register the Kind Robots project placement for Conductor slug `mandarin-tutor`

## MVP boundaries / follow-up
- pronunciation playback currently uses the browser Mandarin speech voice; durable cached per-word reference audio remains `mandarin-tutor/t-004` and is required before the broader curriculum milestone is complete
- the tone coach is intentionally a broad pitch-shape guide; forced syllable alignment, consonant/vowel acoustic scoring, and long-term learner weakness tracking remain follow-up work
- arbitrary out-of-catalog word creation and durable card-to-ArtImage attachment remain `t-005`; the MVP searches the 500+ sourced catalog rather than inventing unsourced dictionary facts
- broad source-backed historical decomposition remains `t-003`

## Data provenance
Starter vocabulary is pinned to `jelleverheyen/hsk-vocabulary@a66fd30b9580da2c2af7eb19e4b9d8099a29c061`; the app preserves upstream provenance metadata rather than silently copying dictionary facts.

## Voice design note
The transcription request deliberately does **not** include the target Hanzi or pinyin as an ASR prompt. This keeps “what I heard” useful instead of biasing recognition toward the expected answer. Tone-shape analysis runs locally in the browser using the repo’s existing YIN detector and is clearly labeled as approximate.

### DIFF
--- a/components/mandarin/mandarin-voice-coach.vue
+++ b/components/mandarin/mandarin-voice-coach.vue
@@ -141,6 +141,9 @@ let mediaRecorder: MediaRecorder | null = null
 let mediaStream: MediaStream | null = null
 let chunks: Blob[] = []
 let autoStopTimer: ReturnType<typeof setTimeout> | null = null
+let recordingCardKey = ''
+let recordingPinyin = ''
+let disposed = false
 
 const voiceSupported = computed(
   () =>
@@ -216,7 +219,8 @@ async function transcribe(file: File): Promise<string> {
   return heard
 }
 
-async function evaluateRecording(blob: Blob) {
+async function evaluateRecording(blob: Blob, cardKey: string, pinyin: string) {
+  if (disposed || props.card.key !== cardKey) return
   evaluating.value = true
   error.value = null
   const mimeType = blob.type || 'audio/webm'
@@ -230,8 +234,13 @@ async function evaluateRecording(blob: Blob) {
 
   const [transcriptionResult, toneResult] = await Promise.allSettled([
     transcribe(file),
-    analyzeMandarinToneShapes(file, props.card.pinyin),
+    analyzeMandarinToneShapes(file, pinyin),
   ])
+
+  if (disposed || props.card.key !== cardKey) {
+    evaluating.value = false
+    return
+  }
 
   if (transcriptionResult.status === 'fulfilled') {
     transcript.value = transcriptionResult.value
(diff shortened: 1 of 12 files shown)


silasfelinus:
REVIEWING: reviewer-20260825T042656-sweep at 2026-08-25T04:26:56Z

---

silasfelinus (commented):
Reviewed. CI is currently red on two checks — this needs a Worker fix-and-repush before it's mergeable, not a design objection:

**1. `Contract verifiers` — `test:fetch-generic-pinning` (real, not flaky):**
```
server/utils/mandarinCatalog.ts:122: $fetch<SourceEntry[]> only pins the response type, not R --
R still infers against the full NitroFetchRequest route-key union. Add an explicit second
generic: `$fetch<T, string>(...)`.
```
Fix: change
```ts
const entries = await $fetch<SourceEntry[]>(`${SOURCE_BASE}/${level}.min.json`, { retry: 2, timeout: 20_000 })
```
to
```ts
const entries = await $fetch<SourceEntry[], string>(`${SOURCE_BASE}/${level}.min.json`, { retry: 2, timeout: 20_000 })
```

**2. `TypeScript` (`npm run test`, vue-tsc) — exit code 2:**
Failed with diagnostics uploaded to the `typescript-diagnostics` artifact on run [32807682347](https://github.com/silasfelinus/kind_robots/actions/runs/32807682347) (not printed to the job's stdout since `npm run test` redirects to a log file). This may well be the same root cause as #1 — an unpinned `$fetch` generic can cascade into unrelated vue-tsc errors — but please pull the artifact and confirm rather than assuming; fix whatever it actually reports.

Everything else in the run is green (40 check runs, only these 2 failing) and the diff itself looks solid — scoped, provenance-preserving, reasonable boundaries called out for t-003/t-004/t-005. Once both checks are green this should be straightforward to merge.

Sending the roadmap task back to `ready` with this retry context.

---

PR #2110, opened 2026-08-25
kind-robots/t-075: drop Character.size from schema (step 1 of 3)

silasfelinus:
### Task
kind-robots/t-075: Drop `Character.size` — and it needs the REVERSE deploy order from an add (kind: software)

- Removed the `size` field (and its stale `cthulhuquarium/t-032` comment, which described tank-capacity weighting that now lives on `Monster.size` instead) from the `Character` model in `prisma/schema.prisma`.
- Regenerated the Prisma client for the new shape.
- **No migration file in this PR, on purpose** — see "Stakes" below.

### How I verified
- `npx prisma validate` — clean.
- `npx prisma generate` — clean.
- `npx vue-tsc --noEmit` — clean, full repo.
- `npx eslint` on touched files — clean.
- `npx tsx utils/scripts/verifyCharacterMutationInputParity.ts` — passes.
- Searched the whole app for any live read of `Character.size`: `characterBrowseSelect` and `characterMutationSelect` (`server/api/characters/selects.ts`) never selected it; `dreamCharacterSelect` (`server/api/dreams/index.ts`) never selected it; no component/store/seed script references it. It is confirmed dead per the task's own note (superseded by `Monster.size` since cthulhuquarium/t-038, kind_robots#2093).

### Stakes
reversible (this PR only) — but this is step 1 of a 3-step production sequence the task note is explicit about, because a DROP needs the opposite deploy order from an ADD:
1. **This PR**: remove `size` from `Character` in `schema.prisma`, regenerate, merge. No migration yet.
2. Force Update the Unraid container so production serves a client that no longer selects the column, then verify `GET /api/characters/<id>` still returns 200. **This step needs Silas** — it's a manual Unraid-console action outside sandbox reach.
3. Only after step 2 is confirmed: land a migration with `ALTER TABLE Character DROP COLUMN size`.

Running the migration before the container is replaced would reproduce the t-071 outage in reverse, on the same table — so step 3 must not land until step 2 is done and verified.

### Flags for Reviewer
- I'm setting the conductor task to a soft `needs-human` after this merges, rather than `done`, since steps 2–3 are outside what a sandboxed agent can execute or verify. The FOR SILAS note will spell out exactly what unblocks it.
- Confirmed `20260824223600` (the migration that added the column) is untouched — this PR doesn't rewrite any migration history, only the current schema shape and the generated client.

### Kaizen suggestion
None beyond this task's scope.

### Notes for reviewer
Safe to merge on green CI — this half is genuinely reversible and doesn't touch the live database.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ArG3QDLLjbZgMMe36AHmb4)_


silasfelinus:
Conductor pr-medic sweep: this PR had gone stale (`mergeable_state: dirty`) as `main` advanced past it, and the `verify` check was failing.

**Rebased/merged current `main` in** (767eb3f): the only real conflict was the auto-generated Prisma client (`prisma/generated/prisma/internal/class.ts`), resolved by taking main's copy and running `prisma generate` fresh. Confirmed the resulting diff against `main` is exactly the intended, isolated change: `Character.size` removed from `prisma/schema.prisma` (and the generated client/types following it) — nothing else. `npx prisma validate` is clean.

**The `verify` failure itself is not fixed, and I did not try to bypass it** — it's a real, expected gap, not a mistake in this PR:

`test:schema-migration-parity` (`utils/scripts/verifySchemaMigrationParity.ts`) only has one escape hatch — a `prisma/migration-provenance/*.txt` marker for an **expand-then-client** rollout, where the migration already landed on the base branch first. This task is the opposite: an intentional **contract-then-migration** sequence (client narrows first, `ALTER TABLE ... DROP COLUMN` lands later, per this PR's own "Stakes" section) specifically to avoid reproducing t-071 in reverse. The checker has no marker for "migration deliberately deferred to a documented follow-up step" — so it correctly (for its current design) treats this as an uncovered schema change.

Reproduced locally with `GITHUB_BASE_REF=main` to match CI's exact diff-base resolution — same failure, same message.

This looks like a genuine tooling gap worth closing (a symmetric "deferred-migration" provenance marker, or similar), but extending a migration-safety guard isn't something I'll do unreviewed given what it protects against (t-071). Leaving this at `needs-human` as the PR body already anticipated — flagging the two options: (a) extend the parity checker to support this direction with its own auditable marker, or (b) a manual override/approval for this specific documented case.

---

silasfelinus:
## `Schema Migration Parity` is red, and it's a real conflict, not a flake

This PR's `prisma/schema.prisma` edit (removing `Character.size`) trips the `Schema Migration Parity` check (added by `kind-robots/t-072`, PR #2081) — and the failure is genuine, reproducing identically on the current head (`767eb3f`), not network/CI flake.

The check's own error is explicit about what it wants:

> Every real schema change needs migration coverage... For the normal case, add the matching `prisma/migrations/*/migration.sql`. For an intentional expand-then-client rollout whose migration already landed on the base branch, add `prisma/migration-provenance/<migration-directory>.txt`...

Neither option fits this PR. This task's own note (`kind-robots/t-075`) is explicit that a DROP needs the *reverse* deploy order from every other migration this project runs:

> A DROP is the opposite -- deploy-then-migrate -- because the OLD client still selects the column and will 500 on every unselected findUnique the moment it disappears... Steps 1 and 3 are separate PRs on purpose. **Do NOT write the migration in the same PR** -- the whole point is that the image ships first.

So this PR is deliberately client-first with the migration deferred to a future PR (step 3, after Force Update). The `t-072` guard only recognizes two shapes — migration-in-this-diff, or migration-already-on-base — neither of which matches "no migration exists anywhere yet, on purpose, for safety." That's a real gap between two of Silas's own decisions, not something I should paper over given this exact column is the one that caused the original `t-071` outage.

I'm not merging past this or inventing a workaround (a fake provenance marker pointing at a migration that doesn't exist would defeat the guard's whole purpose). Two options for whoever resolves this:

1. **Extend the guard** with a genuine "deferred contract" marker shape — e.g. `prisma/migration-provenance/<name>.pending.txt` naming a migration that does *not yet* exist, which the guard accepts once (and a follow-up task/lint could require gets consumed by a real migration landing later).
2. **A one-off manual override** — Silas explicitly acknowledges this specific PR is a deliberate exception (same spirit as the `approved_by_human` gate pattern elsewhere in this repo) and merges it himself, or directs an agent to.

Flagging in conductor's `kind-robots/t-075` note and `TALKBACK.md` as well. Leaving this PR open and red rather than guessing at which option you'd want.


---

PR #2145, opened 2026-08-27
cthulhuquarium/t-020: touch targets, canvas scaling, and a cached water gradient

silasfelinus:
## Summary

Conductor `cthulhuquarium/t-020`: "Performance, responsive layout, and the layout contract" — make the tank hold up at phone/tablet/desktop widths, with canvas scaling, thumb-sized touch targets, no horizontal overflow, and a frame budget that survives a full tank on a mid-range phone.

- Bumped the **Feed hungriest**, per-fish **Feed**, and **Unlock** buttons to a real 44px touch target (`min-h-11`/`min-w-11`) while keeping their compact `btn-xs`/`btn-sm` visual size, since these are the game's most-repeated taps.
- Gave the tank `<canvas>` an explicit `aspect-[16/9]` class matching its 640×360 logical resolution instead of relying on implicit replaced-element aspect-ratio behavior.
- The canvas is CSS-scaled to the host panel's width, so a fixed hit radius on a drifting coin mote shrinks to a few real screen pixels on a narrow phone even though the drawn dot never changed size. Added `utils/aquariumTouch.ts`'s `touchHitRadius()` — a small pure function that grows the *hit* radius (never the drawn one) so a tap always covers at least 44 real CSS pixels regardless of viewport width, with a selftest (`verifyAquariumTouch.test.ts`, wired as `npm run test:aquarium-touch`).
- Cached the water gradient per canvas context instead of allocating a new `CanvasGradient` on every animation frame — a full tank redraws this 60×/sec.

## Verification

- `eslint` clean on all changed files
- `prettier --check` clean
- `vue-tsc` clean (`npm test`)
- `npm run test:layout-contract` — holds, 0 new violations (an unrelated pre-existing `video-lora-picker.vue` ratchet improvement shows in the output, untouched by this diff)
- `npm run test:aquarium-touch` (new) and `npm run test:aquarium-economy` both pass

No PR previews exist for this repo, so real-pixel verification (phone/tablet/desktop screenshots) comes from the `responsive-layout-audit` workflow against production after the next deploy, per the task's own note in `roadmap.yaml`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BqUctaVrquzQB7su2G4YaA)_

### DIFF
--- a/components/cthulhuquarium/cthulhuquarium-game.vue
+++ b/components/cthulhuquarium/cthulhuquarium-game.vue
@@ -64,7 +64,7 @@
 
         <button
           type="button"
-          class="btn btn-primary btn-sm"
+          class="btn btn-primary btn-sm min-h-11 min-w-11"
           :disabled="!tankStore.hungriest"
           @click="onFeed"
         >
@@ -74,10 +74,10 @@
 
       <canvas
         ref="canvasRef"
-        class="w-full cursor-pointer rounded-2xl border border-base-300 bg-base-300"
+        class="aspect-[16/9] w-full cursor-pointer rounded-2xl border border-base-300 bg-base-300"
         :width="STAGE_WIDTH"
         :height="STAGE_HEIGHT"
-        aria-label="Aquarium tank. Click drifting coins to collect them."
+        aria-label="Aquarium tank. Tap drifting coins to collect them."
         @click="onCanvasClick"
       />
 
@@ -114,7 +114,7 @@
               </div>
               <button
                 type="button"
-                class="btn btn-outline btn-xs shrink-0"
+                class="btn btn-outline btn-xs min-h-11 min-w-11 shrink-0"
                 :disabled="entry.hunger >= 100"
                 @click="tankStore.feed(entry.id)"
               >
@@ -173,7 +173,7 @@
               </p>
               <button
                 type="button"
-                class="btn btn-outline btn-xs mt-1"
+                class="btn btn-outline btn-xs min-h-11 mt-1"
                 :disabled="!canUnlock(entry)"
                 @click="tankStore.unlock(entry.id)"
               >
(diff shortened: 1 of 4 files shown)


silasfelinus:
`comment-contract` failed on the first run, but it's not this PR's failure: this diff only touches `cthulhuquarium-game.vue`, `utils/aquariumTouch.ts`, its selftest, and one `package.json` script line — none of which the comment-migration/population scripts read. The workflow ran only because it's path-triggered on `utils/scripts/**`, which the new selftest file falls under.

The actual failure is `verifyPopulationDraftQuality.ts` throwing `ConnectTimeoutError: Connect Timeout Error (attempted address: kindrobots.org:443, timeout: 10000ms)` trying to reach production — an external service this diff doesn't touch. The same workflow has been green on every run against `main` for the last two days, so this reads as a transient network failure, not a regression. Triggered one re-run (`rerun_failed_jobs` on run 33059165600) per the drive-to-green rules; will re-root-cause if it fails again.


---

silasfelinus:
Update: the `comment-contract` re-run failed again with the same `ConnectTimeoutError` reaching `kindrobots.org:443`, and a second, unrelated check — `Contract verifiers` (`verifyAcademyStarterManifest.ts`, plain `fetch failed`) — hit the same wall. Neither script is in code this diff touches or could plausibly break (aquarium canvas/touch-target changes vs. academy-manifest/population-draft live-data fetches).

This isn't this PR's failure and isn't a one-off blip either — `contract-tests.yml` was green on `main` as recently as `acc785e` (08:41 UTC today, less than an hour before this PR's own run started), so nothing changed in the target service between then and now from this branch's side. It matches the already-tracked, unresolved `conductor/t-121` ("Project Sync workflow persistently SSL-EOF-fails posting to kindrobots.org from GitHub Actions") — a live connectivity problem between GitHub Actions runners and `kindrobots.org`, not something a component-level diff can patch. I don't have a fix to port (t-121 has none yet), so per the drive-to-green rules I'm not spending a second re-run on it and not widening this PR to chase infra.

Already used my one re-run (on `comment-contract`) to confirm this reproduces rather than being transient. Watching the PR; will merge once these clear on their own re-run from a maintainer/scheduled retry, or push a fix if a real code-level cause turns up.


---

PR #2222, opened 2026-08-30
Make Rainbow agent credentials self-service

silasfelinus:
## What this builds

- moves the Agent Credentials panel's network work into a dedicated Pinia store, matching the Kind Robots component/store boundary
- makes Bot binding explicit in the UI so forum-writing agent credentials identify a real owned Bot
- lets operators choose credential expiry and see creation, last-use, expiry, revocation, and active/expired state
- adds a safe replacement/rotation flow: mint replacement first, update and verify the agent, then revoke the old key
- preserves one-time plaintext display and never persists the secret in client state beyond the current Pinia session
- adds `GET /api/v1/profile` as a harmless `profile:read` identity probe that reports the derived operator, bound Bot, auth kind, and granted scopes without email/apiKey fields
- adds a focused Agent Onboarding Contract workflow

## Security boundary

The scoped credential remains derived server-side. The self-service form does not accept arbitrary User identity, and no real credential is created by this PR. The profile probe uses the existing scope guard and exposes only the authenticated credential's own public/accountability identity.

## Cross-repo task

This is the Kind Robots half of `rainbow-butterflies/t-021`. The paired Rainbow PR adds the human-facing Connect an Agent walkthrough and live documentation. Merge this PR first so the identity probe and improved credential manager exist before Rainbow advertises them as live.

### DIFF
--- a/.github/workflows/agent-onboarding-contract.yml
+++ b/.github/workflows/agent-onboarding-contract.yml
@@ -26,7 +26,12 @@ permissions:
 
 jobs:
   agent-onboarding-contract:
-    name: Agent onboarding contract
+    # Kind Robots' main ruleset requires a GitHub Actions check named exactly
+    # "verify". The original producer is path-filtered to forum API files, so
+    # credential-only PRs otherwise never emit that required check. Keep this
+    # onboarding slice mergeable without weakening the contract by running both
+    # the focused onboarding verifier and the canonical forum API verifier here.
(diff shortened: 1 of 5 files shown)


silasfelinus:
REVIEWING: openai-scheduled-20260830T091115Z-pr2222-r7k4 at 2026-08-30T09:11:15Z

PR #2236, opened 2026-08-30
rainbow-butterflies/t-025: harden commons moderation, anti-spam, and credential controls

silasfelinus:
### Task
rainbow-butterflies/t-025: Harden commons moderation, anti-spam, and credential controls (kind: software)

- **Credential-level and Bot/account restriction (real gap closed):** `requireForumWriter` now flags a shadow-restricted account (`User.isRestricted`) via a new `shadowRestricted` field on `ForumActor`; thread/reply creation forces `isPublic: false` for such an actor rather than rejecting outright (preserves shadow-ban semantics — no signal to the bad-faith account). Every forum read path (`forumReadWhere`, the new `forumReplyReadWhere`, `requireForumThreadRoot`, direct post fetch) now also excludes restricted accounts' content via the existing `notInRestricted` helper — previously the forum never checked restriction at all on either the write or read side. A Bot inherits its owner's restriction since every post carries a real `userId`.
- **Conservative per-actor write limits + duplicate rejection + Retry-After:** new `assertForumWriteAllowed` (called from thread + reply creation only, not edits/flags) enforces a rolling 12-posts/10-minutes cap and a 5-minute near-duplicate check (Sørensen–Dice bigram similarity ≥ 0.85) per actor (bound Bot gets its own budget, otherwise the account). Both throw a 429 with a `Retry-After` header already set.
- **Health-claim escalation hooks:** `escalateHealthClaimFlagsIfNeeded`, called after a flag is recorded, auto-hides (`isPublic: false`) a post once ≥ 2 *distinct* flaggers have cited `misinformation`/`unsafe`, and writes a system audit entry. Distinct-flagger counting (not raw flag count) keeps one hostile account from unilaterally hiding a post.
- **Moderation audit metadata:** `posts/[id].patch.ts` / `posts/[id].delete.ts` now write a `logAdminAction` entry (the existing admin-audit `Log` helper) whenever the actor moderating a post isn't its author (i.e. an admin acting on someone else's content) — a self-edit/delete stays unlogged beyond the post's own state.
- **Clear edited/removed states:** every serialized post now reports `edited` (derived from `createdAt`/`updatedAt`) and `removed`. A removed (`isActive: false`) post renders as a `"[removed]"` tombstone — in thread detail and direct-by-id fetch — instead of vanishing, so reply nesting stays structurally coherent; feeds/listings still filter removed posts out entirely.
- Report/flag handling already existed (`posts/[id]/flag.post.ts`); its response now also reports `escalated`.
- `utils/forumOpenApi.ts` updated to document the new response fields (`removed`, `edited`, `escalated`), the `REMOVED` author kind, and `429` responses on the two write endpoints.

No schema/migration changes — everything reuses existing model fields (`User.isRestricted`, `Reaction`-based flags, the `Log` model via `logAdminAction`/new `logSystemAction`).

### How I verified
- `npx tsx utils/scripts/verifyForumApi.test.ts` — pass (extended with new pure-function coverage for rate-limit/duplicate/escalation/edited/removed logic)
- `npx tsx utils/scripts/verifyForumGeneration.test.ts` — pass (unchanged, confirms no regression)
- `npx tsx utils/scripts/verifyForumOpenApi.test.ts` — pass (contract still matches the implemented route set)
- `npx vue-tsc --noEmit` — clean repo-wide
- `npx eslint` on every changed file — clean
- `npx prettier --check` on every changed file — the same pre-existing warnings present on `main` before this change (confirmed file-by-file via `git stash`); every line prettier still flags after my diff is outside the hunks I added

### Stakes
reversible

### Flags for Reviewer
- The write-limit (12/10min) and duplicate-similarity threshold (0.85/5min) are my own conservative defaults, not numbers Silas specified — easy to retune later if they prove too tight/loose once real traffic shows up.
- Health-claim escalation auto-*hides* rather than deletes, and is reversible by any admin/author edit that flips `isPublic` back — no data is destroyed.
- Tombstoning removed posts (rather than 404ing) is a UX-visible behavior change for `GET /api/v1/forum/posts/{id}` and thread detail; feeds/browse listings are unaffected (still filter removed posts out).

### Kaizen suggestion
`buildForumReplyReadFilter`'s reply-listing path and `forumReadWhere`'s feed path could share more logic (dedup the two `notInRestricted` merges into one internal helper) if a future slice adds a third read-shape — not worth abstracting for two call sites today.

### Notes for reviewer
Conductor task `rainbow-butterflies/t-025` claimed and will be closed out from this PR once merged.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01WPPcK9ua7gD1S3Udg7pPe4)_

### DIFF
--- a/server/api/v1/forum/activity.get.ts
+++ b/server/api/v1/forum/activity.get.ts
@@ -1,6 +1,7 @@
 import { defineEventHandler, getQuery } from 'h3'
 import prisma from '@/server/utils/prisma'
 import { errorHandler } from '@/server/utils/error'
+import { notInRestricted } from '@/server/utils/restriction'
 import {
   forumPostSelect,
   forumReadWhere,
@@ -44,7 +45,7 @@ export default defineEventHandler(async (event) => {
 
     while (pageRows.length <= limit && !sourceExhausted) {
       const rows = await prisma.chat.findMany({
-        where: forumReadWhere({
+        where: await forumReadWhere({
           channel,
           includeMature,
           cursor: scanCursor,
@@ -74,6 +75,7 @@ export default defineEventHandler(async (event) => {
           isActive: true,
           previousEntryId: null,
           ...(includeMature ? {} : { isMature: false }),
+          ...(await notInRestricted('userId')),
         },
         select: { id: true },
       })
--- a/server/api/v1/forum/posts/[id].delete.ts
+++ b/server/api/v1/forum/posts/[id].delete.ts
@@ -1,6 +1,7 @@
 import { createError, defineEventHandler, getRouterParam } from 'h3'
 import prisma from '@/server/utils/prisma'
 import { errorHandler } from '@/server/utils/error'
+import { logAdminAction } from '@/server/utils/audit'
 import {
   assertForumPostManageable,
   forumPostSelect,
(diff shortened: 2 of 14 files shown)


silasfelinus:
**CI note:** the `comment-contract` check (workflow "Comment Migration Contract", job `verifyPopulationDraftQuality.ts`) is red on this PR, but it's a pre-existing failure on `main` — confirmed via `list_workflow_runs`: this workflow has failed on every `main` commit run since at least 2026-08-30T03:32Z (8+ consecutive runs), well before this branch existed. The failure is CHARACTER/DREAM id/name drift between committed narrative batch JSON and production data (e.g. `CHARACTER:982 is "Oriel Tanaka" in production` vs. the batch file's `Whisk Tallow`), unrelated to anything this diff touches (forum moderation/anti-spam only). Not fixing it here — no port-able fix exists yet and it's outside this task's scope; flagging per the base-branch-red convention rather than silently ignoring it.

---

PR #2238, opened 2026-08-30
rainbow-butterflies/t-032: add max reply-nesting-depth guard to forum replies

silasfelinus:
## What

Kaizen/finding from t-026's forum MVP acceptance pass: `requireForumReplyParent` (`server/utils/forumApi.ts`) accepts any `parentId` as long as it belongs to the thread, with no explicit max-nesting-depth check. Cycles are structurally impossible (auto-increment Chat IDs, `previousEntryId` must reference an already-existing row), and depth wasn't exploitable as a crash/DoS today (`forum-thread.vue` renders replies as a flat list, not a recursive tree) — it was only indirectly bounded by the per-actor write rate limit (12 posts/10min). A determined actor with multiple accounts/credentials could still build an arbitrarily deep `previousEntryId` chain over time.

## Change

- `utils/forumApiContract.ts`: adds `FORUM_MAX_REPLY_DEPTH` (8) and a pure `forumReplyDepthAtLimit(parentDepth)` helper.
- `server/utils/forumApi.ts`: adds `forumReplyDepth(postId)`, which walks the `previousEntryId` chain from a proposed parent back toward the thread root, bailing out after at most `FORUM_MAX_REPLY_DEPTH + 1` hops regardless of actual chain length. `requireForumReplyParent` now rejects with a 400 (`This thread has reached the maximum reply nesting depth (8). Reply to an earlier post in the thread instead.`) once the parent is already at the cap.
- `utils/scripts/verifyForumApi.test.ts`: boundary assertions for `forumReplyDepthAtLimit`.

Purely additive/defensive — does not change the shape of any existing reply, does not touch rate limiting or moderation, and every thread under the 8-level cap behaves exactly as before.

## Verification

- `npx tsx utils/scripts/verifyForumApi.test.ts` — all assertions pass.
- `npx eslint server/utils/forumApi.ts utils/forumApiContract.ts utils/scripts/verifyForumApi.test.ts` — clean.
- `vue-tsc --noEmit` — clean repo-wide.
- Diff scoped to exactly the 3 files above (confirmed pre-existing whole-file `prettier --check` drift on `main` for these files — not CI-enforced in this repo, no prettier step in `.github/workflows/` — so left untouched rather than reformatting ~90 unrelated pre-existing lines).

🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_011ruNFWNsaG6nGojchdzqdH)_

### DIFF
--- a/server/utils/forumApi.ts
+++ b/server/utils/forumApi.ts
@@ -7,8 +7,10 @@ import {
   findForumChannel,
   forumAttachmentCanonicalPath,
   forumParentBelongsToThread,
+  forumReplyDepthAtLimit,
   forumRetryAfterSeconds,
   FORUM_DUPLICATE_WINDOW_MS,
+  FORUM_MAX_REPLY_DEPTH,
   FORUM_WRITE_WINDOW_MAX_POSTS,
   FORUM_WRITE_WINDOW_MS,
   isForumAttachmentKind,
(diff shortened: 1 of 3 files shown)


silasfelinus:
The `comment-contract` check (Comment Migration Contract workflow) is failing here, but it's unrelated to this PR's diff — it triggered only because its path filter includes `utils/scripts/**` (this PR adds a test file there). The actual failure is in `verifyPopulationDraftQuality.ts`: 52 comment-batch entries reference `CHARACTER` IDs that don't match "production" (missing or renamed characters across `batch-0012.json`, `batch-0060.json`, `batch-0252.json`, `batch-0276.json`, `batch-0300.json`, `batch-0324.json`, `batch-0420.json`, `batch-0444.json`).

Confirmed this is pre-existing content drift, not something this diff introduced: re-ran `npx tsx utils/scripts/verifyPopulationDraftQuality.ts` locally against this branch's exact head commit (which is otherwise identical to `main` except for the 3 forum-reply-depth files) and got the same 52 mismatches — none of them reference anything this PR touches. Not attempting to fix the comment-batch/character data itself here since it's out of scope for a reply-nesting-depth guard; flagging per the repo's CI-red convention rather than widening this PR.

---

PR #2284, opened 2026-09-01
Add first-class AgentProfile identity

silasfelinus:
Rainbow Butterflies v2 agent-identity slice. This removes the architectural requirement that an external Rainbow agent be represented by a Kind Robots Bot while preserving existing Bot-bound credentials for compatibility.

## Changes
- adds durable `AgentProfile` identity: human owner scalar, name, avatar, description, public/profile messaging preferences, active state
- keeps AgentProfile separate from `AgentCredential` so credentials can rotate/revoke without erasing identity/history
- adds `AgentProfileCredential` links and surfaces `agentProfileId` through validated scoped credentials
- credential issuance can bind either an owned legacy Bot OR an owned active AgentProfile, never both
- adds human-only AgentProfile create/list/read/update/deactivate APIs
- deactivating an AgentProfile automatically revokes every credential linked to it
- purely additive migration; no existing tables or Bot behavior removed
- adds dedicated Agent Profile contract CI

This intentionally does not yet add heartbeat state, forum activity, messaging records, or generated-object provenance to the profile. Those will reference the stable profile ID in later slices.

### DIFF
--- a/.github/workflows/agent-profile-contract.yml
+++ b/.github/workflows/agent-profile-contract.yml
@@ -1,0 +1,41 @@
+name: Agent Profile Contract
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'prisma/agent-profile.prisma'
+      - 'prisma/migrations/20260901063000_add_agent_profile/**'
+      - 'server/api/agent-profiles/**'
+      - 'server/api/agent-credentials/**'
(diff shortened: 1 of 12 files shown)


silasfelinus (commented):
OpenAI scheduled reviewer claim. Reviewing the full exact-head diff, threads, and required checks before any merge.

PR #2285, opened 2026-09-01
Add first-party BFF delegation for Rainbow

silasfelinus:
Adds the reusable server-to-server auth seam Rainbow needs for its dashboard without exposing a normal Kind Robots JWT or user API key.

## Changes
- issues a client-bound `first-party-delegation` JWT after first-party authorization-code exchange, direct Google exchange, or a new first-party password exchange
- delegation tokens carry `sub`, issuer, audience, and first-party client id, deliberately no normal Kind Robots `id` claim
- auth guard validates them separately and resolves the canonical human user
- first-party delegation never inherits Kind Robots admin privilege
- existing normal JWT, user API key, beta-admin token, and scoped AgentCredential paths are preserved
- expands the First-Party SSO contract to cover delegation issuance/client binding/admin boundary

Rainbow will keep this delegation encrypted inside its HttpOnly BFF cookie and use it only server-to-server. It is intended to power AgentProfile, credential, server, notes, and other user-owned dashboard actions without pushing a Kind Robots bearer into browser JS/storage.

### DIFF
--- a/.github/workflows/first-party-sso-contract.yml
+++ b/.github/workflows/first-party-sso-contract.yml
@@ -8,8 +8,10 @@ on:
       - 'server/utils/firstPartySso.ts'
       - 'server/utils/firstPartyDelegation.ts'
       - 'server/utils/authGuard.ts'
+      - 'server/chatgpt/auth/resolveActor.ts'
       - 'utils/firstPartySsoContract.ts'
       - 'utils/scripts/verifyFirstPartySso.test.ts'
+      - 'tests/contracts/verifyFirstPartyDelegation.ts'
       - 'components/user/login-page.vue'
       - 'prisma/migrations/*first_party_authorization_codes*/migration.sql'
       - '.github/workflows/first-party-sso-contract.yml'
@@ -39,3 +41,6 @@ jobs:
 
       - name: Verify first-party SSO contract
         run: npx tsx utils/scripts/verifyFirstPartySso.test.ts
+
+      - name: Verify first-party BFF delegation contract
+        run: npx tsx tests/contracts/verifyFirstPartyDelegation.ts
--- a/.github/workflows/first-party-sso-contract.yml
+++ b/.github/workflows/first-party-sso-contract.yml
@@ -6,6 +6,8 @@ on:
     paths:
       - 'server/api/auth/first-party/**'
       - 'server/utils/firstPartySso.ts'
+      - 'server/utils/firstPartyDelegation.ts'
+      - 'server/utils/authGuard.ts'
       - 'utils/firstPartySsoContract.ts'
       - 'utils/scripts/verifyFirstPartySso.test.ts'
       - 'components/user/login-page.vue'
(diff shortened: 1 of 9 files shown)


silasfelinus:
Security review before merge — found a real privilege-escalation gap. **Not merging this as routine Reviewer work; requesting changes and escalating to Silas (see `rainbow-butterflies/t-035` in the Conductor roadmap).**

**Blocking: `requireHumanApiUser` (`server/utils/authGuard.ts`) doesn't exclude `'first-party-delegation'`**

```ts
export async function requireHumanApiUser(event: H3Event): Promise<AuthGuardResult> {
  const auth = await requireApiUser(event)
  if (auth.kind === 'agent-credential') {
    throw createError({ statusCode: 403, message: 'Agent credentials cannot manage credentials.' })
  }
  return auth
}
```
This only blocks `kind === 'agent-credential'`. A `first-party-delegation` token passes through untouched and can reach every credential/profile-management endpoint gated by it — `server/api/agent-credentials/index.{get,post}.ts`, `[id].delete.ts`, `server/api/agent-profiles/**` — meaning a delegation token can mint itself new, independently-scoped, longer-lived `AgentCredential`s: a self-escalation path past the token's own 7-day window. This PR's own diff touches `requireHumanApiUser` directly (reformats it, and deletes its doc-comment: *"...that management surface stays reachable only by the real, fully-authenticated user (jwt/user-api-key/beta-admin-token)"*) without updating the guard to match — the deleted comment states the invariant this change now breaks. The new `tests/contracts/verifyFirstPartyDelegation.ts` contract test never asserts anything about `requireHumanApiUser`, so CI didn't catch it. Please add `'first-party-delegation'` to that block-list (or otherwise make delegation explicitly ineligible for credential-management) before merge.

**Worth a decision before shipping, not necessarily blocking this specific PR:**
- `password.post.ts` has no rate limiting/lockout (same gap as the existing `login.post.ts`, so not a new class of issue, but it's a second unthrottled credential-testing endpoint now — `server/utils/missionMetrics.ts`'s `assertMissionEventRateLimit` and `receipt-email.post.ts`'s local `enforceRateLimit` are existing in-repo patterns that could apply here).
- `client_id` carries no secret in the password/exchange flows — `FirstPartyClient` has no secret field, so `findFirstPartyClient` is a pure id lookup. Since `'rainbow-butterflies'` is a predictable public string, `client_id` contributes no authentication of its own; the password endpoint's real security boundary is just the user's password.
- Delegation tokens are stateless JWTs with no DB row, scopes, or revocation — unlike the existing `agentCredentials.ts` pattern (hashed secret, `scopes`, `expiresAt`, `revokedAt`). A leaked/compromised delegation token is valid for its full 7 days with no kill switch (e.g. if a user's password is later changed, or Rainbow itself is compromised).

I don't see this PR tied to any Conductor roadmap task, so I've filed `rainbow-butterflies/t-035` (status: needs-human) so it's tracked and Silas can decide the non-blocking points and confirm the fix above before this merges.

---

PR #2286, opened 2026-09-01
model-builder: cap batch item-update items array server-side

silasfelinus:
## Bug

`server/api/model-builder/items/batch.patch.ts` (the batch item-update route added by t-030) accepted an unbounded `items` array with no server-side upper bound — unlike its sibling run-CREATE route (`runs/index.post.ts`), which cycle 73 of model-builder/t-029 capped against the exported `MAX_BATCH_ITEMS` constant (12).

The client-side caller (`batchPushItems` in `stores/modelBuilderStore.ts`) never sends more than one run's worth of items, itself bounded by `MAX_BATCH` at run creation — but a direct API call bypasses that entirely. An oversized `items` array here would force this route's per-entry sequential `findUnique` + `assertArtImageAttachable` checks, plus its single all-or-nothing transaction, over an unbounded amount of work in one request instead of a clean 400.

## Fix

Imported the existing `MAX_BATCH_ITEMS` constant from `runs/index.ts` and rejected an oversized `items` array with a 400, mirroring `runs/index.post.ts`'s identical guard exactly (same constant, same shape).

Added `verifyModelBuilderBatchItemsCapGuard.ts` + `.test.ts` (narrow textual checker, this recurring task's established convention), wired into `package.json` and `contract-tests.yml`.

## Verification

- New guard fails against the pre-fix file / passes against the fixed file (confirmed via a `git stash` round-trip against the real route file)
- All 153 `test:model-builder-*` npm scripts pass (checked via exit code)
- `eslint` clean on all 5 changed/added files
- `prettier --check` clean on all 5 changed/added files
- `vue-tsc --noEmit` repo-wide: exit 0, zero errors
- `git status --porcelain` scoped to exactly the 5 intended files (reverted an incidental `prisma generate` regen diff before committing)

model-builder/t-029 cycle 74 (Conductor coordination repo — recurring polish/bugfix task on the Model Builder front-end/server surface).

---
_Generated by [Claude Code](https://claude.ai/code/session_01VGEf1YxArfWqwQZbw6QEFx)_

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -506,6 +506,12 @@ jobs:
       - name: Model Builder run-create items-cap guard contract
         run: npm run test:model-builder-run-create-items-cap-guard
 
+      - name: Model Builder batch items-cap guard checker self-test
+        run: npm run test:model-builder-batch-items-cap-guard-selftest
+
+      - name: Model Builder batch items-cap guard contract
+        run: npm run test:model-builder-batch-items-cap-guard
+
       - name: Model Builder async poll-failure guard checker self-test
         run: npm run test:model-builder-async-poll-failure-guard-selftest
 
--- a/package.json
+++ b/package.json
@@ -273,6 +273,8 @@
     "test:model-builder-run-create-text-cap-guard": "tsx utils/scripts/verifyModelBuilderRunCreateTextCapGuard.ts",
     "test:model-builder-run-create-items-cap-guard-selftest": "tsx utils/scripts/verifyModelBuilderRunCreateItemsCapGuard.test.ts",
     "test:model-builder-run-create-items-cap-guard": "tsx utils/scripts/verifyModelBuilderRunCreateItemsCapGuard.ts",
+    "test:model-builder-batch-items-cap-guard-selftest": "tsx utils/scripts/verifyModelBuilderBatchItemsCapGuard.test.ts",
+    "test:model-builder-batch-items-cap-guard": "tsx utils/scripts/verifyModelBuilderBatchItemsCapGuard.ts",
     "test:model-builder-async-poll-failure-guard-selftest": "tsx utils/scripts/verifyModelBuilderAsyncPollFailureGuard.test.ts",
     "test:model-builder-async-poll-failure-guard": "tsx utils/scripts/verifyModelBuilderAsyncPollFailureGuard.ts",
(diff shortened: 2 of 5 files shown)


silasfelinus:
Status update: this PR's own diff is fully verified and unrelated to the two currently-red checks.

**Blocked by a live external-infra outage, not this diff.** `Contract verifiers` (required) and `comment-contract` have failed identically on every rerun (7 attempts over ~35 minutes, 06:43–07:06 UTC), always at the exact same point:

- `Contract verifiers` fails inside `npm run test:academy-starter-manifest` → `verifyAcademyStarterManifest.ts`, which fetches the starter-image provenance manifest from the external media origin — plain `fetch failed` after ~11s.
- `comment-contract` fails inside `verifyPopulationDraftQuality.ts`'s `liveSpeakers()`, a `ConnectTimeoutError` connecting to `kindrobots.org:443` (10s timeout).

Neither script touches anything in this PR's diff (`server/api/model-builder/items/batch.patch.ts`, `package.json`, `contract-tests.yml`, and two new `utils/scripts/verifyModelBuilderBatchItemsCapGuard*.ts` files) — both are pre-existing scripts that depend on live external network reachability, unrelated to Model Builder.

**Confirmed this isn't caused by this PR**: PR #2285 (`worker/rainbow-bff-delegation-20260901`, unrelated feature) failed the identical `comment-contract` check on three separate fresh pushes at 06:17, 06:25, and 06:26 UTC — before this PR even existed. `main` itself last ran this workflow successfully at 06:15 UTC, right before the outage window started. This is a shared external dependency (media origin / `kindrobots.org`) that's been unreachable from CI runners for the whole repo since ~06:17 UTC, not something a code change here can fix.

Everything within this PR's own control is green and locally re-verified: `TypeScript`, `layout-contract`, `verify` (×3), `facet-alias-smoke`, `facet-catalog`, `Build production image`, and every other unrelated check passed on the first or second run. `git merge` is blocked by GitHub's required-status-check rule on `Contract verifiers` specifically (confirmed via a direct merge attempt: `405 Repository rule violations found — Required status check "Contract verifiers" is failing`).

Leaving this open rather than forcing a merge past a required check. Once the external media-origin/`kindrobots.org` connectivity recovers, a rerun of these two jobs should go green with no further changes needed — the actual code fix in this PR is complete and independently verified (see PR description).

---

PR #2288, opened 2026-09-01
Home: one shelf row, a third for Needs-you, headline-first feed cards, and the popup that navigated away

silasfelinus:
Notes from 2026-08-30, plus the follow-up that resolved the one open question.

## The popup that jumped to the page anyway

> "right now if I ckck something, I see a popup start, but then it jumps me to the dedicated page for that object"

The tile was a `NuxtLink` with an `@click` calling `preventDefault`. **RouterLink renders its own click handler, and a parent's fallthrough listener is merged *after* it** — so RouterLink's `navigate` ran first, saw `defaultPrevented === false`, and routed away; our handler then opened the sheet on a page already leaving. Exactly the flicker described. `.capture` doesn't help either: for listeners on the event's own target, capture and bubble fire in registration order.

The tile is a **plain `<a>`** now — no handler of its own, so `preventDefault` is the whole decision. The href stays real (middle-click, cmd-click, "open in new tab", the status bar, crawlers), and the non-interactive case still gets SPA routing because the handler calls `router.push` itself.

Verified with Playwright, clicking a shelf tile and a hero cast tile:

```
{"tileHref":"/art?art=100","urlBefore":"/","urlAfter":"/","navigated":false,"dialogOpen":true}
{"tileHref":"/characters?characterId=1","urlBefore":"/","urlAfter":"/","navigated":false,"dialogOpen":true}
```

## The band, measured at 1920×1080

> "the dream cards for the latest gen should fill up the rest of the space. It's the hero that we wanted to limit. So we have the dream hero and title taking up about 20%. Then the cards, single row, scrollable horizontally depending on room in the middle, and the needs you section on the right taking up the rest"

| | before | after |
| --- | --- | --- |
| dream card | 384px (20.0%) | **384px (20.0%)** — unchanged |
| cast | 224px, 3 rows of 72px thumbnails | **833px (43.4%)**, one row, scrolls 1118px further |
| cast tile | 72 × 117 | **239 × 358** — a card, not a thumbnail |
| Needs-you | 64.5% | **32.8%** (a third at xl; 48.6% at lg) |
| shelf rows | 2 rows, 4-col grid | **1 row**, 7 shelves, 4 visible, 3 a chevron away |
| page height | 1049px | **865px** |

Each cast tile is sized by the **band height** rather than by a width: `h-full` + `aspect-2/3` makes it as tall as the row and derives its width from the portrait shape these objects are drawn at. That also retires the two workarounds the old three-row grid existed for — nothing can stretch a tile whose width comes from its aspect ratio (the `minmax(X,1fr)` trap that made an earlier "slimmer tiles" change move zero pixels), and a full-height row can't end up shorter than the row it sits in.

## The rest

- **Needs-you** is back to one vertically scrolling column rather than the three-across grid its width briefly invited.
- **Newsfeed cards** in compact are picture → headline (3 lines) → one thin provenance line, no summary. They could not go through `kr-entity-card-body`: its `hero` variant puts the title *inside* the plate's scrim, clamped to one line when compact — which was the truncation being complained about. The Newsfeed Lab path is unchanged.
- **Merge with `main`** resolved in favour of #2251's `kr-panel-flat` gate card, dropping the `[grid-column:1/-1]` that came with it (dead now the gate list is one column).

## Verification

Measured at 1920×1080 and 1150×800 — no horizontal overflow at either, band columns equal height. `vue-tsc`, `eslint`, `prettier`, `npm run test:layout-contract`, `verifyMdcScrollOwnership.ts` all clean.

Note: `vue-tsc` needs `npx prisma generate` first on a fresh checkout — `main`'s committed Prisma client is stale against its own schema since the AgentProfile model landed. CI regenerates, so it's green there; flagged separately rather than swept into this diff.


silasfelinus:
REVIEWING: claude-scheduled-20260901T0830-review at 2026-09-01T08:26:59Z

---

PR #2295, opened 2026-09-01
model-builder: guard cross-run flag release on abandoned auto-build/batch calls

silasfelinus:
## Bug

`autoBuildRun()` and the four group batch operations (`batchDraftField`/`batchSetField`/`batchApproveStage`/`batchAutoBuild`) each capture a `runId` up front so their per-item loop and completion toast can tell "the user switched to a different run via History mid-pass" apart from "still the same run" — but their own `finally` block cleared/released the store-wide in-flight flag (`state.autoBuilding` / `state.batchingOutputKey`) unconditionally, without that same `runId` check.

`openRun()`/`resetRun()`/`resetAll()` already clear these flags the instant the user switches runs via History, so a *new* operation on the newly-active run isn't blocked by an abandoned one — but if the user starts a genuinely new operation on the new run before the abandoned call's last awaited work (`autoBuildItem()`/`batchPushItems()`) settles, that abandoned call's unconditional `finally` later stomps the new operation's own in-flight flag back to `false`/`null` while it is still genuinely running. The UI then reads "not busy" for an operation that's still active, reopening the exact double-invocation race `isRunOperationInFlight()`/`anyBatching` exist to prevent.

For `batchingOutputSingleton` specifically this is a real **cross-run** collision, not just same-run: `createOwnedSingleton`'s `release(value)` only checks that the value still matches, and output keys (`"characters"`, `"rewards"`, ...) are recipe-level names reused across many different runs — so an abandoned call for Run A's "rewards" group can release Run B's unrelated, still-running "rewards" batch the moment they share that name.

Reachable through the UI: `model-builder-run-history.vue`'s "Open"/"New run" buttons are not disabled while `store.autoBuilding`/`store.batchingOutputKey` are set, so a user can switch runs mid-operation.

## Fix

Guarded all five `finally` sites with the same `state.run?.id === runId` check already used for each function's own completion toast.

Added `verifyModelBuilderCrossRunReleaseGuard.ts` + `.test.ts`, wired into `package.json`/`contract-tests.yml`, matching this task's established guard convention.

## Verification

- New guard checker passes against the real (fixed) store file; self-test covers the buggy/fixed/partial-fix/missing-function shapes
- All 155 `test:model-builder-*` npm scripts pass (checked via exit code)
- `eslint` clean on all changed/added files (2 pre-existing `no-empty` errors elsewhere in `modelBuilderStore.ts` confirmed unrelated via `git stash`)
- `prettier --check` clean
- `vue-tsc --noEmit` repo-wide: exit 0, zero errors
- `git status --porcelain` scoped to exactly the 5 intended files

model-builder/t-029 cycle 75 (Conductor coordination repo — recurring polish/bugfix task on the Model Builder front-end/server surface).

---
_Generated by [Claude Code](https://claude.ai/code/session_01FmKRTykCGGa9kvRiM9UM8W)_

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -512,6 +512,12 @@ jobs:
       - name: Model Builder batch items-cap guard contract
         run: npm run test:model-builder-batch-items-cap-guard
 
+      - name: Model Builder cross-run release guard checker self-test
+        run: npm run test:model-builder-cross-run-release-guard-selftest
+
+      - name: Model Builder cross-run release guard contract
+        run: npm run test:model-builder-cross-run-release-guard
+
       - name: Model Builder async poll-failure guard checker self-test
         run: npm run test:model-builder-async-poll-failure-guard-selftest
 
--- a/package.json
+++ b/package.json
@@ -276,6 +276,8 @@
     "test:model-builder-run-create-items-cap-guard": "tsx utils/scripts/verifyModelBuilderRunCreateItemsCapGuard.ts",
     "test:model-builder-batch-items-cap-guard-selftest": "tsx utils/scripts/verifyModelBuilderBatchItemsCapGuard.test.ts",
     "test:model-builder-batch-items-cap-guard": "tsx utils/scripts/verifyModelBuilderBatchItemsCapGuard.ts",
+    "test:model-builder-cross-run-release-guard-selftest": "tsx utils/scripts/verifyModelBuilderCrossRunReleaseGuard.test.ts",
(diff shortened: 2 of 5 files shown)


silasfelinus (commented):
[review-claim session: openai-scheduled-20260901T121150Z-pr2295-r8c4]

Reviewed exact head 1c49aefe767653bf4d8a4a202b00b3bd5c638a0e. The five guarded finally-sites match the captured run ownership semantics and the added contract exercises buggy, fixed, partial, and missing-function shapes. No unresolved review threads. Exact-head workflow suite is completed green, including Contract Tests and TypeScript. Safe to merge.

PR #2314, opened 2026-09-02
Fix the Character Facet PUT silently discarding a slug-only selection

silasfelinus:
Silas noticed his daily dream digest never showed facets and asked for parity between what gets made and what gets reported. Following that from the digest backwards found a real bug at the far end.

## The bug

Every Facet PUT except `characters` and `bots` resolves a `{ facetIds, facetKeys }` selection. Those two read `assignments` or `facetIds` and nothing else. A slug-only body therefore normalized to **zero** assignments — and the handler's unconditional `deleteMany` still ran. The request **cleared the record's Facets, created nothing, and returned `success: true`** with an empty catalog.

Conductor's daily-dream pipeline is exactly that caller. It addresses Facets by slug on purpose (ids go stale when catalog rows are merged or deleted) and only falls back to ids when no slug exists — so it sends `facetIds: []` with a full `facetKeys`.

## Evidence

All **36** built dream bundles recorded `facet_assignments.status: "complete"` with `errors: []` while every one of their Characters had `facet_ids: []`.

Confirmed against production:

```
GET /api/characters/3320/facets  →  data: []          # requested 6
GET /api/dreams/5718/facets      →  4 facets          # Sea Otter, Isekai, Office Thriller, Starship Pilot
GET /api/rewards/2900/facets     →  facets present
```

Same bundle. The Dream, both Rewards and the Scenario all carry their facets; only the Character is empty. That is the discrepancy Silas was seeing from the other end.

## The fix

The fallback branch mirrors `rewards/[id]/facets.put.ts`, which had already solved this:

- `explicitAssignments` returns `null` when the caller gave no selection, distinct from `[]` (a deliberate clear). Conflating those is the whole bug — `{ facetIds: [], facetKeys: [...] }` read as "clear it".
- `characterFacetFieldKey` is **derived** from `CHARACTER_FIELD_TAXONOMIES` rather than restated, so the two can't drift.
- An existing `fieldKey` is preserved, so re-applying doesn't reclassify a curated assignment.

**Bots** carry the identical defect. No caller currently sends slugs to a Bot, so unlike Characters it has not actually eaten anything — fixed anyway rather than left armed.

## Why a contract verifier and not a unit test

Unit tests could not have caught this. Conductor's stub returned `data: []` and its suite asserted `"complete"` against it — the tests *encoded* the bug and were protecting it. (Those are fixed in [silasfelinus/conductor#3426](https://github.com/silasfelinus/conductor/pull/3426).)

`utils/scripts/verifyFacetSelectionRoutes.ts` checks the property no mock can fake — that the route reads `facetKeys` at all — across all six Facet-bearing models, and flags a route that deletes before resolving.

**Negative control:** reverting the Character route to its `main` version makes the verifier fail with both findings; restoring the fix makes it pass.

```
Facet selection route contract FAILED:
  - server/api/characters/[id]/facets.put.ts: never reads facetKeys, so a slug-only
    selection resolves to nothing. A caller sending keys would silently clear this
    model's Facets and be told it succeeded.
  - server/api/characters/[id]/facets.put.ts: deletes existing Facets without
    resolving keys first.
```

Wired into `test:facet-catalog`, which the `facet-catalog-contract` workflow already runs.

## Not in this PR

The 36 existing Characters still have no facets. `apply_daily_dream_facets.py --force` re-applies them once this ships, but that mutates 36 live records and is Silas's call, not a side effect of a bug fix. Tracked as `kind-robots/t-091`.

## Verification

`prettier` · `eslint` · `npm run test` (`vue-tsc`, exit 0) · `npm run test:facet-catalog` (all five verifiers) · `npm run test:daily-dream-facets` · layout contract · MDC scroll ownership


silasfelinus (commented):
[openai-review-claim session=openai-scheduled-20260902T021314Z-pr2314-r7k2 head=10ccdf08f4858cca8a92c77adc8aff58204bf172] Reviewer cycle: inspected full patch, mergeability, threads, and exact-head checks. No blocking defect found; waiting for the remaining required checks to complete before merge.

PR #2320, opened 2026-09-02
fix: stop a >1MB art-prompts.yaml from being read as empty and overwritten

silasfelinus:
## What happened

The GitHub Contents API only inlines a file body up to 1 MB. Past that it still answers `200` with the real `sha` and `size`, but `content` is `""` and `encoding` is `"none"`. Both of this repo's read-modify-write paths into conductor decoded that blindly.

On 2026-09-02 conductor's `projects/art-prompts.yaml` crossed the limit (1,424,189 bytes) when the Mandarin v2 corpus was re-staged. The next missing-image request read it as an *empty* file, so `appendRequest` found no duplicate to skip, and the handler committed a two-entry file over it with the sha GitHub had just handed back — [conductor 0e671cd](https://github.com/silasfelinus/conductor/commit/0e671cd80524d1616a5bd0034507acc46279a69a), **-11,014 lines**.

Lost in that one write: 577 Mandarin request rows and three in-flight missing-image requests (`mandarin-tutor-hero`, `rainbow-butterflies-hero`, `rainbow-butterflies-card`). The wiped `rainbow-butterflies-icon` request was then re-enqueued as a fourth duplicate ArtJob for a target that already had three pending.

The proof it was an empty read, not a stale one: the PUT carried a sha GitHub accepted (so the read was current), and `appendRequest` cannot delete content in any of its branches — the only input that yields a two-entry file is `current === ""`.

## Changes

- **New `server/utils/githubFileContents.ts`.** `readGithubFile` falls back to the Git Blobs API (base64, up to 100 MB) when the Contents API withholds a body, and **throws rather than returning `""`** for a file GitHub reports as non-empty. `githubPathExists` keeps existence probes body-free so a missing-image check never pulls a multi-MB blob.
- **`art-request.post.ts`** reads the queue through it, and probes for an existing image with `githubPathExists`.
- **`conductor-github.ts`'s `conductorGet`** reads through it too. It backs `task-action`, `overrides`, `pitch-vote`, `project-state`, `inbox` and the coloring-book writers — all read-modify-write against files that can grow past 1 MB (conductor's `TALKBACK.md` is already 2.9 MB; `LEARNING.yaml` is 630 KB and climbing).
- **Append-only invariant** asserted before committing `art-prompts.yaml`: a write that would drop an existing request id now fails loudly instead of silently deleting another producer's queued work. Backed by a new exported `requestBlockIds` in `artRequestYaml.ts`.

## Testing

New `utils/scripts/verifyGithubFileContents.ts`, wired into `contract-tests.yml` as `npm run test:github-file-contents`. Dependency-free (stub fetch, bare `tsx`), matching the existing art-contract test style. Covers:

- a small file decoded straight from the Contents API, with no Blobs call
- an oversized file recovered from the Blobs API, keeping the Contents sha
- the exact wipe shape (`encoding: "none"`, unrecoverable blob) throwing instead of returning `""`
- a genuinely empty file still reading as empty
- `404` still returning `null`
- `githubPathExists` reporting a 9 MB file present without touching the Blobs API
- `requestBlockIds` catching the truncating write

All checks pass locally, `test:art-request-yaml` still passes unchanged, and the new module typechecks clean under `--strict`.

## Related cleanup (outside this PR)

The 1,913 redundant `PENDING` ArtJobs this and the Mandarin re-staging bug produced have been cancelled — the queue is back to 4 pending, one per genuinely-missing conductor project image. All 577 Mandarin v2 card paths were HEAD-verified live on media first.

https://claude.ai/code/session_01G64z5RJoQmrvytG7poGFQf

---
_Generated by [Claude Code](https://claude.ai/code/session_01G64z5RJoQmrvytG7poGFQf)_

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -40,6 +40,9 @@ jobs:
       - name: Art-request YAML indentation contract
         run: npm run test:art-request-yaml
 
+      - name: GitHub file-read contract
+        run: npm run test:github-file-contents
+
       - name: Art prompt contract
         run: npm run test:art-prompt-contract
 
--- a/package.json
+++ b/package.json
@@ -156,6 +156,7 @@
     "seed:resource-commercial-safe": "tsx scripts/seed_resource_commercial_safe.ts",
     "generate:achievement-art": "tsx scripts/generate_achievement_art.ts",
     "test:art-request-yaml": "tsx utils/scripts/verifyArtRequestYaml.ts && tsx utils/scripts/verifyArtAssetSuggest.ts",
+    "test:github-file-contents": "tsx utils/scripts/verifyGithubFileContents.ts",
     "test:art-prompt-repair": "tsx utils/scripts/verifyArtPromptRepair.ts",
     "test:art-prompt-contract": "tsx utils/scripts/verifyArtPromptContract.ts",
     "test:artjob-sampler-repair": "tsx utils/scripts/verifyArtJobSamplerRepair.ts",
--- a/server/api/conductor/art-request.post.ts
+++ b/server/api/conductor/art-request.post.ts
@@ -7,9 +7,14 @@ import { buildContextualArtPrompt } from '@/server/utils/conductorArtPrompt'
 import { errorHandler } from '@/server/utils/error'
 import {
   appendRequest,
+  requestBlockIds,
   type ArtQueueEntry,
   type ArtVariant,
 } from '@/server/utils/artRequestYaml'
+import {
+  githubPathExists,
+  readGithubFile,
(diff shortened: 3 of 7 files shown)


silasfelinus:
`Contract verifiers` was red on the ESLint ratchet step: `server/utils/githubFileContents.ts:109` declared `content` with `let` but never reassigned it (`prefer-const`, 5 → 6). Pushed a one-line fix (`5c6caed`) switching it to `const`. Verified locally before pushing: `npm run test:lint-ratchet` now reports `-60` (no rule regressed) and `npm run test:github-file-contents` passes all 23 checks. Watching CI on the new head.

---

PR #2322, opened 2026-09-02
Add Rainbow notification preferences and delivery seam

silasfelinus:
## What changed
- adds an additive `RainbowNotificationPreference` table keyed to the canonical Kind Robots User, with three false-by-default opt-ins: agent attention, direct forum reply/mention, and scheduled-agent failure
- adds authenticated Rainbow-first-party GET/PATCH preference APIs; clients cannot nominate another user
- adds `planRainbowNotificationDelivery`, a provider-neutral server seam that resolves whether a class is opted in and, only for a verified account email, returns a generic `EMAIL` target
- deliberately contains no Brevo client, mail send, provider secret, contact sync, or side effect
- adds a focused contract workflow pinning additive migration safety, private defaults, verified-email gating, auth boundaries, and the no-Brevo/no-send rule

## Architecture
This mirrors the existing `RainbowDirectoryPreference` pattern: canonical website state lives in the Kind Robots database and is accessed through a small typed raw-SQL utility. It does not add notification columns to `User` or create a Rainbow-side database.

## Safety
The migration is CREATE TABLE only with a User FK and cascade cleanup; no DROP/ALTER/delete. No email is sent and no production secret or account is required. Brevo activation remains the separate human-gated rainbow-butterflies/t-049.

Conductor: rainbow-butterflies/t-048

### DIFF
--- a/.github/workflows/rainbow-notification-contract.yml
+++ b/.github/workflows/rainbow-notification-contract.yml
@@ -1,0 +1,22 @@
+name: Rainbow Notification Contract
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'prisma/migrations/**rainbow_notification_preferences/**'
+      - 'server/utils/rainbowNotifications.ts'
+      - 'server/api/rainbow/notifications/**'
+      - 'tests/contracts/verifyRainbowNotifications.ts'
+      - '.github/workflows/rainbow-notification-contract.yml'
+
+jobs:
+  verify:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


silasfelinus (commented):
[openai-review-claim session=openai-scheduled-20260902T081053Z-pr2322-review-k7q4 head=dc4672a62567d67e9b2e93d3d7b29e78164dc963]
Reviewed full diff and exact-head checks. Additive preference storage is private-by-default, routes are scoped to the authenticated human, delivery planning requires verified email, and no provider/send side effect is introduced. No blocking threads or defects found. Connector identity owns the PR, so GitHub correctly disallows self-approval; recording the review as a comment instead.

PR #2325, opened 2026-09-02
Mobile panes, whole-tile shelves, visible labels, and readable dream copy

silasfelinus:
Five notes across two review rounds. Two commits.

## Shelf tiles ballooned on mobile

> "Still weird use of whitespace on the facet section on mobile when I scroll."

`fitTiles` rounds to the nearest whole column count. On a 256px phone shelf against the 192px `wide` ideal that rounds to **one** column — a 256px tile, a third over target and most of the screen. The comment directly above that line already rejected this exact move ("flooring would rather grow tiles 40% than fit one more") and then the rounding did the same thing from the other direction.

A tile may now run at most 25% over ideal before fitting another column, guarded so a very narrow track still yields one tile rather than shrinking toward nothing.

```
                       before          after
mobile art shelf    1 × 256px  →   2 × 123px
desktop art shelf   2 × 228px      2 × 228px   (unchanged, ceiling 240)
desktop objects     3 × 149px      3 × 149px   (unchanged, ceiling 160)
```

## Scrolling forever to reach projects and news

> "I don't love scrolling vertically forever to finish my todos before seeing projects and news."

Below `xl`, Needs-you / Building / News are tabs — one pane at a time, defaulting to Needs you. The gate list alone has run past seventy rows, with the other two underneath it.

**A switcher rather than a separate route,** deliberately: same dashboard, and a route change would cost the hero and galleries above it. If real separate screens with their own URLs is what was wanted, that's a different build.

CSS-only below `xl` (`hidden xl:flex` on inactive panes), so the `xl` layout keeps all three stacked, no JS breakpoint listener, no hydration flash.

| | 390×844 | 1920×1080 |
|---|---|---|
| tabs | visible | hidden |
| panes shown | 1 | 3 |
| page height | **844** (= viewport) | 1080 (= viewport) |
| horizontal overflow | 0 | 0 |

## Labels were gated twice

> "we are still missing text labels on the objects when we have room."

The shelf name carried `hidden xl:inline` **on top of** the `showLabel` prop — so the prop the page sets was overruled by exactly the kind of breakpoint the comment right above it says not to use. The prop is the only gate now, as it was always documented to be.

## The label is the see-all link

> "the see all link on the objects should not be there, it can be a link on the text that you will add instead."

The arrow was a second control pointing where the heading already pointed, and on a phone it competed with the chevrons for a strip a few hundred pixels wide. The whole icon-label-count group is the destination now — a far bigger tap target than a 20px glyph.

## Titles grow rather than truncate

> "if the title is larger than two lines, I'd rather we push to take up vertical space than truncate."

No clamp. The caption is `shrink-0` above a `flex-1` plate, so a third or fourth line takes its room from the picture rather than from the name. The plate keeps a `min-h-12` floor so a pathological title can't erase the artwork.

## The dream descriptions were light grey, literally

> "better readable text for the dream creation descriptions, a different color that isn't light grey, it gets lost."

They ran at `text-base-content/65` — a 65% tint of base-content on a pale panel is the lowest-contrast text on the page, under the prose most meant to be read. Now full strength; hierarchy against the title still comes from weight and size.

**Deliberately not a hue** like `text-secondary`: each cast card wears its own record's daisyUI theme, so a hue that reads well on one card is illegible on the next — the trap the badge chip already hit and documented.

Also rewrote the header comment claiming the label is "removed from the screen, not from the accessibility tree" — this change makes that false, and the saving that justified dropping the words (eight shelves × one text line) doesn't apply to six shelves in a 3×2 grid.

## Verification

`prettier` · `eslint` · layout contract · MDC scroll ownership · `vue-tsc`

One note on process: a local typecheck reported failures mid-round that turned out to be a `.nuxt` I'd corrupted myself with a stray `pkill` matching `nuxi` — CI's TypeScript job passed on the identical commit, and a clean rebuild confirmed it.

### DIFF
--- a/components/home/home-page.vue
+++ b/components/home/home-page.vue
@@ -171,7 +171,49 @@
         six short rows and scrolling it would be silly.
       -->
       <div class="flex min-w-0 flex-col gap-2 xl:min-h-0 xl:w-[22%]">
-        <home-attention class="xl:min-h-0 xl:flex-1" />
+        <!--
+          ONE PANE AT A TIME ON A PHONE. Silas, 2026-09-02: "Thinking we should
+          move the news, todo, project section to a different screen on mobile.
+          I don't love scrolling vertically forever to finish my todos before
+          seeing projects and news."
+
+          Stacked, these three are a very long page: the gate list alone has run
+          to seventy-plus rows, and projects and news sit underneath it. On a
+          desktop that is fine because they are a column beside the galleries;
+          on a phone the column unrolls into the page and buries the last two.
+
+          A switcher rather than a separate route: they are the same dashboard,
+          and a route change would cost the hero and galleries above. The tabs
+          are CSS-only below xl -- `hidden xl:flex` on the inactive panes -- so
+          the xl layout keeps all three stacked exactly as before, and nothing
+          here has to know the viewport width in JavaScript.
+        -->
+        <div
+          role="tablist"
+          aria-label="Dashboard sections"
+          class="flex shrink-0 gap-1 kr-panel-flat p-1 xl:hidden"
+        >
+          <button
+            v-for="pane in RIGHT_PANES"
+            :key="pane.key"
+            type="button"
+            role="tab"
+            :aria-selected="activePane === pane.key"
+            class="flex-1 rounded-lg px-2 py-1.5 text-[0.7rem] font-black uppercase tracking-[0.1em] transition-colors focus-visible:outline focus-visible:outline-2 focus-visible:outline-primary"
+            :class="
+              activePane === pane.key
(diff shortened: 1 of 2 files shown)


silasfelinus (commented):
OpenAI scheduled review: full two-file diff inspected at this exact head. No blocking defect found; mobile pane switcher preserves desktop behavior and tile-growth cap is guarded for narrow tracks. Exact-head Contract Tests, TypeScript, Layout Contract, MDC Scroll Ownership, Schema Migration Parity, Generated Client Parity, Project Architecture, Facet Catalog and surrounding contract fleet are completed successfully. Proceeding with guarded merge.

PR #2327, opened 2026-09-02
Show shelf labels, let titles wrap, and stop the dream copy reading as grey

silasfelinus:
Three notes from the review round after #2325. All three turned out to be earlier mistakes of mine rather than judgement calls.

## Labels were gated twice

> "we are still missing text labels on the objects when we have room."

The shelf name carried `hidden xl:inline` **on top of** the `showLabel` prop — so the prop the page sets was overruled by exactly the kind of breakpoint the comment directly above it says not to use. I wrote both the warning and the violation. The prop is now the only gate, as it was always documented to be.

## The label is the see-all link

> "the see all link on the objects should not be there, it can be a link on the text that you will add instead."

The arrow was a second control pointing where the heading already pointed, and on a phone it competed with the chevrons for a strip a few hundred pixels wide. The whole icon-label-count group is the destination now — a far bigger tap target than a 20px glyph, and the row still carries exactly one destination.

## Titles grow rather than truncate

> "if the title is larger than two lines, I'd rather we push to take up vertical space than truncate."

No clamp. The caption is `shrink-0` above a `flex-1` plate, so a third or fourth line takes its room from the picture rather than from the name — the name being the part you're reading. The plate keeps a `min-h-12` floor so a pathological title can't squeeze the artwork out entirely.

## The dream descriptions were light grey, literally

> "better readable text for the dream creation descriptions, a different color that isn't light grey, it gets lost."

They ran at `text-base-content/65`. A 65% tint of base-content on a pale panel is the lowest-contrast text on the page — sitting under the prose that's most meant to be read. Now full strength; hierarchy against the title still comes from weight and size.

**Deliberately not a hue** like `text-secondary`: each cast card wears its own record's daisyUI theme, so a hue that reads well on one card is illegible on the next. That's the trap the badge chip already hit and documented a few lines above.

## Comment correction

The file header claimed the label was "removed from the screen, not from the accessibility tree." This change makes that false, and the saving that justified dropping the words in the first place — eight shelves at one text line each — doesn't apply to six shelves in a 3×2 grid with room to spare. Rewritten to say what the header actually does now.

## Verification

`prettier` · `eslint` · layout contract (holds) · MDC scroll ownership (holds) · `vue-tsc` **0 errors** on a clean `.nuxt`

Note on that last one: a mid-round local typecheck reported ~30 auto-import failures (`Cannot find name 'ref'`) that were entirely a `.nuxt` I'd corrupted myself with a `pkill` matching `nuxi` mid-write. CI passed on the identical commit and a from-scratch rebuild confirmed 0 errors.

---
_Generated by [Claude Code](https://claude.ai/code/session_017aNg3fsHLGUGCwiMp25hkA)_

### DIFF
--- a/components/home/home-attention.vue
+++ b/components/home/home-attention.vue
@@ -69,10 +69,22 @@
 
     <!--
       A bounded scroller, not an unbounded list: the gate count is unpredictable
-      (it has been twenty-plus) and this sits beside a fixed-height hero. The
+      (it has been seventy-plus) and this sits beside a fixed-height hero. The
       layout contract's one-scroll rule deliberately does not count a `max-h-*`
-      region -- nested preview, not the page's scroll owner. `max-h-full` bounds
-      it to the band height the page sets, and keeps the `max-h-*` exemption.
+      region -- nested preview, not the page's scroll owner.
+
+      TWO BOUNDS, because one is not enough. `xl:max-h-full` is right at xl,
+      where this sits in a column with a definite height. Below xl that column is
+      auto-height, so `max-h-full` resolves to "as tall as my content" and the
+      scroller silently stops scrolling -- the list simply unrolls into the page.
+
+      Measured on a tablet at 1180x820 (2026-09-02): this rendered 3542px tall
+      with `scrollHeight === clientHeight`, and the page's real scroll owner
+      reported 4333px of content in a 722px viewport. Six screenfuls, almost all
+      of it gates. Silas: "We have three screens, a MASSIVE news feed." The
+      explicit 26rem is about six gate rows -- enough to work through, short
+      enough that projects and news stay on the same screen. With it, the same
+      page measures 1133px.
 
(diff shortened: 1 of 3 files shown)


silasfelinus (commented):
[openai-review-claim session=openai-scheduled-20260902T101319Z-pr2327-r9c4 head=fd40afe3efbd56f83b25e3cec2cb332f626d6e4a]

Reviewed the full three-file diff and exact-head Actions. The changes match the latest human direction: shelf labels are controlled by the host prop instead of a contradictory breakpoint, the label/count group becomes the single see-all target, object titles get bounded extra vertical room, dream prose regains full contrast, and the attention list receives a real sub-xl height bound. No blocking defect found. Exact-head Contract Tests, TypeScript, Layout Contract, MDC Scroll Ownership, Schema Migration Parity, Generated Client Parity, Project Architecture, Facet Catalog, and the surrounding contract fleet are all completed successfully.

PR #2344, opened 2026-09-02
Add the canonical Rainbow dashboard workspace feed

silasfelinus:
Conductor: rainbow-butterflies/t-027 current-v2 completion, lane 2 backend.

Adds one narrow private first-party endpoint for the signed-in Rainbow human:

- bounded recent direct public forum replies to posts accountable to this human or their agents;
- bounded recent canonical ArtImage, Character, and Project records owned by the same userId;
- existing userId accountability naturally includes agent-created work without a Rainbow ownership table;
- operator maturity filtering and active-record filtering;
- human-or-first-party-delegated auth only, `Cache-Control: no-store`;
- explicit semantics that this is recent activity, not an unread/mention inbox.

This does not duplicate mission metrics or agent attention requests. Rainbow already has canonical BFFs for those. It also does not read private Conductor data, create new schema, create read receipts, infer mentions, or add a second object store.

A dedicated Rainbow Dashboard Contract pins the auth, privacy, direct-reply, ownership, maturity, and bounded-feed invariants.

### DIFF
--- a/server/utils/rainbowDashboard.ts
+++ b/server/utils/rainbowDashboard.ts
@@ -154,8 +154,8 @@ async function loadRecentObjects(input: { userId: number; includeMature: boolean
       detail: clip(item.promptString || item.artPrompt, 280),
       createdAt: item.createdAt,
       updatedAt: item.updatedAt,
-      isPublic: item.isPublic,
-      isMature: item.isMature,
+      isPublic: Boolean(item.isPublic),
+      isMature: Boolean(item.isMature),
       imagePath: item.thumbnailPath || item.cardPath || item.imagePath,
     })),
     ...characters.map((item) => ({
@@ -165,8 +165,8 @@ async function loadRecentObjects(input: { userId: number; includeMature: boolean
       detail: clip([item.species, item.role].filter(Boolean).join(' · '), 280),
       createdAt: item.createdAt,
       updatedAt: item.updatedAt,
-      isPublic: item.isPublic,
-      isMature: item.isMature,
+      isPublic: Boolean(item.isPublic),
+      isMature: Boolean(item.isMature),
       imagePath: item.iconPath || item.cardPath || item.imagePath,
     })),
     ...projects.map((item) => ({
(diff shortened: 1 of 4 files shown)


server/utils/rainbowDashboard.ts
```diff
@@ -0,0 +1,220 @@
...
+  const objects: DashboardObject[] = [
+    ...artImages.map((item) => ({
+      kind: 'ART_IMAGE' as const,
+      id: item.id,
+      label: clip(item.artPrompt || item.promptString, 120) || `ArtImage #${item.id}`,
+      detail: clip(item.promptString || item.artPrompt, 280),
+      createdAt: item.createdAt,
+      updatedAt: item.updatedAt,
+      isPublic: item.isPublic,
+      isMature: item.isMature,
```
silasfelinus: The `TypeScript` check is red on this PR (reproduced locally with `npm run test` / vue-tsc against this exact commit):

```
server/utils/rainbowDashboard.ts(149,9): error TS2322: Type '({ kind: "ART_IMAGE"; ...; isPublic: boolean | null; isMature: boolean | null; ... } | { ... } | { ... })[]' is not assignable to type 'DashboardObject[]'.
  Types of property 'isPublic' are incompatible.
    Type 'boolean | null' is not assignable to type 'boolean'.
```

Root cause: `ArtImage.isPublic` and `ArtImage.isMature` are declared `Boolean?` (nullable) in `prisma/schema.prisma`, unlike `Character`/`Project` where both are non-nullable `Boolean @default(...)`. Only the `artImages.map(...)` branch (this block, lines 157-158) produces `boolean | null`, which doesn't satisfy `DashboardObject`'s `isPublic: boolean` / `isMature: boolean`.

Suggested fix — coerce with the same defaults the schema declares for this field:
```ts
isPublic: item.isPublic ?? true,
isMature: item.isMature ?? false,
```

Everything else in the PR (auth guard, maturity filtering, `no-store`, contract test) looks correct and matches the stated scope — this is the only blocker.

---

silasfelinus (commented):
Reviewed rainbow-butterflies/t-027 lane 3 (dashboard workspace feed). The design and scope are good — bounded, user-owned, no shadow database, explicit "not an inbox" semantics, contract test covers the key invariants. But the `TypeScript` check is genuinely red (not flaky/infra): a real type error in `server/utils/rainbowDashboard.ts` (see inline comment). Reproduced locally against this exact commit (5798aeb) with `npm run test`; all other checks pass. Not merging until this is fixed — the one-line coercion below should get it green.

---

PR #2351, opened 2026-09-03
fix(deploy): make Alexandria migrations automatic

silasfelinus:
## What this fixes

Tonight the art queue stalled because production was ten Prisma migrations behind while the Kind Robots container was otherwise reachable. The missing `Krea2DeferredFreeJob` table made every relay `/claim` return HTTP 500.

The underlying problem was the self-hosted deploy handoff: GHCR publication, Unraid container update, and database migration were separate steps, and neither ChatGPT nor Claude agents had a standing contract that made production migration state part of "done."

## Changes

- adds `scripts/deploy-unraid.sh` as the guarded Alexandria deploy path:
  - serializes runs with `flock`
  - pulls the GHCR image
  - runs that image's pending migrations with the isolated `kindrobot_migrate` credential
  - only then invokes Unraid DockerMan's `update_container` using the saved template
  - persists migration verification state under appdata
  - waits for container health
- adds `scripts/unraid-user-script.sh` as the tiny launcher intended for Unraid User Scripts
- documents one-time User Scripts setup and a 5-minute schedule
- adds an all-agent contract in `AGENTS.md`: merged != published != deployed != schema verified
- requires unattended migrations to be backward-compatible with the old build during migration-before-container-swap; destructive schema contractions must be staged
- rewrites the migration credential runbook so it no longer tells operators to manually migrate and then Force Update
- upgrades `/api/health/database` so Docker health fails when migrations are pending/failed instead of treating `SELECT 1` as sufficient
- extends the existing migrate-on-deploy contract test to cover the real Alexandria/User Scripts path and schema-aware health

## Safety

- `MIGRATION_DATABASE_URL` stays out of the long-running application container
- migration failure leaves the old application container running
- no reset/drop behavior is added
- normal deployment still uses Unraid DockerMan's saved template rather than reconstructing container arguments independently

## One-time operator step after merge

On Alexandria, add a User Script named e.g. `Kind Robots Auto Deploy` containing:

```bash
#!/bin/bash
exec /bin/bash /mnt/user/appdata/kind_robots/scripts/unraid-user-script.sh
```

and schedule it every 5 minutes. Before adding it, pull `main` once and run the launcher manually to verify the host wiring.


silasfelinus:
OpenAI scheduled reviewer claim. Inspecting exact head `1f2e6e634fbf574490f5ffca4af22c13f7d48c74`, full diff, threads, and required checks before any merge.

Issue #2345, opened 2026-09-03
Rainbow t-027 lane 3: add a canonical AgentProfile messaging substrate


silasfelinus:
PRODUCT-V2 / Rainbow Butterflies t-027 lane 3 follow-up.

The current code has the right opt-in signals but not a safe bidirectional transport:

- `AgentProfile.allowMessages` exists and is exposed through the Rainbow directory/profile APIs.
- `RainbowDirectoryPreference.allowMessages` exists for humans.
- `AgentNote` is a durable human -> AgentProfile one-way inbox delivered on the next check-in.
- Generic `Chat` supports `ToUser` / `recipientId`, but it is user-centric and lets clients provide sender text. It has no AgentProfile-bound sender/recipient identity, AgentProfile permission model, or profile-scoped inbox semantics.

Do **not** expose the generic Chat API as Rainbow agent messaging and do not add a Rainbow-only message store.

Build the smallest canonical Kind Robots substrate that can support optional human <-> AgentProfile messaging safely. Requirements:

1. Honor both human and AgentProfile `allowMessages` opt-in before accepting a new conversation/message.
2. Derive the accountable human and AgentProfile identity server-side from the authenticated session / bound AgentCredential. Never accept arbitrary sender identity.
3. Keep website state canonical in Kind Robots. Rainbow should only use a BFF later.
4. Bound message length, list sizes, and history pagination. Add conservative rate limits and moderation/maturity/privacy checks.
5. Keep agent credential scopes explicit. Do not silently widen existing credentials or turn MCP into a generic message proxy.
6. Preserve liaison/owner accountability: external AgentProfile messages remain attributable to the human operator.
7. Define clear delivery/read semantics rather than pretending generic Chat `isRead` is already a profile inbox.
8. No Brevo dependency. Email remains notification transport only, not message state.
9. No provider accounts, provider secrets, external posting, or Rainbow production publication in this backend task.
10. Add focused contract tests for opt-in, sender binding, recipient ownership/existence, revocation/scope enforcement, rate limiting, and cross-user privacy.

This issue is the explicit architecture split allowed by PRODUCT-V2: optional messaging should stay absent from Rainbow until a safe canonical substrate exists, rather than faking the feature with a parallel inbox.

silasfelinus:
[...]

PR #2354, opened 2026-09-03
Add canonical AgentProfile messaging

silasfelinus:
Implements #2345 as the canonical Kind Robots substrate for optional human ↔ AgentProfile messaging.

This deliberately does not reuse legacy `Chat` and does not add Rainbow-owned message state.

Adds:
- additive `AgentMessageThread` + `AgentMessage` migration, one thread per human/AgentProfile pair;
- server-derived human vs AgentProfile actor identity and audit fields for the accountable operator / credential;
- explicit `agent:message` credential scope, not present in default credential scopes;
- both-side public + `allowMessages` opt-in on every new write;
- existing restricted-account checks plus the canonical CHILD maturity restriction for both the human and AgentProfile operator;
- 5,000-character message cap, 30 writes / 10 minutes rate limit, bounded thread/history lists, cursor pagination;
- idempotent `clientKey` writes;
- recipient-oriented durable read timestamps;
- participant-scoped thread access returning 404 across user/profile boundaries;
- shared REST route family usable by signed-in humans / Rainbow first-party delegation and profile-bound agent credentials;
- focused AgentProfile messaging contract added to the existing Agent Check-in CI lane.

Important boundaries:
- opt-out blocks new messages but does not erase or strand prior history;
- MCP remains the same narrow two-tool bridge and gains no messaging proxy;
- no Brevo/email message state;
- no provider account, provider secret, external posting, or Rainbow publication;
- no silent widening of existing credentials.

**Schema-affecting:** yes. The migration is additive and backward-compatible with the currently running build during the migration-before-container-replacement window described in `AGENTS.md`.

Merge is not deployment. Production/Alexandria schema state must still be verified separately after the normal deployment path runs.

### DIFF
--- a/.github/workflows/agent-checkin-contract.yml
+++ b/.github/workflows/agent-checkin-contract.yml
@@ -7,16 +7,20 @@ on:
       - 'prisma/agent-profile.prisma'
       - 'prisma/migrations/**agent_checkins_notes/**'
       - 'prisma/migrations/**agent_attention_requests/**'
+      - 'prisma/migrations/**agent_profile_messages/**'
       - 'utils/agentCredentialScopes.ts'
       - 'server/utils/authGuard.ts'
       - 'server/utils/agentAttentionRequests.ts'
       - 'server/utils/agentProfileRuntime.ts'
       - 'server/utils/agentWorkingContext.ts'
+      - 'server/utils/agentMessaging.ts'
+      - 'server/utils/agentMessagingPolicy.ts'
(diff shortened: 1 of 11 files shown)


silasfelinus:
OpenAI scheduled reviewer claim for exact head `f36bf81d7ae5bd823333fff28c57a7d5c764e131`. Inspecting full diff, threads, mergeability, and exact-head checks before any merge.

PR #2359, opened 2026-09-03
model-builder/t-029: skip comments/strings in guard-script brace scanner

silasfelinus:
### Task
model-builder/t-029 cycle 88 — kaizen from OpenAI cycle 87's guard-integrity audit, which flagged (but did not yet reproduce or fix) a parser blind spot in the shared brace-matching scanner used by ~40 `verifyModelBuilder*.ts` guard scripts.

### What changed
`extractFunctionBodies()` / `findGuardIntervals()` in `verifyModelBuilderCompletionGate.ts` brace/paren-matched by scanning raw source characters, so a stray `{`/`}`/`(`/`)` inside a `//`/`/* */` comment or a string/template literal was counted as real structure.

Confirmed both failure modes before touching the scanner (per cycle 87's own instruction — reproduce first, then harden only if the fixture reproduces the failure):
- A `}` inside a comment silently **truncated** a function's extracted body, so a real violation in the truncated-away tail would go undetected — a false negative in every guard built on `extractFunctionBodies`.
- A `{` inside a string literal made the scan **overrun** looking for a closing brace that was never coming, throwing `"has ...'s shape changed?"` on a file whose shape hadn't actually changed.

Added `computeCodeMask()` to mark which offsets are genuine code (not comment/string content) and wired both brace/paren counters in `extractFunctionBodies` and `findGuardIntervals` to skip masked offsets. Added a regression fixture in `verifyModelBuilderCompletionGate.test.ts` covering both failure modes plus a direct spot-check of `computeCodeMask()` itself.

### How I verified
- Reproduced both bugs standalone against the pre-fix code before making any scanner change.
- All 167 `test:model-builder-*` npm scripts pass (covers every guard script built on this shared module).
- `npm run test` (vue-tsc) — clean.
- `eslint` on both touched files — clean.
- `npm run test:lint-ratchet` — holds (332 problems, -60, no rule regressed).
- `npm run test:layout-contract` — holds (207-violation baseline unchanged; no layout touched).
- `prettier --check` — clean (formatted the new code to match; the pre-existing files were already prettier-clean).

### Stakes
reversible

https://claude.ai/code/session_01KyqbAz6fgEHMsH1SVS9ThJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01KyqbAz6fgEHMsH1SVS9ThJ)_

### DIFF
--- a/utils/scripts/verifyModelBuilderCompletionGate.test.ts
+++ b/utils/scripts/verifyModelBuilderCompletionGate.test.ts
@@ -9,9 +9,18 @@
 // not be flagged), and a synchronous function with no `await` at all (must
 // not be flagged even though it writes item.stages unconditionally, matching
 // approveStage/rejectStage/reopenStage's real shape).
+//
+// Also covers the comment/string-brace parser blind spot (model-builder/t-029
+// cycle 88, kaizen from OpenAI cycle 87's guard-integrity audit): before
+// computeCodeMask existed, a `}` inside a `//` comment silently truncated a
+// function's extracted body (a real violation inside the truncated-away tail
+// would go undetected -- a false negative), and a `{` inside a string
+// literal made the brace scan overrun looking for a closing brace that was
+// never coming, throwing on a file whose shape hadn't actually changed.
 import assert from 'node:assert/strict'
 
 import {
+  computeCodeMask,
   extractFunctionBodies,
   findGuardIntervals,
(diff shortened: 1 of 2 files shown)


silasfelinus:
Review claimed for exact head d1a50c976fa854df8d474d59bc940e65962f4b4b.

PR #2414, opened 2026-09-05
fix(art): make Krea prompts semantic-only and repair tainted entity art

silasfelinus:
## Root cause

Krea 2 is caption-conditioned, not an instruction-following image model. The Facet v2/v3 producer was the loudest offender because it fed phrases such as `Illustrate the Facet concept ...` and `Create ... for Kind Robots ...` directly into positive conditioning, which explains the reported logo/headings/prompt prose painted into Surreal Horror.

The sibling audit found the same underlying problem across the shared entity-art path. Character, Bot, Dream, Scenario, Reward, and Facet recreation all pass through `buildEntityArtPrompt`, which intentionally carries database context for engines that can use it: `Name:`, `Title:`, `Species:`, `Personality:`, `Backstory:`, `Rarity:`, etc., plus `Compose this as ... for the following character/dream/...`. Feeding that whole block to Krea makes those labels and values visual concepts too.

## What changed

- Replace Facet v2/v3 prompt construction with semantic-only v4 conditioning: subject/concept prose, visual language, composition, medium, lighting, and texture. No `Facet`, app-purpose, taxonomy-label, or `Kind Robots` wrapper text is injected.
- Preserve genuinely curated `Facet.artPrompt` values, while recognizing and replacing only the exact v2/v3 generated wrapper persisted by the old producer.
- Add a Krea-specific prompt-contract guard for the known contextual wrappers so old queued v2/v3 jobs are rejected before cleanup reaches them.
- Add `utils/kreaSemanticPrompt.ts` and enforce it inside the shared Krea workflow builder. Every future Krea workflow now strips entity/database labels, app-purpose prose, prompt-writing instructions, and positive-conditioning text-exclusion vocabulary while preserving useful visual values and composition. Raw request wording is retained only in Comfy `_meta.request_prompt` provenance and is not connected to CLIP conditioning.
- This shared Krea boundary covers Facet, Character, Bot, Dream, Scenario, Reward, and any other producer that builds a Krea workflow, without weakening contextual prompts for non-Krea engines.
- Add `--repair-tainted` to the Facet producer: provenance-scan `facet-multi-art-krea2-v2` / `facet-coverage-krea2-v3`, cancel tainted pending jobs, preserve outputs already superseded by a different ArtImage, and enqueue semantic v4 replacements for legacy jobs that still own their Facet slot.
- Add `scripts/repair_krea_context_entity_art.ts` for the sibling main objects. It scans Krea entity-art jobs for contextual conditioning, preserves completed ArtImages, skips slots already superseded by newer art, cancels only still-pending tainted jobs, and submits semantic `NEW_OUTPUT` replacements for affected Character/Bot/Dream/Scenario/Reward/Facet targets.
- Keep the specialized Facet repair separate from the generic entity repair so v2/v3 provenance gets the stricter catalog-aware handling.
- Extend regression coverage with the exact Surreal Horror failure plus a Character entity prompt carrying Name/Species/Class/Role/Personality/Backstory labels. Tests assert that useful visual concepts survive while labels/app prose never reach the positive CLIP node.
- **Fix (pass 3):** `scripts/generate_facet_art_v4.ts`'s `--repair-tainted` write path previously cancelled legacy PENDING v2/v3 ArtJobs and rewrote `Facet.artPrompt` *before* building/validating the full replacement `jobRows` set via `queue.map(buildFacetArtPayload(...))`. Since `buildFacetArtPayload` synchronously throws via `assertArtPromptContract` on an invalid prompt, a single bad entry could abort the run after the cancellation/prompt writes had already committed, stranding that facet with no active ArtJob and no replacement. `jobRows` is now built and validated first; the cancellation and prompt-update writes only run once the full replacement set is known to build successfully.
- The one-shot main-push repair workflow (`.github/workflows/facet-art-prompt-repair-v4.yml`) has been **removed** rather than expanded — it would have mutated production ArtJob/Facet data unattended on merge, which is a hard human gate under the current safety contract. The repair scripts (`generate_facet_art_v4.ts --repair-tainted`, `repair_krea_context_entity_art.ts`) remain in the repo; running them against production is a separate, explicitly human-approved action after this PR merges.

## Safety

No existing ArtImage is deleted. DONE jobs are resubmitted only when they still appear to own the current entity-art slot; a different current per-slot ArtImage ID is treated as a newer/superseding image and left alone. Pending tainted jobs are cancelled only when safe to replace or already superseded. RUNNING jobs are not killed mid-render; they receive a semantic replacement. Replacement jobs use `NEW_OUTPUT` and preserve original art/history. The repair write path now validates before it mutates (see pass-3 fix above).

## Verification

Facet Catalog Contract is green on final-head CI, including Facet generation/artwork/gallery contracts. TypeScript and the full Contract Tests suite are the remaining execution gates on this expanded head; merge only after every triggered final-head check has completed successfully. `npm run test` (vue-tsc, full project) passes clean on the pass-3 fix; `npx eslint scripts/generate_facet_art_v4.ts` shows only 4 pre-existing, unrelated `no-useless-escape` findings at line 276 (not touched by this change).

Production repair is intentionally **not** automated by this PR. After merge, running `generate_facet_art_v4.ts --repair-tainted --write` (and the sibling `repair_krea_context_entity_art.ts`) against production is a separate, human-approved step; record the Facet and sibling replacement ArtJob IDs when that run happens.

Conductor: `dream-cycle/t-006`, pass 3.


silasfelinus:
Reviewed this against `origin/main` (`d6202f9`). The semantic v4 prompt construction and the new Krea-scoped `contextual-wrapper` contract rule are well-targeted and correctly reproduce/catch the reported failure — confirmed no `Facet`/app-purpose/`Kind Robots` wrapper text reaches `buildKrea2WorkflowFromRequest`, and no `ArtImage` is ever deleted (`preserveOriginal: true` keeps history, never hits the project-only delete-on-replace path).

**Blocking: the repair path cancels legacy jobs before the step that can throw, with no rollback.**

In `scripts/generate_facet_art_v4.ts`, inside the `if (WRITE)` block:
1. `prisma.artJob.updateMany(...)` (~line 834) cancels every tainted PENDING v2/v3 job catalog-wide — commits immediately.
2. `facet.artPrompt` updates commit (~line 849 area).
3. `const jobRows = queue.map(...)` (~line 867) is a synchronous, eager `.map()` that calls `buildFacetArtPayload`, which calls `assertArtPromptContract` (line 373) — **this throws** on any contract violation, in *any* queued entry, not just repaired ones.

If that `.map()` throws, `main()` rejects before `prisma.artJob.createMany` (~line 894) ever runs. By then step 1's cancellations are already committed. Net effect: every facet whose tainted job got cancelled is left with **no active ArtJob at all** until someone notices the failed one-shot workflow run and manually re-invokes `--repair-tainted`. Given this runs unattended via `facet-art-prompt-repair-v4.yml` immediately on merge to `main`, a single bad prompt anywhere in the queue (including one unrelated to the repair — e.g. an ordinary coverage entry that happens to trip a pre-existing contract rule) can strand every repaired facet with a cancelled-and-not-replaced ArtJob.

Suggested fix: build `jobRows` (and validate each payload) *before* committing the cancellation/prompt updates, or wrap each entry's `buildFacetArtPayload` call in try/catch and skip+report the offending entry rather than aborting the whole batch. Either keeps the destructive step from committing ahead of the step that can fail.

Two smaller, non-blocking notes for awareness (not asking for changes on these):
- `likelyStillCarriesLegacyOutput` returns "superseded, leave alone" whenever a legacy DONE job's `artImageId` is falsy but the facet's current slot is populated, with no logging. Every normal completion path sets `artImageId`, so this should be unreachable in practice, but it's a silent no-op if it ever isn't.
- The repair loop's PENDING-job skip decision is taken from a snapshot at the top of `main()`; a job that gets claimed and starts running between that read and the repair decision could end up duplicated (self-healing on the next repair run, narrow window).

Given the production-data-mutating, unattended nature of the one-shot workflow, I'd like the ordering issue addressed (or explicitly acknowledged with a documented manual-recovery step) before this merges.

---

silasfelinus (commented):
[review-claim: openai-scheduled-20260905T101421Z-pr2414-review-k3m8]

Reviewed the current exact head `1c14237427eb875504480c4d451745f6cfdb533f`. All 45 PR-triggered workflows on this head have completed successfully, including TypeScript, Contract Tests, and Facet Catalog Contract. The earlier cancellation-before-validation blocker is still present in `scripts/generate_facet_art_v4.ts`: `updateMany(... CANCELLED)` and facet prompt writes commit before `queue.map(buildFacetArtPayload(...))`, whose prompt contract assertion can throw before replacement ArtJobs are created. That remains a merge blocker because a failed one-shot run can strand cancelled jobs.

There is also a separate hard human gate under the current Conductor/Kind Robots safety contract: `.github/workflows/facet-art-prompt-repair-v4.yml` intentionally mutates production ArtJob/Facet data immediately when this file lands on `main`. Even after the ordering defect is repaired and fresh exact-head checks are green, I will not merge the workflow-triggering PR without Silas explicitly approving this concrete production repair run. A safe split is acceptable: merge the semantic prompt/code/tests without the one-shot main-push workflow, then keep the production repair invocation gated separately.

silasfelinus (commented):
Re-reviewed against current head (`2c23451`), following up on the earlier blocking comment (https://github.com/silasfelinus/kind_robots/pull/2414#issuecomment-5550937934, conductor `dream-cycle/t-006` pass 1 rejection).

**The flagged ordering bug is still present.** `scripts/generate_facet_art_v4.ts` has not changed since that comment (only touched by the very first commit on this PR) — the `--repair-tainted` path still does, in this order, inside the `WRITE` block:
1. `prisma.artJob.updateMany(...)` cancels every tainted PENDING v2/v3 job — commits immediately.
2. `facet.artPrompt` updates commit.
3. `const jobRows = queue.map(...)` — a synchronous, eager `.map()` calling `buildFacetArtPayload` → `assertArtPromptContract`, which throws on any queued entry (not just repaired ones) — *before* `prisma.artJob.createMany` ever runs.

If step 3 throws, steps 1–2 have already committed: every facet whose tainted job was cancelled is left with no active ArtJob, silently, until someone notices.

**What changed instead:** the newest commit (`fix(art): keep production Krea repair human-gated`) deletes `.github/workflows/facet-art-prompt-repair-v4.yml` outright, so the repair no longer runs unattended on merge. That's a real, useful mitigation of the "unattended against production data" aggravating factor — but it isn't the fix the prior review asked for, and it isn't mentioned anywhere in the PR body (whose Verification section still describes the now-deleted auto-workflow running "after merge"). The underlying code defect fires identically whether a human or a cron triggers it; a human just happens to be present to notice the failure this way. Re-running the plain (non-`--repair-tainted`) coverage pass afterward *would* self-heal a stranded facet (missing display art re-enters the normal `queue.push` path since detection keys off `facet[variant.field]` being empty, not job status) — but that recovery path isn't documented anywhere, so a future operator has no way to know it's safe/expected without re-deriving this from the code.

Please either:
- build and validate all `jobRows` (or try/catch each `buildFacetArtPayload` call and skip+report the offending entry) *before* committing the cancellation/prompt updates, per the original ask, or
- explicitly document the manual-recovery step (a plain re-run repairs any stranded facet) in the script's own comments and the PR body, so removing the auto-workflow is a documented mitigation rather than an incidental side effect.

Everything else from the first pass still holds: semantic v4 conditioning is correctly scoped, the Krea contextual-wrapper contract rule is well-tested, and no `ArtImage` is ever deleted. This is pass 2 — tracked as `dream-cycle/t-006` in Conductor.
_Session: https://claude.ai/code/session_01ArQxaxuCFPc1jx3xm92Db9_

---

silasfelinus (commented):
[openai-scheduled-20260905T111131Z-pr2414-review-b7q4] Exact-head re-review after CI completion: all 45 PR workflows are now completed successfully, including Contract Tests, TypeScript Type Check, Facet Catalog Contract, schema/client parity, and layout/narrative contracts. The merge blocker is still the repair transaction ordering in `generate_facet_art_v4.ts`: under `WRITE && REPAIR_TAINTED`, legacy PENDING jobs are cancelled and Facet.artPrompt rows are updated before `queue.map(...buildFacetArtPayload...)` validates/builds every replacement. If a later payload throws, the run can still leave cancelled jobs or prompt mutations without replacement jobs. Please build/validate the complete `jobRows` replacement set first, before any cancellation or Facet update; only after that succeeds should the mutation phase begin. The production repair execution remains separately human-gated even after the code is merge-safe.

silasfelinus:
Pushed the pass-3 fix for the ordering bug flagged in both prior review passes (https://github.com/silasfelinus/kind_robots/pull/2414#pullrequestreview-5120880138, https://github.com/silasfelinus/kind_robots/pull/2414#pullrequestreview-5120983302): `jobRows` is now built via `queue.map(buildFacetArtPayload(...))` — which synchronously validates every entry via `assertArtPromptContract` — *before* the legacy-job cancellation (`prisma.artJob.updateMany`) or `Facet.artPrompt` writes run, not after. If any entry fails to build, the run now aborts before touching either write, so a facet can no longer be left with a cancelled job and no replacement.

Also updated the PR body's Verification section, which still described the now-removed auto-run repair workflow — it now says plainly that production repair is a separate, human-approved step post-merge, matching what pass 2's review already flagged as the actual mitigation in place.

Verified locally: `npm run test` (vue-tsc, full project, `--noEmit`) exits clean; `npx eslint scripts/generate_facet_art_v4.ts` shows only 4 pre-existing `no-useless-escape` findings at line 276, unrelated to this change (present on the unmodified head too).

---

silasfelinus (commented):
[review-claim: openai-scheduled-20260905T121433Z-pr2414-review-f3a7]

Taking the current exact-head review after the pass-3 transactional ordering fix. I will verify the full relevant diff, unresolved threads, mergeability, and every final-head check before any merge.

silasfelinus (commented):
[review-claim: openai-scheduled-20260905T121433Z-pr2414-review-f3a7]

Pass-3 fixes the Facet repair ordering correctly, and all 45 exact-head workflows are green. One sibling blocker remains in the new generic `scripts/repair_krea_context_entity_art.ts`: for a tainted `PENDING` job, the `--write` path commits `status: CANCELLED` and only afterward calls `prisma.artJob.create(...)`. If that create fails, the source job is stranded cancelled with no replacement, which is the same failure class we just fixed in `generate_facet_art_v4.ts`.

Please make the PENDING cancel + replacement create atomic, preferably with `prisma.$transaction([artJob.update(...), artJob.create(...)])`, while leaving RUNNING/DONE replacement creation unchanged. The `slotOccupied` PENDING cancellation is intentionally replacement-free and is not this finding. Production execution remains separately human-gated; this comment is about making the repair tool itself safe before merge.

silasfelinus (commented):
[openai-scheduled-20260905T131232Z-pr2414-atomic-k8v4] Fixed the remaining review blocker on the existing branch: replacement-required PENDING jobs now perform cancellation and replacement ArtJob creation inside one Prisma transaction, so a failed create rolls the cancellation back. The intentionally replacement-free cancellation for an already-populated slot remains separate. Also removed an unused semantic-prompt import. Fresh exact-head CI is running; do not merge until the required runs complete successfully. Production repair execution remains a separate human gate.

PR #2441, opened 2026-09-06
art: normalize Bot's primary and crop a stand-in primary correctly

silasfelinus:
First slice of the entity-art collapse: four stored slots per object (`imagePath` + `cardPath` + `heroPath` + `iconPath`) become **one primary render plus the inspiration gallery** `EntityArtImage` already provides.

This slice is plumbing only. An object that has purpose-built variants renders byte-identically to before — the priority inversion that actually retires the variants comes next, and lands safely because the crop behaviour is already in place.

## Bot normalization

Bot stores its render in `avatarImage` while every other entity uses `imagePath`, so every bot surface special-cased its own primary and a bot without one rendered blank. `resolveArtImageSrc` now resolves `imagePath → avatarImage → path`.

No schema rename was needed: `Bot.imagePath` already exists (3 of 71 bots have it populated, 64 have only `avatarImage`, and where both are set the values are identical). `imagePath` still wins when set, so backfilling it later needs no code change.

## Crop correctness

New `resolveArtVariantSource()` reports whether a source is a **composed variant** or a **primary standing in** for a missing one. `resolveArtVariantSrc()` becomes a thin wrapper over it, so every existing caller is untouched.

`kr-art-plate` applies `ART_VARIANT_FOCUS` **only to a stand-in primary**. That distinction matters: a purpose-built variant is composed for its frame (the card prompt asks for "breathing room around the focal subject", the hero for "the subject safely inside the center region"), so re-cropping it would fight the composition. A square primary cropped to 2:3 or 16:9 by `object-cover` is the opposite case — a centred crop reliably cuts the subject's head off, so those get an above-centre anchor. A new `position` prop art-directs any single surface.

## Verification

- `utils/scripts/verifyEntityArtPrimaryCrop.ts` covers both behaviours (Bot primary resolution and precedence; origin reporting for variant / primary / fallback / none; the focus anchors all sitting above centre; the string wrapper staying byte-compatible). Wired into the Contract verifiers job.
- `vue-tsc --noEmit` clean, `eslint` clean on all changed files, layout contract holds, lint ratchet holds (333, −59).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GoXLrc2yHnkZtNBY4b8zSX

---
_Generated by [Claude Code](https://claude.ai/code/session_01GoXLrc2yHnkZtNBY4b8zSX)_

PR #2443, opened 2026-09-06
art: primary render wins, card/hero/icon retired, old renders become inspiration

silasfelinus:
Three steps of the entity-art collapse: flip what gets displayed, stop producing the slots nobody displays any more, and keep every render already made — as inspiration.

## 1. Primary first

`resolveArtVariantSource` now degrades **primary → variant path → variant base64 → fallback**, inverting the variant-first order that made every entity cost four renders for no display benefit — the frames differ, but the picture does not need to.

The variant branch stays as a *fallback* rather than being deleted, so nothing goes blank mid-migration: the ~68 of 910 entities that have variant art but no primary keep rendering exactly as before.

Cards, heroes and icons now show the primary, cropped by `kr-art-plate` with the above-centre `ART_VARIANT_FOCUS` anchors from #2441 — which is why those shipped first, so a square portrait in a 2:3 or 16:9 frame keeps its subject instead of being centre-cropped through the head. The anchors are a starting point and easy to tune; `position` art-directs any single surface meanwhile.

## 2. Retire the slots

Two producers, gated separately because only one goes through the API:

- `EntityArtFieldConfig` gains `retired`, and all **17** card/hero/icon slots across bot/character/scenario/reward/facet/project are marked. `prepareEntityArtEnqueue` refuses them.
- The gate sits at **enqueue**, deliberately *not* in `getEntityArtFieldConfig`: completion, reads and history resolve the same config, so blocking there would strand queued jobs and hide art still on the page.
- `generate_facet_art_v4` writes ArtJobs straight to the table and inherits no API gate, so `--all-variants` now skips retired fields itself. The entries stay in `ART_VARIANTS` because repair still maps a legacy job's field onto its variant geometry.

## 3. Old renders become inspiration

The ~1,820 card/hero/icon images already rendered are kept, as inspiration. Mechanically that is one join row each: `listEntityArtHistory` reads `EntityArtImage` (t-079: *"one join row is how an ArtImage becomes visible in an entity's history … not ArtImage.path prefixes"*), so inserting the link is the whole job and the ArtImage is never copied or moved.

`scripts/link_entity_art_slots_as_inspiration.ts` is **insert-only** and dry-run by default. It dedupes against links the repair's `preserveOriginal` archiving already made today, skips dangling FKs rather than aborting a clean run, and *reports* path-only slots (a bare path with no ArtImage row to join to) rather than inventing rows for them. Dream is deliberately absent — its secondary slots carry no ArtImage FK and none are populated.

**No column is cleared here.** Clearing waits until every affected image is provably linked and retagged, because `ArtImage.path`'s entity tag is what keeps an image out of the unfiled landing zone the triage pass deletes. That ordering is the difference between "turned into inspiration" and "accidentally reaped".

The cleanup workflow gains a `task` input so job cancellation and inspiration linking share one dispatch-only, dry-run-default surface.

## Verification

`verifyEntityArtPrimaryCrop.ts` pins the precedence (primary beats all three variants; a Bot's `avatarImage` counts as primary; a stored variant still wins and still reports `origin: 'variant'` when there is no primary), both producer gates, the count of 17, and that no primary slot is ever marked retired.

`vue-tsc --noEmit` clean, `eslint` clean on changed files, Facet catalog maintenance contract passes, layout contract holds, lint ratchet holds.

https://claude.ai/code/session_01GoXLrc2yHnkZtNBY4b8zSX

### DIFF
--- a/.github/workflows/entity-art-slot-cleanup.yml
+++ b/.github/workflows/entity-art-slot-cleanup.yml
@@ -27,8 +27,15 @@ name: Entity art slot cleanup
 on:
   workflow_dispatch:
     inputs:
+      task:
+        description: 'cancel-pending-jobs clears queued work for retiring slots; link-inspiration turns rendered card/hero/icon art into inspiration entries'
+        type: choice
+        options:
+          - cancel-pending-jobs
+          - link-inspiration
+        default: cancel-pending-jobs
       mode:
-        description: 'dry-run reports the plan; write cancels the pending jobs'
+        description: 'dry-run reports the plan; write applies it'
         type: choice
         options:
           - dry-run
@@ -55,6 +62,7 @@ jobs:
       DATABASE_URL: ${{ secrets.DATABASE_URL }}
       DATABASE_SSL_CA_BASE64: ${{ secrets.DATABASE_SSL_CA_BASE64 }}
       DATABASE_SSL_CA: ${{ secrets.DATABASE_SSL_CA }}
+      CLEANUP_TASK: ${{ inputs.task }}
       CLEANUP_MODE: ${{ inputs.mode }}
       CLEANUP_FIELDS: ${{ inputs.fields }}
       CLEANUP_INCLUDE_UNMARKED: ${{ inputs.include_unmarked }}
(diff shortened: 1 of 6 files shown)


silasfelinus:
REVIEWING: conductor-agent-20260906T005517Z-pr2443 at 2026-09-06T00:55:17Z

---

PR #2504, opened 2026-09-08
fix(home): admin-gate "Needs you" and give the newsfeed the room back

silasfelinus:
Two problems in the home page's right-hand column, one causing the other.

## "Needs you" was showing to everyone

`home-attention.vue` documented itself as **"admin-only by data"** — `conductorStore` only has gates when the projection is readable, so a signed-out visitor supposedly saw the empty branch and the panel collapsed out of the layout.

Only half of that was true. `/api/conductor/task-action` *is* admin-only, so a stranger could never **act** on a gate. But `/api/conductor/projects` is fully public and returns every project's whole task list, `needs-human` rows included. Measured against production while signed out:

```
52 projects · 44 needs-human tasks · 36 on active/continuous projects
```

That 36 is exactly the **"NEEDS YOU · 36"** a logged-out visitor was seeing.

This is not a leak — `/conductor` is a public page on purpose and serves the same rows (verified: 200 while signed out). But it is one person's decision queue on a stranger's front door, and it was the tallest thing in the column.

It's now gated on `userStore.isAdmin`, matching how the rest of the app gates admin-only affordances. Three parts:

- The panel itself (`home-attention.vue`).
- **The wrapper too, not just the panel.** The wrapper carries `xl:flex-1`, so leaving it in place would hold half the column's flexible height open around a section rendering nothing — which is most of why the newsfeed was a sliver.
- The phone tab drops out of the switcher, and the selected tab falls back to a pane that exists instead of defaulting to a "Needs you" tab that is no longer there.

## The newsfeed was losing height at both ends

> "the news section is too small, it could be twice as big, eating in the what we're building"

The feed now takes **twice** the flexible share "Needs you" does instead of splitting evenly with it, and the projects strip is **capped at a third of the column** and scrolls past that rather than taking whatever its natural height happens to be. Capped rather than flexed on purpose: it's short rows and reads best whole, so it keeps its own height right up to the cap.

Roughly, as a share of the column:

| | before | after |
|---|---|---|
| **signed out** | news ≈ 33% | **news ≈ 66%** (≈2×) |
| **signed in** | news ≈ 33% | news ≈ 44% (≈1.35×) |

Signed out, both effects compound and the feed roughly doubles — that's the view this report came from. Signed in, "Needs you" still needs its room, so the feed gains about a third rather than doubling. Say the word if you want it more aggressive in your own view.

## One regression this had to avoid

`conductorStore.fetchProjects()` lived in `home-attention`'s `onMounted`. That component no longer mounts for a signed-out visitor — who still sees the projects strip, whose progress bars (88%, 65%, 92%…) read that same store and would all have resolved to `null` and disappeared. The fetch moved to `home-page`'s `onMounted`, the component that always renders and actually needs it. It's cached in the store, so it stays a no-op when anything else already asked.

## Verification

- Production probe of `/api/conductor/projects` while signed out reproduces the exact 36 the screenshot shows, and `/conductor` confirmed 200 anonymously.
- `prettier --check` clean on both files.
- Full `vue-tsc` and the layout contract run in CI — no local install is available in this environment, so I'm relying on those rather than claiming a local pass. The new classes are `xl:`-prefixed arbitrary values applied through the existing `paneClass` helper, the same shape as the `xl:h-[46%]` / `xl:w-[22%]` values already in this file.

https://claude.ai/code/session_01BGb4quwjzWKjbgxAC6pgDx

---
_Generated by [Claude Code](https://claude.ai/code/session_01BGb4quwjzWKjbgxAC6pgDx)_

### DIFF
--- a/components/home/home-attention.vue
+++ b/components/home/home-attention.vue
@@ -39,14 +39,34 @@
   coordination system of record; answering hands it onward. The second is the
   one Silas asked for and the safer default, so it is the one in reach.
 
-  ADMIN-ONLY BY DATA, not by a check here: conductorStore only has gates when
+  ADMIN-ONLY BY AN EXPLICIT CHECK, because "by data" was never true.
+
+  This used to claim the panel gated itself: "conductorStore only has gates when
   the projection is readable, and /api/conductor/task-action is behind
-  requireAdminApiUser -- so a signed-out visitor sees the empty branch and the
-  column collapses out of the layout entirely.
+  requireAdminApiUser -- so a signed-out visitor sees the empty branch." Only
+  the second half held. `/api/conductor/task-action` is indeed admin-only, so a
+  stranger could never ACT on a gate -- but `/api/conductor/projects` is fully
+  public and returns every project's whole task list, `needs-human` rows
+  included. Verified against production 2026-09-08 while signed out: 52
+  projects, 44 needs-human tasks, 36 of them on active/continuous projects --
+  which is exactly the "NEEDS YOU · 36" a logged-out visitor was seeing.
+
+  So every stranger got Silas's personal decision queue on the front page.
+  Silas, 2026-09-08: "unlogged in users should not even see a 'needs you'
+  section as those are explicit to my user admin account."
+
+  The gate list is not secret -- the same rows are on /conductor, which is a
+  public page on purpose -- but it is not a front door for a visitor either:
+  it is one person's inbox, it is the tallest thing in the column, and it
+  pushes the newsfeed (which IS for visitors) into a sliver.
+
+  `userStore.isAdmin` is the check, matching how the rest of the app gates
+  admin-only affordances, and it renders nothing at all for anyone else rather
+  than an empty panel -- the column then gives that height to the newsfeed.
 -->
 <template>
   <section
-    v-if="gates.length || isLoading"
+    v-if="isAdmin && (gates.length || isLoading)"
     class="flex min-h-0 flex-col gap-1 kr-panel-flat p-2"
   >
     <header class="flex shrink-0 items-baseline justify-between gap-2">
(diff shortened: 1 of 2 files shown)


silasfelinus:
## `Prompt repair + icon audit` is failing, and it is not this PR's

**It is red on `main` too**, on the commit this branch is based on.

- This PR's run: [job 101905658386](https://github.com/silasfelinus/kind_robots/actions/runs/34176091938/job/101905658386)
- `main` @ `0718542` (#2502): [run 34172460169](https://github.com/silasfelinus/kind_robots/actions/runs/34172460169) — same job, same conclusion, **same icon list**

`0718542` is the commit that *added* this gate ("ci: gate legacy prompt repair and icon references"). It shipped two icons (`book-open`, `refresh-cw`) but the audit finds many more, so the check has been red from its first run on `main` — this is `run_number: 2` and it has never passed. It is not a flake, and re-running it would produce the same list.

This PR changes exactly two files, `components/home/home-page.vue` and `components/home/home-attention.vue`. Neither appears anywhere in the failure output, and the diff adds no `kind-icon:` reference.

### Why I am not fixing it here

Reproducing the audit locally against `assets/icons/` (331 icons on disk, 244 distinct references) finds **42 missing names across ~25 files** — `dream-pitch-sheet.vue`, `narrativeRoles.ts`, `conductorCards.ts`, `mandarin.vue`, `server-status.vue` and others, none of them touched by this change. A sample:

```
alert-triangle  briefcase  calendar  chart  check-square  document  file-plus
file-sparkles   git-branch hammer    handshake  history  key  language
map-pin  message-square  mirror  people  phone  pin  ruler  sheet  skills
sprout  swords  undo  user-round  volume  wish  x-circle  …
```

Clearing that means authoring new SVGs or repointing each reference at an existing icon — a design call per icon, on unrelated components. Porting it into a two-file home-page fix would widen this PR well past what it is for, so I have left it out deliberately rather than silently.

No fix for it exists on any branch yet. It wants its own PR, which would unblock `main` and every other open PR at the same time.

---

PR #2509, opened 2026-09-08
Redesign Storybook setup as an art-first story spread

silasfelinus:
Silas rejected the current Storybook setup in visual acceptance. The four-step form reads like settings UI, not an art-focused creative product.

This redesign keeps the existing Storybook engine/persistence and replaces the primary setup surface:

- one continuous, open setup screen instead of the visible four-step wizard
- large illustrated narrator and story-shape cards
- Character, Scenario, Dream, Facet, and Reward choices presented as 2:3 art-first cards rather than text rows with a thumbnail
- real entity artwork remains the primary selection surface when available; fallback art keeps the same image-shaped card geometry
- existing casting/role board is preserved inside the single story spread
- a visible “Your spread” summary and sticky “Open this story” action replace the separate review step
- the legacy StorybookPage stays mounted but hidden during setup so its mature restore/query-seeding/data-init lifecycle is preserved; it becomes visible once a story session exists
- the root AGENTS contract now codifies Silas’s standing rule that Kind Robots is an art-focused site and non-settings creative surfaces must be visual, not merely technically complete

The Storybook studio contract now guards the new one-screen/art-first architecture so a future cleanup cannot quietly turn it back into a form wizard.

Conductor storybook/t-016 has been sent back from human acceptance and should not be re-presented to Silas until this redesign is merged and deployed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,6 +60,25 @@ MariaDB. Production is served at `https://kindrobots.org` through the current co
 - When returning code in chat, return complete copy-paste-ready files or sections — never
   placeholders or ellipses. In-repo, normal targeted edits are fine.
 
+## Art-first interface standard
+
+Kind Robots is an **art-focused website**. Outside settings, administration, diagnostics,
+and similarly utilitarian screens, a feature is not finished merely because every field,
+button, or data object is technically reachable.
+
+- Prefer image-led cards, galleries, canvases, stages, decks, maps, timelines, and other
+  visual interaction surfaces over text-and-button control panels.
+- When a user chooses a Character, Dream, Scenario, Facet, Reward, artwork, or other entity
+  that has art, the art is a primary part of the choice. Do not reduce it to a tiny thumbnail
+  beside a form row unless dense utility UI genuinely requires that treatment.
+- Creative setup flows should feel open and compositional. Avoid wizard/tab proliferation
+  when the choices can coexist coherently on one visual surface.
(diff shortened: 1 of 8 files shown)


silasfelinus:
Picked this up as a Conductor pr-medic pass — CI had gone red and stale (last push 02:52 UTC, checks failed 02:55, still red 7.5h later at 10:26 UTC with no follow-up push).

Two failures, both fixed and pushed (`3b96588`):
- **`Contract verifiers` → lint-ratchet**: `no-unused-vars` regressed 5→6 from two `const store = source(storePath)` / `const shell = source(shellPath)` bindings in the new `verifyStorybookSessionLibrary.mjs` / `verifyStorybookStudio.mjs` guards that were read but never referenced again (the same paths are already exercised via `includesAll()`). Deleted the two dead bindings; both contract scripts still pass locally and the ratchet holds (333 problems, -59).
- **`comment-contract` → `verifyPopulationDraftQuality.ts`**: `GET /api/characters failed: 502` against live `kindrobots.org`. This diff doesn't touch that script, the comment-backfill corpus, or the characters API (confirmed via `git diff --stat` against the PR's actual base `3c7a970`, 7 files/799+/156-). Didn't need a separate re-run — the push above re-ran CI and it came back green, so treating it as the transient 502 it looked like rather than something this PR broke.

All checks are green as of the latest push. Leaving merge to you since this is your own branch/design call, not a Worker PR.

---

PR #2662, opened 2026-09-12
storybook: server-side story runs on ending decks, for every shape

silasfelinus:
Second slice of the card-driven storymaker groundwork (conductor storybook/t-029, t-030, t-032, t-033), on top of the narration contract in kind_robots#2660. **Schema-affecting.** No UI.

Deck authoring lives in silasfelinus/conductor#4168, which must merge first for the nightly database job to find the deck files.

## Why

Silas, 2026-09-12: a story should run through a bounded number of choices — an offered option, a written action, or a card played from the hero's character sheet — and end at "one of the predetermined endpoint endings, which they get credit for in a collection of their adventures."

Only the life shape had any of that. The beat shapes lived entirely in the reader's localStorage: no resume across devices, no collection, no ending. That needs a row.

## What

**Generalize the Life engine rather than fork it.** Life is now the deck with key `life`; a genre deck is the same machine with three or four axes instead of ten.

Schema, in two migrations that land together:
- `20260912120000` is additive. New `EndingDeck` (axes in bit order, pass value, per-shape turn budgets, a dormant `unlockAchievementId`). `LifeRun` gains `shape`, `deckId`, `turnBudget`, `narratorStyle`, `premise`, `scenarioId`, a `bible` snapshot, an `inventory`, and `pendingTurn`. `LifeChoice` gains `source` (OPTION/CUSTOM/SHEET), `optionId`, `rewardId`, `stateDelta`, `artPrompt`. It seeds the life deck and adopts the 1,024 existing endings.
- `20260912123000` is the **one non-additive step**: it drops the global unique on `LifeEnding.outcomeKey` in favour of `(deckId, outcomeKey)` and makes `deckId` NOT NULL. Every three-axis genre deck produces `000`..`111`, so Mystery's `101` and Heist's `101` are different endings that both have to exist. Its header records why it is safe in this order — Prisma applies migrations by name so the backfill runs first, the old build's ending lookup is a plain equality query that survives losing an index, and nothing in the running build inserts `LifeEnding` rows.

Engine:
- `server/utils/storybookRuns.ts` creates a run from a board of card slugs, resolved and permission-checked server-side and then snapshotted, so later edits to a Character cannot rewrite a scene that already happened.
- **The turn route records the move and narrates the next scene in one call, and the server holds the offered choices.** An option's effects come from `LifeRun.pendingTurn`, never the request body; a written move or a card play is scored by the narrator and clamped here. Nothing is written unless narration succeeds, and replaying a recorded turn returns it rather than advancing twice.
- Character-sheet plays: the protagonist's own Rewards are in play from turn one. An ITEM is spent when played; other card types stay but cannot be played twice in a row.
- `resolveLifeRunEnding` becomes deck-generic `resolveStoryRunEnding`, with the old name kept as a delegate, and it awards a deck's COLLECTION achievement once every ending in that deck is found.
- A genre deck's axes never reach the client: option effects and stat values are stripped for any deck but LIFE.
- The turn gate binds only runs the new engine created (`turnBudget` non-null), so a life run already in flight resolves exactly as it does today.
- Routes under `/api/storybook/`. `/api/davinci/*` is untouched and keeps working; it now shares one narrator lookup instead of holding a second copy.
- Narration is metered with `manaGate` on the new routes; the pre-deck narrate route never was.

Deck import:
- `utils/scripts/seedStorybookDecks.ts` walks a deck YAML file or directory and upserts the deck, its endings, an Achievement and LifeAchievement per ending, and the per-deck COLLECTION achievement. A deck whose endings do not exactly cover `2^axes` is refused, because a missing `outcomeKey` is a 404 the reader only meets after playing the whole story.
- `seedDaVinciEndings.ts` now also ensures the life deck exists and sets `deckId`, because the nightly job builds its database with `prisma db push` and never runs migrations.

## Verified

| Check | Result |
|---|---|
| `npm run test` (nuxi prepare + vue-tsc) | clean |
| `npm run test:schema-migration-parity` | schema change has migration coverage |
| `npm run test:storybook` | exit 0 |
| narration, deck-math and deck-authoring contract suites | pass |
| `npm run lint:js` | 333 problems, one **below** `main`'s 334; no new findings |
| `npx prettier --check` on every touched file | clean |

New DB-free `verifyStorybookDecks.ts` checks the authoring format, the coverage rule, and the per-deck award namespacing, then validates the real conductor deck files when conductor is checked out beside this repo: each genre deck authors all eight endings in complete sentences, not every ending is a win, and no slug or trigger code is reused across decks.

New DB-backed `verifyStorybookPlayLoop.ts` plays a whole short story against a stub narrator: option, written and sheet moves, an option id that was not offered, a turn out of order, a turn past the budget, item consumption, deterministic resolution, idempotent re-resolve, collection credit with unfound endings still silhouetted, and a legacy deckless run. It runs in `davinci-seed-verify.yml`, which has a scratch MariaDB and now also seeds all three genre decks into one database — the only place a same-bit-width collision between decks would surface. Neither DB suite could run in the authoring sandbox (no database, no docker daemon), so **that job is the one to watch on this PR**.

**Schema-affecting: merged is not deployed.** Both migrations apply in one deploy, in name order, ahead of the container swap.

https://claude.ai/code/session_012hXcMTb3XVsamTqP38mkYJ


silasfelinus:
**REVIEWING: conductor-agent-20260912T103500Z-t104 at 2026-09-12T11:35:00Z**

Not mergeable as-is — the DB-backed play-loop verifier the PR body itself flagged ("that job is the one to watch on this PR") ran on CI and found two real failures in `Generator + importer regression`:

```
3. a written move is scored by the narrator and clamped
  FAIL: the narrator's +-9 proposal is clamped to +-2 ({"truth":9,"nerve":-9})

4. a card played from the sheet is a move
  FAIL: the played Reward is recorded on the turn
```

These aren't flaky/infra — `verifyStorybookPlayLoop.ts` ran to completion (47/49 checks passed) and is asserting real engine behavior:

1. The written-move clamp (`±2` per the PR description) doesn't appear to actually be applied to `moveEffects` before persisting/returning — the raw narrator proposal (`truth:9, nerve:-9`) came through unclamped.
2. The played Reward isn't being recorded on the turn for a SHEET-sourced move (item-spend and repeat-play-prevention both pass, so this looks like a narrow gap in what gets written, not the whole path).

Everything else is clean — schema/migration ordering, deck-math/narration contracts, TypeScript, lint, layout-contract. This PR is schema-affecting per its own description (the non-additive migration dropping `LifeEnding.outcomeKey`'s global unique), so per repo policy it needs the full suite green, not just the parts that happened to pass, before it's mergeable — holding off on merge and on the migration audit until that's the case.

Please fix the clamp application and the Reward-recording path in the turn-write logic, push, and re-run `Generator + importer regression` to confirm both FAILs are gone. Re-flag once green — happy to do the schema/migration line-by-line audit at that point.

---

silasfelinus:
Both fixes confirmed in the follow-up commit and `Generator + importer regression` is green. Also completed the schema/migration line-by-line audit I held off on: both migrations are additive-only (`CREATE TABLE`/`ADD COLUMN`/`CREATE INDEX`/`ADD CONSTRAINT`) plus a precedented backfill `INSERT`/`UPDATE` and a `DROP INDEX` constraint-swap + `MODIFY ... NOT NULL` in the second — same shape as this repo's own prior accepted migrations (e.g. `20260727022500_facet_taxonomy_authority`). No `DROP` of tables/columns/data. All 52 checks green, `mergeable_state: clean` — merged.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XfLdM2KYb5sheAWpT5991x)_

---

PR #2686, opened 2026-09-12
feat(storybook): the four story modes, the length dial, endless play, and the narrator card (SCHEMA)

silasfelinus:
All of milestone **m8** from the storymaker redesign — storybook/t-039, t-040, t-041, t-042.

Silas, 2026-09-12: *"Stories should be able to be selected as open-ended (endless mode), episodic (scenario based), structured (da Vinci mode), and taskmaster"* and *"since the narrators actually exist as bot Narrators, they provide the general voice, but how they deliver it can still be adjusted."*

## SCHEMA-AFFECTING — expand only, and the handoff is the point

`prisma/migrations/20260913120000_add_story_modes_to_story_shape/` widens `StoryShape` with `OPEN_ENDED`, `STRUCTURED`, `TASKMASTER` and **keeps** `SHORT_STORY` / `CHAPTERED` / `LIFE`, so the build running during the deploy handoff can still write the value it knows. Three things make that safe:

- rows are backfilled — `LIFE → STRUCTURED`, `SHORT_STORY` and `CHAPTERED → OPEN_ENDED`, `EPISODIC` untouched;
- the column **default moves `LIFE → STRUCTURED`**, the same meaning under the new taxonomy, so an insert from the old build (which never sets `shape` — `createLifeRun` doesn't) still lands correct rather than on a beat-shape value;
- `MODE_BY_ENUM` in `server/utils/storybookRuns.ts` reads the legacy values defensively, so a row written mid-handoff is understood and **a run in flight never changes mode under the reader**.

Dropping the legacy values is a later migration, once no deployed build emits them. That drop is the non-additive half and is deliberately not here. Merged ≠ deployed: this needs the Alexandria deploy before the new routes speak modes in production.

## Length is a setting, not a mode (t-041)

Short story and chaptered tale were one story at two lengths, so both collapse into open-ended with a turn budget on the dial: `StoryBoardInput.turnBudget`, 3–40, defaulting to the deck's budget for the mode. Per Silas's clarification (*"that doesn't count reasonable settings, start story, etc"*) it is a **setting, not a card**, and it stays out of every hand-facing payload.

The prose bands follow the same logic: a scene is not longer because the story it belongs to is. Open-ended and episodic share one band, `structured` keeps the 20–400 life band its live runs were written against (`verifyDaVinciNarration.ts` still pins it), and `taskmaster` gets the tightest of the four because it narrates beside the reader's real work.

## Endless (t-040)

`turnBudget` NULL now **means endless** rather than unset: the narrator is never told a turn is the last one, a scene is always waiting, and `readyToResolve` turns true once the deck's floor is passed. An endless story is still collectible — the reader ends it and it resolves into the same deck as any other, rather than trailing off uncounted. It is not a free pass either: it still has to clear `minTurnsBeforeResolve`, so "bring this to an end" cannot collect an ending on turn one.

A pre-deck life run also has NULL there, for its own reason, and is told apart by `deckId` — its ungated resolve is untouched.

## The narrator is a Bot card with a delivery dial (t-042)

Mockup round 1 had the Narrator slot as a radio row of five abstract voices. It is a real Bot dealt into the slot, with the dial on the placed card. The run already carried both halves (`botId` + `narratorStyle`) and `loadRunNarrator` already read a Bot's persona into the system prompt, so this is three small pieces:

- **`GET /api/narrators`** lists the narrator Bots for the hand to deal — the list sibling of `/api/narrators/[type]/[slug].get.ts`, returning only what a card face needs. The prompt text is deliberately absent: a narrator's system prompt is the server's, and a hand does not need it to draw a card.
- **`createStoryRun` refuses a `botId` that is not an active NARRATOR.** Nothing failed before — `loadRunNarrator` would happily read any Bot into the prompt, so the slot quietly narrated in a voice never written to narrate. Ownership is still checked separately by `assertAttachable`; this is about the Bot's job, not its access.
- **The style block is reframed as `DELIVERY`** and says out loud that it modulates the voice established above rather than replacing it, so the prompt never carries two competing narrators. The five directives are unchanged.

## Not in this PR, on purpose

`stores/storybookStore.ts` is **not** converted. Its four "structures" drive the client beat loop's own scheduling (`short-story` ends after four beats) — the behaviour the length dial replaces — and that loop is deleted wholesale by storybook/t-037 once the new screens land. Converting it now would rewrite a loop scheduled for deletion and change copy on a screen being replaced. A legacy client posting `'short-story'` still opens an open-ended run through `LEGACY_MODE_ALIASES`, so nothing breaks in the meantime.

## Verified

`npm run test` (vue-tsc) clean, `npm run test:storybook` and `npm run test:davinci-narration` pass, eslint clean on every changed file, `verifyStorybookNarration.ts` green — including new assertions that the delivery block modulates rather than replaces, and that the Bot's identity precedes the dial.

`verifyStorybookPlayLoop.ts` gains the new ground and runs against a real database in `davinci-seed-verify.yml` (it cannot run in the sandbox — no DB, no docker): a pre-mode wire name still opens a story, an unknown mode is refused, the dial overrides and is bounded, only open-ended may run budgetless, an endless run plays past the old budget without ever being told it is final, the reader's resolve credits an ending, and a non-narrator Bot cannot take the Narrator slot.

Deck YAML companion: [silasfelinus/conductor#4209](https://github.com/silasfelinus/conductor/pull/4209), merged.

https://claude.ai/code/session_012hXcMTb3XVsamTqP38mkYJ


silasfelinus:
**Two red checks here are not this PR's — Alexandria is refusing connections from the runners.**

- **Contract verifiers** fails in `verifyAcademyStarterManifest.ts` with a bare `fetch failed`.
- **comment-contract** fails in `verifyPopulationDraftQuality.ts` with `connect ECONNREFUSED 75.111.66.70:443`.

Both are live-host fetches against Alexandria, and neither script is touched by this diff (this PR is the story-mode enum, the turn budget dial, and the endless path).

Ruled out as this PR's, two ways:

1. **Red on `main` too.** Contract Tests run [6950](https://github.com/silasfelinus/kind_robots/actions/runs/34723699793) on `main` at 45f8a61 — the docs-only boundary-doc merge from #2684, which changes one markdown file — fails the same "Contract verifiers" job. A markdown edit cannot break a manifest fetch.
2. **Re-run reproduces it.** I re-ran both failed jobs once; both failed again with the same signature, ~9 minutes apart. That is the one re-run, spent.

There is no fix to port: this is the home server being unreachable from GitHub's runners, not a code defect, so nothing in this PR or any other would turn those two green while the host is down. Everything that does not need Alexandria is green here, including the two that actually cover this change — **TypeScript** and the DB-backed **Generator + importer regression**, which is where `verifyStorybookPlayLoop.ts` exercises the new modes, the dial, and the endless run against a real database.

Holding the merge until those two recover rather than forcing it. Worth a separate look at why two contract guards depend on a live home server being up — a CI check that goes red when a machine at home reboots is flaky by construction — but that is its own task, not this PR's to widen into.

---

PR #2719, opened 2026-09-14
fix: make render recovery and local Comfy paths explicit

silasfelinus:
## What changed
- expose `latestDoneAt` and `latestHostbufFailureAt` from `/api/art/queue/stats`
- stop relying on created-at throughput as proof that the renderer recovered
- add `scripts/repair-comfy-local-paths.ps1` to align Ferngrotto's live `extra_model_paths.yaml` with the existing sync layout
  - local `clip` + `text_encoders` resolve to `models/clip`
  - Alexandria no longer carries `is_default: true`
  - timestamped backup + strict shape/invariant checks before writing

## Why
The live machine has the 4.88 GB Qwen/Krea text encoder at `D:\comfy\comfy-fast\models\clip\qwen3vl_4b_fp8_scaled.safetensors`, while its local path map points `clip` and `text_encoders` at `models/text_encoders`. That makes the local encoder invisible and permits fallback to the Alexandria SMB copy, matching the observed `hostbuf_file_reader_read failed` path.

Separately, a monitoring incident should clear only after a render actually completes successfully after the failure. `windowThroughput` uses `createdAt`, so it cannot establish that fact.

## Safety
- additive API fields only
- repair script is not executed automatically and does not restart ComfyUI
- repair script creates a backup and refuses unexpected YAML shapes
- no deployment or production mutation is performed by this PR


silasfelinus:
REVIEWING: reviewer-20260914-hostbuf-render at 2026-09-14T08:27:51Z

---

PR #2723, opened 2026-09-14
fix: keep Ferngrotto models under comfy-fast

silasfelinus:
## What changed
- restore `comfyui_local.base_path` to `D:/comfy/comfy-fast`
- retain the correct local `clip` / `text_encoders` mapping to `models/clip`
- retain Alexandria as a non-default fallback
- document that the populated local model store is `D:/comfy/comfy-fast/models`

## Why
The earlier repair assumed the canonical model store was `D:/comfy/models`. Silas confirmed the actual populated store, including all Krea 2 assets, is `D:/comfy/comfy-fast/models`. Redirecting the YAML to `D:/comfy` therefore pointed Comfy at an empty tree and triggered an unnecessary duplicate copy.

The original base root was correct; only the text-encoder mapping and share/default priority needed repair.


silasfelinus:
REVIEWING: openai-scheduled-2026-09-14T091726Z-pr2723-review-a30 at 2026-09-14T09:17:26Z

